PR #4, opened 2019-11-13
Re-write history to get rid of big binary fine in one of the initial commits

j-fuentes:
I committed by mistake the preflight binary when `.gitignore` wasn't excluding `/builds`. (see commit [d1d1cc8](https://github.com/jetstack/preflight/commit/d1d1cc871d63a19499e0a68fbc7cfbaafddae4df))

As a result, GitHub was slow on every pull/push. It was even showing a warning about that file.

I have used [BFG](https://rtyley.github.io/bfg-repo-cleaner/), as [GitHub recommends](https://help.github.com/en/github/managing-large-files/removing-files-from-a-repositorys-history), to get rid of that file in all the commit history.


j-fuentes:
Closing this as it has been fixed by dropping the problematic commit and pushing to master.

PR #11, opened 2019-12-02
Add ci-test and ci-build and ci-publish targets

j-fuentes:
These new targets are going to be triggered by prow CI.


j-fuentes:
At the moment, this is executing all the dependency targets sequentially, which is suboptimal, but we will improve that later by adding Bazel or something else that allows parallel execution.

PR #20, opened 2019-12-17
Include preflight-version, and package information in reports

j-fuentes:
This changes the `api.Report` schema to include both the version of Preflight that was used and also the additional information of the Package.

Notice this is a breaking change since we are mutating the existing `Package` property in the report to a `Package` object.

### DIFF
--- a/api/report.go
+++ b/api/report.go
@@ -2,17 +2,32 @@ package api
 
 // Report contains the fields of a Preflight report
 type Report struct {
-	// Unique ID of the report
+	// Unique ID of the report.
 	ID string `json:"id"`
-	// Timestamp indicates when the report was generated
+	// PreflightVersion indicates the version of preflight this report was generated with.
+	PreflightVersion string `json:"preflight-version"`
+	// Timestamp indicates when the report was generated.
 	Timestamp Time `json:"timestamp"`
-	// Cluster indicates which was the target of the report
+	// Cluster indicates which was the target of the report.
 	Cluster string `json:"cluster"`
-	// Package indicates which package was used for the report
-	Package     string          `json:"package"`
-	Name        string          `json:"name"`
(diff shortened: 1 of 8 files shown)


api/report.go
```diff
@@ -2,17 +2,32 @@ package api
...
-	// Timestamp indicates when the report was generated
+	// PreflightVersion indicates the version of preflight this report was generated with.
+	PreflightVersion string `json:"preflight-version"`
+	// Timestamp indicates when the report was generated.
 	Timestamp Time `json:"timestamp"`
-	// Cluster indicates which was the target of the report
+	// Cluster indicates which was the target of the report.
 	Cluster string `json:"cluster"`
-	// Package indicates which package was used for the report
-	Package     string          `json:"package"`
```
j-fuentes: Actually, in order to keep better backwards compatibility, let's not repurpose the `Package` property.

Let's keep `Package string` and create a new `PackageInformation Package`.

WDYT @wwwil ?

wwwil: Yeah I think it's better to avoid breaking change. Should the `Package` struct also be renamed `PackageInformation`?

j-fuentes: Yes, that's what I have done.

api/report.go
```diff
@@ -2,17 +2,34 @@ package api
...
+	// Name is the name of the package that was used for this report.
+	Name string `json:"name"`
+	// Description is the description of the package that was used for this report.
+	Description string `json:"description,omitempty"`
+	// Sections contains the sections of the package that was used for this report.
+	Sections []ReportSection `json:"sections,omitempty"`
+}
+
+// Package contains all the details to identify a package.
+type Package struct {
```
wwwil: I was thinking we should also rename this to `PackageInformation`.

j-fuentes: ok, makes sense.

PR #25, opened 2019-12-23
Do not enforce numeric IDs and enforce RootQuery

j-fuentes:
Numeric IDs are very artificial and not meaningful at all.

I wouldn't enforce them in the linter.

Also, the linter wasn't enforcing RootQuery, but it is mandatory.

### DIFF
--- a/pkg/lint/manifest.go
+++ b/pkg/lint/manifest.go
@@ -111,16 +111,6 @@ func LintSection(manifestID string, section packaging.Section) []LintError {
 	if section.ID == "" {
 		lint("Section ID absent")
 	}
-	re := regexp.MustCompile(`^(\d+)\.(\d+)$`)
-	matches := re.FindStringSubmatch(section.ID)
-	if len(matches) != 3 {
-		lint("Malformed Section ID")
-	} else {
-		if matches[0] != section.ID {
-			lint("Incorrectly formatted Section ID")
-		}
-		// No need to check matches[2], the regex implicitly did so
-	}
 
 	if section.Name == "" {
 		lint("Section Name absent")
@@ -163,19 +153,6 @@ func LintRule(sectionID string, rule packaging.Rule) []LintError {
 
 	if rule.ID == "" {
 		lint("Rule ID absent")
-	}
-	re := regexp.MustCompile(`^(\d+)\.(\d+)\.(\d+)$`)
-	matches := re.FindStringSubmatch(rule.ID)
-	if len(matches) != 4 {
-		lint("Malformed Rule ID")
-	} else {
-		if matches[0] != rule.ID {
-			lint("Incorrectly formatted Rule ID")
-		}
-		if fmt.Sprintf("%s.%s", matches[1], matches[2]) != sectionID {
-			lint("Rule ID not decendant of Section ID")
-		}
-		// No need to check matches[3], the regex implicitly did so
 	}
 
 	if rule.Name == "" {
--- a/pkg/lint/manifest.go
+++ b/pkg/lint/manifest.go
@@ -36,6 +36,10 @@ func LintPolicyManifest(manifest packaging.PolicyManifest) []LintError {
 		lint("Manifest PackageVersion must be semver")
 	}
 
+	if manifest.RootQuery == "" {
+		lint("Manifest RootQuery absent")
+	}
+
 	sections := manifest.Sections
 	if len(sections) == 0 {
 		lint("No sections in manifest")
--- a/pkg/lint/manifest.go
+++ b/pkg/lint/manifest.go
@@ -116,6 +116,12 @@ func LintSection(manifestID string, section packaging.Section) []LintError {
 		lint("Section ID absent")
 	}
 
+	// section IDs should follow the same syntax as rules IDs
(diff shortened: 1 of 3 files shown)


wwwil:
This generally looks good to me!

Is there anything currently checking that rules defined in the policy manifest have corresponding rules in Rego?

I wonder if the IDs should just be `int`, rather than the `1.2.3` format. Then referring to rules in sections in packages can be done using that notation implicitly. E.g. rule with ID `3` in section with ID `2` in package with ID `1` could be referred to in text as `1.2.3`.

j-fuentes:
> This generally looks good to me!
> 
> Is there anything currently checking that rules defined in the policy manifest have corresponding rules in Rego?

Not at the moment. But it is good to have. For now, if Preflight cannot find a result for a rule in the manifest it marks it as missing in the final report.

> I wonder if the IDs should just be `int`, rather than the `1.2.3` format. Then referring to rules in sections in packages can be done using that notation implicitly. E.g. rule with ID `3` in section with ID `2` in package with ID `1` could be referred to in text as `1.2.3`.

What's the value on that? I personally prefer to use meaningful IDs, so I can have an idea of what it is just by looking at it. I think it is more natural and easier deal with for humans. In case of a machine, it really does not make a difference if it is a number or not.

wwwil:
My original thoughts were that the meaningful part was the name, and the main value of numeric IDs would be to make referring to things more concise: You could say "`1.1.1` and `1.1.2` failed" rather than "`pods_basic. resources. container_cpu_limit` and `pods_basic. resources.container_mem_limit` failed".

But I don't know if that has much value, and the other terms are a lot more immediately descriptive. So we probably are better to drop the requirement for numeric IDs.

@charlieegan3 What are your thoughts?

pkg/lint/manifest.go
```diff
@@ -111,16 +115,6 @@ func LintSection(manifestID string, section packaging.Section) []LintError {
 	if section.ID == "" {
 		lint("Section ID absent")
 	}
-	re := regexp.MustCompile(`^(\d+)\.(\d+)$`)
```
wwwil: Should the rule IDs be completely free form? Or should we restrict spaces and symbols? Just thinking we might want to use them, for example, in URLs in the Preflight Web UI.

charlieegan3:
I guess the value in a semantic-style id (1.2.3) was the hierarchy? I'd _hope_ we could achieve this with OPA packages (e.g. `package pods.containers.resources.requests` if we had so many rules it was worth breaking down of course) 

When I saw those versions initially (some months ago) I figured it was related to the k8s version the rules were used for.

pkg/lint/manifest.go
charlieegan3: Can't OPA perform this validation? letters and underscores I think?

wwwil: This is the rule ID from the _policy manifest_, so it needs to match to a Rego rule so we can create the report. So we should validate it here to make sure it conforms to Rego's naming requirements otherwise it won't be able to match a Rego rule.

charlieegan3: Ahh right - makes more sense.

wwwil: @j-fuentes What do you think about enforcing Rego's rule name requirements for IDs? So the regex would be something like `^[_a-zA-Z][_a-zA-Z0-9]$` based on [Rego grammar](https://www.openpolicyagent.org/docs/latest/policy-reference/#grammar) for `var` names.

j-fuentes: Yeah, that makes sense.

j-fuentes:
> I guess the value in a semantic-style id (1.2.3) was the hierarchy? I'd _hope_ we could achieve this with OPA packages (e.g. `package pods.containers.resources.requests` if we had so many rules it was worth breaking down of course)

Yeah, the benefit of the `1.2.3` format was the hierarchy. However, hierarchy is already imposed by the policy manifest.

> When I saw those versions initially (some months ago) I figured it was related to the k8s version the rules were used for.

:P I agree, it can be confusing.

wwwil:
Ah, I remember now, I copied the numbering from the CIS Benchmark style and [kube-bench](https://github.com/aquasecurity/kube-bench/blob/master/cfg/cis-1.5/controlplane.yaml). I'm still in agreement we should changed it, but I knew there was some other reason for it that I couldn't recall before.

charlieegan3:
This all looks good to me. I have one question, what does a rule that starts with `_` e.g. `_preflight_blah`? Do we want to allow this (we appear to have a test case for it)

wwwil:
> This all looks good to me. I have one question, what does a rule that starts with `_` e.g. `_preflight_blah`? Do we want to allow this (we appear to have a test case for it)

It's allowed by the Rego grammar so I don't see why not.

j-fuentes:
> This all looks good to me. I have one question, what does a rule that starts with `_` e.g. `_preflight_blah`? Do we want to allow this (we appear to have a test case for it)

The only reason I added it is because rego allows it. Notice in the regexp there is a special check for the first character not being a number.

charlieegan3:
Fair enough

/approve
/lgtm

Issue #26, opened 2019-12-23
Support for testing packages


j-fuentes:
Currently, users need to use OPA's CLI to perform tests on the REGO files.

Preflight CLI should have a way to execute the tests of a package with `preflight package test <package>` without the user having to install OPA's CLI.

PR #29, opened 2019-12-31
Package test command

j-fuentes:
This adds the `preflight package test <pkg>` command.

It allows the user to point Preflight CLI to a pkg and execute the OPA tests that it contains.

Closes: #26


cmd/test.go
```diff
@@ -0,0 +1,83 @@
...
+
+var testParams = struct {
+	verbose bool
+	timeout time.Duration
+}{}
+
+var testCmd = &cobra.Command{
+	Use:   "test",
+	Short: "Test REGO inside a Preflight package",
+	Long: `This uses OPA's engine to run all the testsuites
```
wwwil:
```suggestion
	Long: `This uses OPA's engine to run all the test suites
```

PR #30, opened 2020-01-03
Add AKS data-gatherer

j-fuentes:
This adds the first implementation of the AKS datagatherer and a sample package that uses it.

This also bumps Go version to 1.13 since that is the branch we are using in CI.


j-fuentes:
It can be run with:

```
go run . check -c ./examples/aks.preflight.yaml
```

This is an example of the data that is gatherered:

```json
{
  "aks": {
    "Cluster": {
      "properties": {
        "networkProfile": {
          "dockerBridgeCidr": "172.17.0.1/16",
          "dnsServiceIP": "10.0.0.10",
          "serviceCidr": "10.0.0.0/16",
          "podCidr": "10.244.0.0/16",
          "networkPlugin": "kubenet"
        },
        "enableRBAC": true,
        "nodeResourceGroup": "MC_preflight_preflight-test_uksouth",
        "addonProfiles": {
          "httpApplicationRouting": {
            "config": null,
            "enabled": false
          }
        },
        "servicePrincipalProfile": {
          "clientId": "***********************"
        },
        "agentPoolProfiles": [
          {
            "osType": "Linux",
            "maxPods": 110,
            "storageProfile": "ManagedDisks",
            "osDiskSizeGB": 100,
            "vmSize": "Standard_B2s",
            "count": 1,
            "name": "agentpool"
          }
        ],
        "fqdn": "preflight-test-dns-f2de96e5.hcp.uksouth.azmk8s.io",
        "dnsPrefix": "preflight-test-dns",
        "kubernetesVersion": "1.14.8",
        "provisioningState": "Succeeded"
      },
      "type": "Microsoft.ContainerService/ManagedClusters",
      "name": "preflight-test",
      "location": "uksouth",
      "id": "/subscriptions/*****************/resourcegroups/preflight/providers/Microsoft.ContainerService/managedClusters/preflight-test"
    }
  }
}
```

j-fuentes:
The credentials file required for this has been obtained with the Azure CLI by running `az account get-access-token > /tmp/credentials.json`.

This access token is very short-living (max 1 hour), which is suboptimal for this use case. I am exploring other alternatives to authenticate against Azure.

pkg/datagatherer/aks/aks.go
```diff
@@ -0,0 +1,107 @@
...
+		ctx:           ctx,
+		resourceGroup: resourceGroup,
+		clusterName:   clusterName,
+		credentials:   credentials,
+	}, nil
+}
+
+// Fetch retrieves cluster information from AKS.
+func (g *AKSDataGatherer) Fetch() (interface{}, error) {
+	client := &http.Client{}
```
wwwil: Is there a reason for using the HTTP client rather than the Azure client? https://github.com/Azure-Samples/azure-sdk-for-go-samples/blob/master/compute/container_cluster.go

wwwil: I don't have a strong opinion about how it should be done, just comparing it to other gatherers.

j-fuentes: Actually, I didn't know about https://github.com/Azure-Samples/azure-sdk-for-go-samples.

I checked https://github.com/Azure/azure-sdk-for-go and https://github.com/Azure/aks-engine and none of them can serve as a client for what we wanted here.

I will give https://github.com/Azure-Samples/azure-sdk-for-go-samples a try.

wwwil:
@j-fuentes I can confirm I've been able to run this against an actual AKS cluster and it worked correctly, so I've approved it in its current form. If you change the AKS client I will give another test, or if you decide to leave it as is I can `lgtm` it!

j-fuentes:
> @j-fuentes I can confirm I've been able to run this against an actual AKS cluster and it worked correctly, so I've approved it in its current form. If you change the AKS client I will give another test, or if you decide to leave it as is I can `lgtm` it once rebased.

I have just rebased the current master.

After checking `Azure-samples/azure-sdk-for-go-samples` it does not seem like something we should rely on because of several reasons: it looks like "sample" code, I don't think it is designed to be a library.

Issue #36, opened 2020-01-08
Expose context information about why a check is failing


charlieegan3:
![Screenshot from 2020-01-08 12-18-49](https://user-images.githubusercontent.com/1774239/71977586-123c5c80-3211-11ea-848f-986cb5368449.png)

This is a screenshot from https://preflight.jetstack.io/

It shows the result of running our basic pod checks. We can see that some pods are missing requests and limits - oh dear.

This is based on the following report: https://github.com/jetstack/preflight/blob/master/examples/reports/pods.json#L7

All that's exposed in the report is the sucess/fail - not the reason.

When using OPA as something that backs a k8s webhook the idiomatic way to do this is to return a message if the rule fails, otherwise return nothing.

In our pod example, I think i'd return the pod name & namespace in a single string, one for each violation.

I guess we might also remove the `success` attribute  and replace it with a `violations` list - if there are violations then it failed?

j-fuentes:
Sounds reasonable.

charlieegan3:
Closing as this as been fixed by the above PRs.

PR #38, opened 2020-01-08
Show violations

charlieegan3:
This is an attempt at the start of implementing #36 

I believe to surface this information about violations, we'll need to make some fairly fundamental changes to the rule format (away from non booleans rules). Early days so hopefully ok? :grimacing: 

This is a draft, it is incomplete. I have only converted one rule `preflight_container_cpu_limit`

wdyt @j-fuentes @wwwil ?

p.s. sorry about the whitespace changes... https://github.com/jetstack/preflight/commit/eb10fde7a924d13a7c250b77a2e6181dc6bf1652 is where the good stuff happens


preflight-packages/examples.jetstack.io/pods_basic/pods.rego
```diff
@@ -4,66 +4,66 @@ package pods
 import input["k8s/pods"] as pods
 
 # Rule 'container_cpu_limit'
```
wwwil: I think this comment needs to move down.

wwwil:
This looks like a good direction to go in. It's a similar style to [Conftest](https://github.com/instrumenta/conftest), and I wonder if we should actually go further and use the same style as they do so Preflight packages could be compatible with either tool.

I'm concerned about the amount of rewriting that will be done to our internal packages, but I suspected that might have to happen as we worked out exactly how the standard should look.

wwwil:
Could you also look at what changes need to be made to Preflight for it to interpret the new results format?

preflight-packages/examples.jetstack.io/pods_basic/pods.rego
charlieegan3: Fixed in 3096205c404c72b6d325fc353c8e179599ccd7f4

charlieegan3:
Morning Wil, thanks for the review.

> It's a similar style to Conftest
Indeed, this is also similar to how I've seen policies set up elsewhere. Where the rules are for what the action is, rather than what's being validated. `deny`, `allow` instead of `preflight_container_cpu_limit`.

I think this takes us closer towards that for sure but I'm happy to go further towards a package spec that looks more like the conftest one.

> concerned about the amount of rewriting that will be done to our internal packages
Indeed, but now's as good a time as any since the volume of rego is still pretty small (gke+pods only).

> Could you also look at what changes need to be made to Preflight for it to interpret the new results format?

I'll start this today I think in order to get a feel for what this looks like all way through.

charlieegan3:
Then we could have something like this: https://github.com/charlieegan3/policing-christmas-trees/blob/master/server/main.rego#L6-L17 to delegate violations to different parts of the package.

charlieegan3:
> Could you also look at what changes need to be made to Preflight for it to interpret the new results format?

@wwwil It looks a little like this: https://github.com/jetstack/preflight/pull/38/commits/a54ed799f36cb4352c3bbd7db56cf6605890707c#diff-cf2a8b7441d3eb104ee774d17e98516dR26-R31

Basically, failure is now a value as a non-empty string array listing violations messages.

wwwil:
> Basically, failure is now a value as a non-empty string array listing violations messages.

Yeah that's not too much of a change so seems reasonable.

charlieegan3:
Thanks Wil, I'll continue to make the changes to the policies.

charlieegan3:
I've marked this as ready for review. I think I've got everything except the guide Jose had merged recently - I'll try do that first thing.

- [ ] update the guide on writing policies

Makefile
```diff
@@ -63,6 +63,11 @@ bundle-all-platforms:
 	$(MAKE) GOOS=darwin  GOARCH=amd64 ./bundles/preflight-bundle-darwin-amd64.tgz
 	$(MAKE) GOOS=windows GOARCH=amd64 ./bundles/preflight-bundle-windows-amd64.tgz
 
+# Packages
+
+package-test:
+	opa test -v --explain=notes $(ROOT_DIR)/preflight-packages/examples.jetstack.io
```
charlieegan3: @j-fuentes, based on the standup comment, there might be some changes to make here?

api/report.go
```diff
@@ -62,7 +62,7 @@ type ReportRule struct {
 	Success bool `json:"success"`
 	// Value contains the raw result of the check.
 	Value interface{} `json:"value,omitempty"`
-	// Missing indicated wherer the REGO rule was missing or not.
+	// Missing indicated wherer the Rego rule was missing or not.
```
wwwil:
```suggestion
	// Missing indicates whether the Rego rule was missing or not.
```

docs/how_to_write_packages.md
```diff
@@ -101,19 +102,19 @@ package pods
...
-
-You will get faster as you write more Rego rules. In order to speed up this process, it's best to write tests for your rules, even if you think they are not needed. It means you can iterate fast while writing rules and make sure the rules are doing what you intended. It is conventional to name the test files for `policy.rego` as `policy_test.rego`.
+Rego is a declarative language and has a bit of a learning curve. It is strongly recommended to read ["The Basics"](https://www.openpolicyagent.org/docs/latest/policy-language/#the-basics). Also, it is useful to have the [language reference](https://www.openpolicyagent.org/docs/latest/policy-reference/) to hand.
 
+In order to speed up the process of writing Rego rules, it's best to write tests. It means you can iterate fast while writing rules and make sure the rules are doing what you intended without int. It is conventional to name the test files for `policy.rego` as `policy_test.rego`.
```
wwwil:
```suggestion
In order to speed up the process of writing Rego rules, it's best to write tests. It means you can iterate fast while writing rules and make sure the rules are doing what you intended. It is conventional to name the test files for `policy.rego` as `policy_test.rego`.
```

docs/how_to_write_packages.md
```diff
@@ -101,19 +102,19 @@ package pods
```
wwwil: I can't add a comment on the correct line, but above this snippet it says "rule in a package named `podsbestpractices`" but the snippet then declares `package pods`.

charlieegan3: Should be fixed in f6dc60a27601b88944a377648df345d29783b74f

preflight-packages/examples.jetstack.io/gke_basic/gke.rego
```diff
@@ -4,42 +4,50 @@ package gke_basic
...
+
+	message := "cluster has legacy abac enabled"
 }
 
 # Rule 'k8s_master_up_to_date'
-default preflight_k8s_master_up_to_date = false
-preflight_k8s_master_up_to_date {
-        re_match(`^1\.1[234].*$`, gke.currentMasterVersion)
+preflight_k8s_master_up_to_date[message] {
+	not gke.currentMasterVersion
```
wwwil: I think `currentMasterVersion` will always have a value. This was using regex to check it started `1.12`, `1.13` or `.14`. Probably needs updating but something we should check.

Makefile
j-fuentes: Yes, since now we have #29 merged, packages can be tested with `preflight package test <path to pkg>`.

preflight-packages/examples.jetstack.io/gke_basic/gke.rego
```diff
@@ -4,42 +4,50 @@ package gke_basic
...
+
+	message := "cluster master version is missing"
 }
+preflight_k8s_master_up_to_date[message] {
+	not re_match(`^1\.1[23467].*$`, gke.currentMasterVersion)
```
wwwil: `1.13` is now default so I think we should also remove `1.12` from this list. This may also require an update to the `policy_manifest.yaml`.

preflight-packages/examples.jetstack.io/gke_basic/gke.rego
```diff
@@ -4,42 +4,50 @@ package gke_basic
...
 }
-node_pools_old_version[name] {
-        np := gke.nodePools[_]
-        name := np.name
-        not node_pools_current_version[name]
+
+# Rule 'k8s_nodes_up_to_date'
+preflight_k8s_nodes_up_to_date[message] {
+	node_pool := gke.nodePools[_]
+	not re_match(`^1\.1[234567].*$`, node_pool.version)
```
wwwil: Probably also remove `1.12` here too.

preflight-packages/examples.jetstack.io/gke_basic/gke.rego
charlieegan3: So this is ok to leave as is? The old tests seem to account for this so I left it in.

Makefile
j-fuentes: You can use `preflight package test` here if you want, and I'd do that, but in this case, I am not sure if it is better to do:

```
package-test: <whatever target is needed to have preflight binary in place>
  preflight package test $(ROOT_DIR)/preflight-packages/examples.jetstack.io
```

or

```
package-test:
  go run ./main.go package test $(ROOT_DIR)/preflight-packages/examples.jetstack.io
```

BTW, if you proceed to do this, double-check that `preflight package test` supports a directory with packages as argument instead of the path to a package. In case it doesn't support it, I think we should make it support it.

preflight-packages/examples.jetstack.io/gke_basic/gke.rego
charlieegan3: Should be fixed in a5e1b0759f9c6142a595de53043849d34b63f9cf

preflight-packages/examples.jetstack.io/gke_basic/gke.rego
charlieegan3: Should be fixed in a5e1b0759f9c6142a595de53043849d34b63f9cf

Makefile
charlieegan3: I've switched to this in ca3dc849bbe57dae2e2f4d2a295be672f53ae2a4

preflight-packages/examples.jetstack.io/pods_basic/pods.rego
```diff
@@ -3,67 +3,74 @@ package pods
...
-default preflight_explicit_image_tag = false
-preflight_explicit_image_tag {
-        count(containers_without_explicit_tag) == 0
+preflight_explicit_image_tag[message] {
+	# find all containers in all pods
+	pod := pods.items[_]
+	container := pod.spec.containers[_]
+	# validate that the image value contains an explicit tag
+	{ re_match(`latest$`, container.image),
+	re_match(`^[^:]+$`, container.image) } & { true } != set()
```
wwwil: I'm slightly confused by this, why is the ` & { true }` required?

charlieegan3: This is just a compact way to write OR in rego... :grimacing: 

https://stackoverflow.com/a/58325979/1510063 outlines some other options

charlieegan3: & is intersection, so { true, x } & { true } == { true }

Which is not empty, and so the message is bound as that's a violation as one of the image requirements was not met

charlieegan3: We could use incremental rules if you'd rather.

preflight-packages/examples.jetstack.io/gke_basic/gke.rego
wwwil: I don't think this would ever not have a value, the old test may have just been unnecessary. https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.FIELDS.current_master_version

charlieegan3: Sounds good, I've removed in e21905e0929073018328ea4b91ea106d0d4ad16a

preflight-packages/examples.jetstack.io/pods_basic/pods.rego
wwwil: > & is intersection

Now it makes sense, I was being slow and thinking of it as a binary thing.

preflight-packages/examples.jetstack.io/pods_basic/pods.rego
```diff
@@ -3,67 +3,74 @@ package pods
...
+
+	# bind a message for reporting
+	message := sprintf("init container '%s' in pod '%s' in namespace '%s' is missing an explicit image tag", [container.name, pod.metadata.name, pod.metadata.namespace])
 }
-containers_without_explicit_tag[container_name] {
-        pod := pods.items[_]
-        container := pod.spec.containers[_]
-        container_name = format_container(pod, container)
-        not explicit_tag(container)
-}
```
wwwil: No newline at end of file.

preflight-packages/examples.jetstack.io/pods_basic/pods.rego
```diff
@@ -3,67 +3,74 @@ package pods
...
-        count(init_containers_without_cpu_limits) == 0
-}
-format_container(pod, container) = name {
-        name := {"namespace": pod.metadata.namespace,
-                 "pod": pod.metadata.name,
-                 "container": container.name}
-}
-cpu_limit_unset(container) {
-        not container.resources.limits.cpu
+format_container(pod, container) = data {
```
wwwil: I don't think this is used any more.

preflight-packages/examples.jetstack.io/pods_basic/pods.rego
charlieegan3: I think there is now though right?

charlieegan3: This is the old content.

preflight-packages/examples.jetstack.io/pods_basic/pods.rego
charlieegan3: Ahh yes, I forgot to clean that out. Fixed in db036d556961bafbb5e1393f4b7991d9f048bc80

preflight-packages/examples.jetstack.io/pods_basic/pods.rego
charlieegan3: If we have more queries about it I might move to more verbose incremental rules - they have the added benefit of more descriptive error messages too.

preflight-packages/examples.jetstack.io/pods_basic/pods.rego
wwwil: Ah yeah, GitHub diff is saying you've removed the lack of newline at the end of the file. So all good!

wwwil:
This all looks good to me, but as it's quite a large change I'm going to wait for @j-fuentes to approve before `lgtm`ing it.

charlieegan3:
Thanks Wil, that's fair :wink: - Jose is having a look too.

pkg/results/results.go
```diff
@@ -105,9 +99,13 @@ func NewResultCollectionFromRegoResultSet(rs *rego.ResultSet) (*ResultCollection
 
 	rc := make(ResultCollection, 0, len(keys))
 	for _, k := range keys {
+		value, ok := values[k].([]string)
```
j-fuentes: This breaks backward compatibility with the existing packages.

I think it is possible to add the feature while keeping b/c:

If order to do so, I'd make `NewResultCollectionFromRegoResultSet` attempt to read the value as a `bool` (the behaviour we are trying to deprecate). If it reads a bool successfully, print a "this is deprecated bleh bleh bleh" message and build the Result object creating a `[]string` that would be either `[]string{}` if the bool was `true` (previously meaning success), or `[]string{"missing context"}` (or something like that) in case the bool was `false`.

pkg/results/results.go
```diff
@@ -12,25 +12,23 @@ import (
...
 )
 
-type outputResult map[string]interface{}
+type outputResult map[string][]string
 
 // Result holds the information about the result of a check
 type Result struct {
 	ID      string
-	Value   interface{}
+	Value   []string
```
j-fuentes: I'd rather introduce a `Success bool` property and a `Violations []string` one, and deprecate `Value`. Although I'd keep `Value` in place, but documented as deprecated.

charlieegan3: Would you rather a `Success bool` over just adding Violations and having the Success state depend on the presence of Violations?

pkg/results/results.go
charlieegan3: I've had a go at this in d116daf9015a14da21ab7a00cc72ab205aacab3b - wdyt? 

I also opted to leave out a `Success` bool for now (as per https://github.com/jetstack/preflight/pull/38/files/d9e42e8d1fa65447295e4f156526fadf08d11bf3#r365223255)

pkg/results/results.go
j-fuentes: We can omit the `Success bool`, I'm fine with that. I'd keep the `IsSuccessState` and `IsFailureState` functions.

charlieegan3: :+1: I've kept those and based them on whether or not the violations list is empty or not.

charlieegan3:
I've brought support for Value back in here: 489f2c6e3ada298dc62c52a4f1649863be6276ef - it maintains the old behavior but uses Violations instead if it's set.

j-fuentes:
Great job @charlieegan3  🙌

charlieegan3:
Thanks guys :blush:

PR #46, opened 2020-01-14
Change how results are unmarshaled from rego

j-fuentes:
We were assuming the results came from rego as `[]string`, but they were begin returned as `[]interface{}`. The logic was detecting those as invalid and erroring as a consequence.


charlieegan3:
Ahh, very nice spot. :clap: 

/lgtm
/approve

PR #47, opened 2020-01-17
Add pods package

wwwil:
This PR moves the previously private `pods` package to be open source, and updates the rule ids to use the new convention.


charlieegan3:
I'd have this replace https://github.com/jetstack/preflight/tree/master/preflight-packages/examples.jetstack.io/pods_basic - personally.

wwwil:
> I'd have this replace https://github.com/jetstack/preflight/tree/master/preflight-packages/examples.jetstack.io/pods_basic - personally.

I've removed the `pods_basic` package, intended to do that but forgot. I think as this is a full package rather than a reduced version of a private package it should have a different namespace though.

preflight-packages/jetstack.io/pods/pods.rego
```diff
@@ -0,0 +1,188 @@
+package preflight.pods
```
wwwil: Should we still have the `preflight` prefix in the Rego package name?

charlieegan3: Do we need it? If not my vote would be to remove - but we still have it in our other policies... might be better to remove it all at once - over all packages.

charlieegan3:
> > I'd have this replace https://github.com/jetstack/preflight/tree/master/preflight-packages/examples.jetstack.io/pods_basic - personally.
> 
> I've removed the `pods_basic` package, intended to do that but forgot. I think as this is a full package rather than a reduced version of a private package it should have a different namespace though.

Yup, agreed, it doesn't make sense to call this pods_basic any more.

preflight-packages/jetstack.io/pods/pods.rego
j-fuentes: it is not needed, just drop it from the `package` statement in the rego and from the `root-query` in the policy manifest.

wwwil: Done!

PR #51, opened 2020-01-20
Add lint rule, update test rule, add both to CI test rule

wwwil:
As we add more full packages to the repo we should also be running lint and test on them in CI to ensure they are correctly formatted and working.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,8 +71,15 @@ bundle-all-platforms:
 
 # Packages
 
-package-test:
-	cd $(ROOT_DIR) && go run ./main.go package test ./preflight-packages/examples.jetstack.io
+packages-lint:
+	cd $(ROOT_DIR) && \
+	go run . package lint $(ROOT_DIR)/preflight-packages/jetstack.io/pods && \
+	go run . package lint $(ROOT_DIR)/preflight-packages/examples.jetstack.io/aks_basic && \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -71,8 +71,15 @@ bundle-all-platforms:
 
 # Packages
 
-package-test:
-	cd $(ROOT_DIR) && go run ./main.go package test ./preflight-packages/examples.jetstack.io
+packages-lint:
+	cd $(ROOT_DIR)/.cache/cli/repo && \
+	go run . package lint $(ROOT_DIR)/preflight-packages/jetstack.io/pods && \
```
j-fuentes: We should make `package lint` match the behaviour of `package test` and make it search for packages recursively.

j-fuentes: (not necessarily as part of this PR)

Makefile
```diff
@@ -96,7 +103,7 @@ export PATH:=$(GOPATH)/bin:$(PATH)
 ci-deps:
 	go install golang.org/x/lint/golint
 
-ci-test: ci-deps test lint
+ci-test: ci-deps test lint package-test package-lint
```
j-fuentes:
```suggestion
ci-test: ci-deps test lint packages-test packages-lint
```

PR #52, opened 2020-01-21
Add Azure Blob Storage option for the output

### LINKED ISSUE #49: Support for Azure Blob Storage
I would like to be able to store reports in Azure Blob Storage.

j-fuentes:
This adds a new option for the output, Azure Blob Storage (`azblob`).

Closes #49.


examples/aks.preflight.yaml
```diff
@@ -16,9 +16,9 @@ enabled-packages:
 
 # This configures how the results will be reported.
 outputs:
-# writes a human readable doc that can be opened with a browser.
-- type: local
-  path: ./output
+# uploads a json file to Azure Blob Storage
+- type: azblobb
```
charlieegan3: `azblobb` -> `azblob`?

charlieegan3:
```suggestion
- type: azblob
```

j-fuentes: good catch!

charlieegan3:
/approve

I have not tested myself but this looks simple enough.

cmd/check.go
```diff
@@ -243,14 +244,43 @@ func check() {
...
+		} else if outputType == "azblob" {
+			outputFormat, ok := outputDefinition["format"].(string)
+			if !ok {
+				log.Fatal("Missing 'format' property in azblob output configuration.")
+			}
+			outputContainer, ok := outputDefinition["container"].(string)
+			if !ok {
+				log.Fatal("Missing 'container' property in azblob output configuration.")
+			}
+			accountName, accountKey := os.Getenv("AZURE_STORAGE_ACCOUNT"), os.Getenv("AZURE_STORAGE_ACCESS_KEY")
```
wwwil: Can we use a key file here like we do for GKE/AKS access?

j-fuentes: I looked into that briefly and didn't found an immediate well-known way of doing that, so I decided to use the env vars.

I can investigate that later.

PD: the token we use for AKS is not ideal. It is very short-lived and I haven't found a good way to extend its life. That is another thing to look into.

I have opened another issue to improve how we authenticate against Azure in general: #56

examples/aks.preflight.yaml
```diff
@@ -16,9 +16,9 @@ enabled-packages:
 
 # This configures how the results will be reported.
 outputs:
-# writes a human readable doc that can be opened with a browser.
-- type: local
-  path: ./output
+# uploads a json file to Azure Blob Storage
```
wwwil: Could you also add a note here that the environment variables have to be set, and link to somewhere that explains how to do that?

examples/aks.preflight.yaml
```diff
@@ -16,9 +16,9 @@ enabled-packages:
...
+- type: azblob
+  container: mycontainer
```
wwwil: Possibly also note that the container must already exist.

examples/aks.preflight.yaml
j-fuentes: Added to `./docs/configuration.md`.

examples/aks.preflight.yaml
j-fuentes: Added to `./docs/configuration.md`.

docs/configuration.md
```diff
@@ -107,13 +105,62 @@ Possible formats are:
...
+
+With the `cli` type output the format is optional
+and defaults to the `cli` format, for a coloured CLI formatted report.
+
+### local
+
+The `local` type output writes the results to a local directory.
+The directory must be specified with the `path` property.
+It will be created in case it does not exist.
+
```
wwwil: Nit: Extra newline.

Issue #27, opened 2019-12-23
Remove the artificial prefix `preflight_` in the rego rules


j-fuentes:
Preflight assumes there is some conversion between the rule IDs in the Policy Manifest and the REGO files.

This conversion is needed because rule IDs in rego have certain limitations.

But also, we have introduced the `preflight_` prefix as part of this convention. In my opinion, this is very artificial and not necessary at all. I think this was introduced in the early times of Preflight to distinguish which rules were actually Preflight rules and which ones were just helpers. This is no longer necessary, as the Policy Manifest is the source of truth for what is a Preflight rule.

j-fuentes:
Notice we might need to increase the `schema-version` of the packages if we do this.

charlieegan3:
Does this need to be updated anywhere other than the gke and pod basics? (if not, I can probably do this as part of #36 )

j-fuentes:
> Does this need to be updated anywhere other than the gke and pod basics? (if not, I can probably do this as part of #36 )

This needs to be updated in all the packages, so everything under `preflight-packages` directory, and also in the [documentation](https://github.com/jetstack/preflight/blob/master/docs/how_to_write_packages.md#the-rego-package).

wwwil:
It has already been removed in [some places](https://github.com/jetstack/preflight/blob/master/preflight-packages/examples.jetstack.io/aks_basic/aks.rego#L7) but the linter hasn't complained which is interesting... Still needs to be removed from `gke_basic`.

j-fuentes:
This is done already.

PR #55, opened 2020-01-23
Update GKE basic package

wwwil:
This PR updates the GKE basic package to remove the `preflight_` prefix from rules and match its rules with the corresponding rules in the more comprehensive private GKE package.

Related to #27.


j-fuentes:
Could you please update `schema-version` to 0.1.1?

PR #59, opened 2020-01-23
Handle the existence or not of the `preflight_` prefix in rego

j-fuentes:
This allows the `preflight_` to exist or not if the schema version of the package is lower than "0.1.0". It assumes the artificial prefix does not exist if the schema version is higher than "0.1.0".

Related: #27

### DIFF
--- a/pkg/exporter/exporter.go
+++ b/pkg/exporter/exporter.go
@@ -32,9 +32,13 @@ const (
 )
 
 func ruleToResult(ruleID string) string {
-	return fmt.Sprintf("preflight_%s", strings.ReplaceAll(ruleID, ".", "_"))
+	return strings.ReplaceAll(ruleID, ".", "_")
 }
 
 func resultToRule(resultID string) string {
-	return strings.TrimPrefix(strings.ReplaceAll(resultID, "_", "."), "preflight.")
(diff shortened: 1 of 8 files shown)


charlieegan3:
/approve 

Can lgtm too but was going to let @wwwil have a look.

pkg/exporter/exporter.go
```diff
@@ -32,9 +32,13 @@ const (
 )
 
 func ruleToResult(ruleID string) string {
-	return fmt.Sprintf("preflight_%s", strings.ReplaceAll(ruleID, ".", "_"))
+	return strings.ReplaceAll(ruleID, ".", "_")
```
wwwil: Isn't this unnecessary now as rule IDs are required to be valid Rego names?

j-fuentes: They are not strictly required though.

We check that in the linter, which flags recommendations but, at least for now, we don't impose those rules.

If at some point we want to impose them, we would run those checks at the beginning of the execution of Preflight.

wwwil: Yeah, I feel that we should be enforcing it. I think packages should be linted when they are loaded as part of a `preflight check` and packages that fail should be ignored. This PR is probably not the place to make that change though.

Issue #54, opened 2020-01-22
Index report bucket contents


charlieegan3:
Currently, Preflight will export data to a bucket in the following 'directory' structure:

```
clusterA:
  timestampA:
    packageA.json
    packageB.json
  timestampB:
    ...
clusterB:
  ... 
```

This bucket is queried when the data is loaded for display in the frontend when it is loaded buy the (closed source) preflight backend.

The issues with this format is that all the reports must be listed to find the latest (since we can't order the results from object storage).

Let's change the format to make it possible to find the latest report in constant time.

Goals:
- minimize the complexity in the writes to object storage
- create a structure which helps minimize the complexity in the reads of the same data
- create a structure which allows the latest report for each cluster to be found and loaded quickly
- avoid duplicated data in the bucket
- never delete any data from the bucket

Non goals:
- pagination of results for a single cluster's reports (this can happen at a later date)

charlieegan3:
My proposal when writing a report, we also update a file for a cluster at `index/cluster_name.json` with the following content:

```json
{
  "clusterName": "example",
  "latest_report_timestamp": "string timestamp"
}
```

Where the timestamp is the same as the one most recently used for the writing of the reports.

j-fuentes:
I like the general idea of this proposal.

I would like to suggest some changes to the proposed `index/cluster_name.json` file.

We need to take into account that the "status" of the cluster is not the latest report, but the aggregation of the latest reports for every package. Let's call that a `ReportSet`.

Fortunately, we already have [some types defined](https://github.com/jetstack/preflight/blob/master/api/cluster.go) for the type of information we want to store.

charlieegan3:
I was being pretty casual with my use of the word Report and was using it to mean a ReportSet. Is it possible to have the latest gke report and the latest pod report in a different ReportSet?

j-fuentes:
>  Is it possible to have the latest gke report and the latest pod report in a different ReportSet?

It is possible, but why would you do that? A `ReportSet` groups reports from the same moment, i.e. reports generated in the same preflight run.

> I suppose we could make the index/cluster_name.json file a dump of the ClusterSummary but then it's more of a recent copy than an index.

It is just an index with some (cheap) metadata. It is not a copy of the full report. Note that `ClusterSummary` contains `LatestReportSet` and a `ReportSet.Reports` is just a slice of `ReportSummary`, which is not the full `Report` but just some metadata.

charlieegan3:
[...]

Issue #1, opened 2019-11-11
Split "exporters" into "matchers" and "exporters"


j-fuentes:
The current "exporters" put together evaluation results and metadata and then render the result in a certain format.

The process of putting together results and metadata should be decoupled from rendering that "pack" into different formats. Here is where the "matcher" + "exporter" pattern comes into place.

As part of this change, the interface for the exporters needs to change to accept a Report struct containing both the result and the metadata, instead of receiving the policy manifest and a result collection as it does now.

charlieegan3:
I'm not familiar with this pattern, but is a matcher that translates result collections and policy manifests into reports ever going to work in more than one way? i.e. will there be more than one matcher?

wwwil:
Ah, I had forgotten about this issue, hence [my comment](https://github.com/jetstack/preflight/pull/63#issuecomment-578777847) on Charlie's PR.

If I understand correctly the idea here is that after Preflight performs evaluation it creates a report from the results and policy manifest, and this report is then used to generate all other forms of output.

charlieegan3:
> and this report is then used to generate all other forms of output.

This makes sense to me - we'd need to change the interface for the exporters though.

charlieegan3:
Closing as this will need to be addressed in a different way with the agent/backend architecture.

/close

PR #63, opened 2020-01-27
Extract report generation to reports package

charlieegan3:
This will allow me to generate a cluster summary and update that too
when writing output to a bucket destination.

Before this, the report generation all happened in the json exporter
which meant the report data structure wasn't available for reuse.

Done as part of https://github.com/jetstack/preflight/issues/54

Related to https://github.com/jetstack/preflight/issues/1

### DIFF
--- a/pkg/reports/reports.go
+++ b/pkg/reports/reports.go
@@ -8,8 +8,8 @@ import (
 	"github.com/jetstack/preflight/pkg/version"
 )
 
-// ConstructReport creates a report from a policy manifest and a results collection
-func ConstructReport(pm *packaging.PolicyManifest, rc *results.ResultCollection) (api.Report, error) {
+// NewReport creates a report from a policy manifest and a results collection
+func NewReport(pm *packaging.PolicyManifest, rc *results.ResultCollection) (api.Report, error) {
 	report := api.Report{
 		// TODO: we are omitting ID, Timestamp and Cluster for now, but it will get fixed with #1
 		PreflightVersion: version.PreflightVersion,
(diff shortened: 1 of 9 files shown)


wwwil:
Looking over this has reminded me that the other output formats are generated directly from results and a policy manifest. I feel it would be better to actually produce a report, then generate other formats of output from that. Though that can be dealt with in a separate issue.

pkg/reports/reports.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"github.com/jetstack/preflight/api"
+	"github.com/jetstack/preflight/pkg/packaging"
+	"github.com/jetstack/preflight/pkg/results"
+	"github.com/jetstack/preflight/pkg/rules"
+	"github.com/jetstack/preflight/pkg/version"
+)
+
+// ConstructReport creates a report from a policy manifest and a results collection
+func ConstructReport(pm *packaging.PolicyManifest, rc *results.ResultCollection) (api.Report, error) {
```
wwwil: Is there any reason this is called `ConstructReport` rather than `Create Report`?

charlieegan3: I guess I'm seeing this as a replacement for an oop constructor method.

create works too.

I'm wondering if `NewReport` might be more idiomatic :man_shrugging:

charlieegan3: I've renamed it to new in the upcoming changes I was preparing to add on top here https://github.com/charlieegan3/preflight/pull/2

charlieegan3: Brought that in for this branch here: b47c45d70086844584894c93a1f17fc0d318a8a1

charlieegan3:
Thanks Wil.

PR #65, opened 2020-01-27
Add rules for node and pod affinity

wwwil:

### DIFF
--- a/preflight-packages/jetstack.io/pods/policy-manifest.yaml
+++ b/preflight-packages/jetstack.io/pods/policy-manifest.yaml
@@ -298,22 +298,3 @@ sections:
       the linked documentation.
     links:
     - "https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity"
-  - id: pod_affinity_used
-    name: Pod affinity used
-    description: >
-      In some cases it may be preferable or required for a Pod to run on the
-      same node as another Pod. This may be due to a shared requirement or
-      dependency. In other cases it may be preferable or required for a Pod not
-      to run on the same node as another Pod. This may be for greater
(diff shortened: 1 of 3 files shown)


wwwil:
I've added two new rules to the policy manifest, if someone could confirm they are sensible I will add some Rego to check them!

preflight-packages/jetstack.io/pods/policy-manifest.yaml
```diff
@@ -277,3 +277,43 @@ sections:
...
+      limited constraint. Instead it's recommended that `nodeAffinity` is used
+      as it greatly expands the types of constraints you can express. There are
+      two types of node affinity; 
+      `requiredDuringSchedulingIgnoredDuringExecution` which specifies a hard
+      requirement, and `preferredDuringSchedulingIgnoredDuringExecution` which
+      specifies a preference. Full details of the constraints can be found in
+      the linked documentation.
+    links:
+    - "https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity"
+  - id: pod_affinity_used
```
charlieegan3: How should this rule be used? e.g. nodeSelector triggers the other one - what triggers this?

wwwil: I was planning to trigger it if no Pods used `podAffinity` or `podAntiAffinity`. This would really require the ability to ignore checks (which I plan to work on next) and might be a bit 'overbearing'.

charlieegan3: Yeah, I might split this and the node selector one tbh. It seems to me like we should be weary of introducing checks that we ourselves would disable by default...

PR #66, opened 2020-01-27
Report index constructors

charlieegan3:
Being done as part of #54

Few things I'm not sure about here:
- if a single test for generating a clustersummary would be enough - there is a lot of duplication
- if some functions such as summarize report should be registered on the report struct.

### DIFF
--- a/pkg/reports/reports.go
+++ b/pkg/reports/reports.go
@@ -10,6 +10,24 @@ import (
 	"github.com/jetstack/preflight/pkg/version"
 )
 
+// ConstructClusterSummary builds a summary for a current cluster based on a
+// freshly generated report set
+func ConstructClusterSummary(reports []api.Report) (api.ClusterSummary, error) {
+	if len(reports) < 1 {
+		return api.ClusterSummary{}, fmt.Errorf("you must supply at least one report")
+	}
+
+	reportSet, err := ConstructReportSet(reports)
+	if err != nil {
+		return api.ClusterSummary{}, fmt.Errorf("error constructing report set: %v", err)
+	}
+
+	return api.ClusterSummary{
+		Cluster:         reportSet.Cluster,
+		LatestReportSet: &reportSet,
+	}, nil
+}
+
 // ConstructReportSet generates a summarized report set from the supplied reports
 func ConstructReportSet(reports []api.Report) (api.ReportSet, error) {
 	if len(reports) < 1 {
--- a/pkg/reports/reports.go
+++ b/pkg/reports/reports.go
@@ -1,12 +1,39 @@
 package reports
 
 import (
+	"fmt"
+
 	"github.com/jetstack/preflight/api"
 	"github.com/jetstack/preflight/pkg/packaging"
 	"github.com/jetstack/preflight/pkg/results"
 	"github.com/jetstack/preflight/pkg/rules"
 	"github.com/jetstack/preflight/pkg/version"
 )
+
+// ConstructReportSet generates a summarized report set from the supplied reports
+func ConstructReportSet(reports []api.Report) (api.ReportSet, error) {
+	if len(reports) < 1 {
+		return api.ReportSet{}, fmt.Errorf("you must supply at least one report")
+	}
+
+	reportSet := api.ReportSet{
+		Cluster:   reports[0].Cluster,
+		Timestamp: reports[0].Timestamp,
+		Reports:   []*api.ReportSummary{},
+	}
+
+	for _, report := range reports {
+		summary := SummarizeReport(report)
+		reportSet.Reports = append(reportSet.Reports, &summary)
+	}
+
+	for _, summary := range reportSet.Reports {
+		reportSet.SuccessCount += summary.SuccessCount
+		reportSet.FailureCount += summary.FailureCount
+	}
+
+	return reportSet, nil
+}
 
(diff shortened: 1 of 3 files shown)


pkg/reports/reports.go
```diff
@@ -1,13 +1,81 @@
...
+
 	"github.com/jetstack/preflight/api"
 	"github.com/jetstack/preflight/pkg/packaging"
 	"github.com/jetstack/preflight/pkg/results"
 	"github.com/jetstack/preflight/pkg/rules"
 	"github.com/jetstack/preflight/pkg/version"
 )
 
+// NewClusterSummary builds a summary for a current cluster based on a
+// freshly generated report set
```
j-fuentes:
```suggestion
// freshly generated slice of reports.
```

j-fuentes: Just to avoid confusion with the `ReportSet` type.

pkg/reports/reports.go
```diff
@@ -1,13 +1,81 @@
...
+	}
+
+	return api.ClusterSummary{
+		Cluster:         reportSet.Cluster,
+		LatestReportSet: &reportSet,
+	}, nil
+}
+
+// NewReportSet generates a summarized report set from the supplied reports
+func NewReportSet(reports []api.Report) (api.ReportSet, error) {
```
j-fuentes: A `ReportSet` is intended to contain summaries of reports with **the same timestamp**.

```
// ReportSet groups one or more reports of different packages with the same timestamp for the same cluster.
```

We will use this with the latest set of reports.

Who is going to be responsible for extracting the latest ones? If it is not this function, I'd document that this assumes the reports have the same timestamp. I'd recommend to check them, just to make sure we are not misusing it.

Another option is to make `NewReportSet` extract from the slice of reports the latest set of reports.

j-fuentes:
> if a single test for generating a clustersummary would be enough - there is a lot of duplication
What do you mean? What is duplicated?

j-fuentes:
> if some functions such as summarize report should be registered on the report struct.

Yes, +1 to `report.Summarize()`.

pkg/reports/reports.go
charlieegan3: I've had a go at this in 7dbad8df549194541a0c70afc1c9d6b8d85856e6 - will try and dry up the test data for all these later.

charlieegan3:
> > if a single test for generating a clustersummary would be enough - there is a lot of duplication
> 
> What do you mean? What is duplicated?

The test data of the reports is duplicated a few times in `reports_test`. I can likely find a way to dry this up though.

charlieegan3:
> > if some functions such as summarize report should be registered on the report struct.
> 
> Yes, +1 to `report.Summarize()`.

I've done this in 2e8fb4a5ab3fee9e1da613e49f105e18d0ad0589

charlieegan3:
> > if a single test for generating a clustersummary would be enough - there is a lot of duplication
> 
> What do you mean? What is duplicated?

I've dried up the test for reports in e317f456972feca610e877fae4aafcfdc5e235f3

j-fuentes:
> > > if a single test for generating a clustersummary would be enough - there is a lot of duplication
> > 
> > 
> > What do you mean? What is duplicated?
> 
> The test data of the reports is duplicated a few times in `reports_test`. I can likely find a way to dry this up though.

We can create factories for that. It is a common pattern in some languages but I haven't seen that too much in Go since we try to create smaller and more specific tests.

pkg/reports/reports.go
```diff
@@ -1,13 +1,74 @@
...
+
+	for _, report := range reports {
+		summary := report.Summarize()
+		reportSet.Reports = append(reportSet.Reports, &summary)
+	}
+
+	for _, summary := range reportSet.Reports {
+		reportSet.SuccessCount += summary.SuccessCount
+		reportSet.FailureCount += summary.FailureCount
+	}
```
j-fuentes: nit: instead of going through all the reports three times, you can have just one loop:

```
reportSet := api.ReportSet{
  Cluster:   reports[0].Cluster,
  Timestamp: reports[0].Timestamp,
  Reports:   []*api.ReportSummary{},
}

for _, report := range reports {
  if report.Cluster != reportSet.Cluster {
    return nil, fmt.Errorf("reports must be for the same cluster")
  }
  if report.Timestamp != reportSet.Timestamp {
    return nil, fmt.Errorf("reports must be for the same timestamp")
  }

  summary := report.Summarize()
  reportSet.Reports = append(reportSet.Reports, &summary)

  reportSet.SuccessCount += summary.SuccessCount
  reportSet.FailureCount += summary.FailureCount
}
```

charlieegan3:
> > > > if a single test for generating a clustersummary would be enough - there is a lot of duplication
> > > 
> > > 
> > > What do you mean? What is duplicated?
> > 
> > 
> > The test data of the reports is duplicated a few times in `reports_test`. I can likely find a way to dry this up though.
> 
> We can create factories for that. It is a common pattern in some languages but I haven't seen that too much in Go since we try to create smaller and more specific tests.

Indeed, I have also not used this in go... https://github.com/thoughtbot/factory_bot used to be our go to in Ruby land.

pkg/reports/reports.go
charlieegan3: Good shout, should be fixed in 0e58fd36ac6cb56a2b2da16d090250772af70f4d

Issue #64, opened 2020-01-27
Config file schema


wwwil:
I feel that it would be better if we defined a Go struct for the Preflight config and un-marshalled YAML into this, rather than using `viper.GetString` throughout the code.

This will make it clearer what the structure of the config should be, and make it clearer when it changes. It also means that the config can be checked more before Preflight actually runs anything.

This would also allow us to add a command to check a config file, if we felt this was useful.

AkvileMar:
it's ready for a review.

j-fuentes:
This is the related PR: https://github.com/jetstack/preflight/pull/82

AkvileMar:
Pull requests are ready to go. It should be merged by the end of today.

wwwil:
Yesterday @j-fuentes PRs #105 and #106 were merged. My PR #107 was merged and planning to finish #104 today.

wwwil:
#104 has been merged, #112 is waiting for review.

#112 tidies up the config loading in `check.go` but does not fully resolve this issue.

j-fuentes:
@wwwil what is missing after #112?

I think it is not worth investing more in this since we are going to transition to the agent soon.

wwwil:
We're still decoding the config from file field by field, rather than unmarshalling directly into a struct.

But I agree that it's not really worth spending more time on this now.

PR #67, opened 2020-01-27
Allow disabling of rules in a package

### LINKED ISSUE #42: Allow disabling of checks in a package
We have [packages](https://github.com/jetstack/preflight#preflight-packages) of checks that preflight can run. While these are intended to be used in full - sometimes checks will need to be disabled.

## Preflight yaml changes

It makes sense to me to configure this in one of the following ways:

**Namespaced List of Excludes**

```yaml
enabled-packages:
  - "examples.jetstack.io/gke_basic"

ignored-checks:
  - "examples.jetstack.io/gke_basic/networking/private_cluster"
```

This is simple and likely most simple to implement but I don't like how the package names and namespaces are duplicated.

**Checks listed under package**

```yaml
enabled-packages:
- name: "examples.jetstack.io/gke_basic"
  disabled-checks:
  - private_cluster
```

This appears to be more concise but requires a change in the data format for packages making it marginally more complex to implement.

**Whitelist/Blacklist**

```yaml
enabled-packages:
- name: "examples.jetstack.io/gke_basic"
  enabled-checks: # whitelist
  - private_cluster
  disabled-checks: # blacklist
  - master_auth
```

Where both fields are optional - if both present, use the whitelist. This is marginally more complex to implement but makes it easier to use a single check from a package (using the whitelist).

## Changes to reports

The excluded checks are not only absent from the report but are also not run. This is important to avoid running checks where the results are not important to the user.

wdyt @j-fuentes @wwwil ?

wwwil:
Closes #42.

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -21,6 +21,7 @@ import (
 	"github.com/jetstack/preflight/pkg/packagesources/local"
 	"github.com/jetstack/preflight/pkg/packaging"
 	"github.com/jetstack/preflight/pkg/reports"
+	"github.com/jetstack/preflight/pkg/results"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -338,23 +339,7 @@ func check() {
 			input[dg] = information[dg]
 		}
 
-		// Determine which rules should be enabled
-		enabledRuleIDs := map[string]bool{}
-		ruleEnabledDefault := true
-		if len(enabledPackage.EnabledRules) != 0 {
-			ruleEnabledDefault = false
-		}
-		for _, ruleID := range manifest.RuleIDs() {
-			enabledRuleIDs[ruleID] = ruleEnabledDefault
-		}
-		for _, enabledRule := range enabledPackage.EnabledRules {
-			enabledRuleIDs[enabledRule] = true
-		}
-		for _, disabledRule := range enabledPackage.DisabledRules {
-			enabledRuleIDs[disabledRule] = false
-		}
-
-		rc, err := packaging.EvalPackage(ctx, pkg, input, enabledRuleIDs)
+		rc, err := packaging.EvalPackage(ctx, pkg, input)
 		if err != nil {
 			if _, ok := err.(*reports.MissingRegoDefinitionError); ok {
 				missingRules = true
@@ -363,6 +348,8 @@ func check() {
 				log.Fatalf("Cannot evaluate package %q: %v", manifest.ID, err)
 			}
 		}
+
+		rc = results.FilterResultCollection(rc, enabledPackage.DisabledRules, enabledPackage.EnabledRules)
 
 		intermediateBytes, err := json.Marshal(input)
 		if err != nil {
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -339,7 +339,16 @@ func check() {
 			input[dg] = information[dg]
 		}
 
-		rc, err := packaging.EvalPackage(ctx, pkg, input, enabledPackage.DisabledRules)
+		// Determine which rules should be enabled
+		enabledRuleIDs := map[string]bool{}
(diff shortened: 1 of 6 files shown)


cmd/check.go
```diff
@@ -300,18 +300,27 @@ func check() {
 		outputs = append(outputs, op)
 	}
 
-	// Loop over enabled packages and evaluate.
-	enabledPackages := viper.GetStringSlice("enabled-packages")
+	type EnabledPackage struct {
+		Name          string
+		EnabledRules  []string `mapstructure:"enabled-rules"`
+		DisabledRules []string `mapstructure:"disabled-rules"`
+	}
```
j-fuentes: This breaks backward compatibility with the existing configuration files, isn't it?

If that is the case, I'd rather keep `enabled-packages` as it is, and have a separate section in the configuration for the whitelist and the blacklist.

j-fuentes: Something like:

```
enabled-packages:
  - "jetstack.io/pods"
  - "jetstack.io/services"
rule-selectors:
  - package: "jetstack.io/pods"
    enabled-rules:
      - rule_1
  - package: "jetstack.io/services"
   disabled-rules:
      - rule_2
```

cmd/check.go
```diff
@@ -328,7 +337,23 @@ func check() {
...
+			enabledRuleIDs[ruleID] = ruleEnabledDefault
+		}
+		for _, enabledRule := range enabledPackage.EnabledRules {
+			enabledRuleIDs[enabledRule] = true
+		}
+		for _, disabledRule := range enabledPackage.DisabledRules {
+			enabledRuleIDs[disabledRule] = false
+		}
+
+		rc, err := packaging.EvalPackage(ctx, pkg, input, enabledRuleIDs)
```
j-fuentes: I'd avoid dealing with the filter during the evaluation at all.

I'd rather filter the `ResultCollection` after the evaluation. That will keep the `packaging` package completely agnostic from this change, which is good for simplicity.

Something like this would be my preference:
```
type Filter struct {
  Package string
  EnabledRules []string
  DisabledRules []string
}

// FilterResults filters the results using a filter. If filter is nil, the same ResultCollection will be returned.
func FilterResults(rc *ResultCollection, filter *Filter) (*ResultCollection, error) {
}
```

It also makes sense to convert `FilterResults` into a function for the type `ResultCollection`.

Making the filter part of the `results` package is convenient so it can be reused as a separate package (ideally, `cmd` should be as simple as possible).

wwwil: Yeah that sounds sensible, I'll change that now.

cmd/check.go
wwwil: Good point, that won't work with old config files. Based on previous discussion with @charlieegan3 I would prefer to keep the enabled/disabled rules under the `enabled-packages` list as it's tidier. I will look into how that could be made backwards compatible.

cmd/check.go
wwwil: So I have moved the filtering to a separate function, but made . it just take two string slices rather than moving the struct because I plan for the struct to form part of the config struct once we start on #64.

cmd/check.go
wwwil: I've changed this so if it can't unmarshal into the struct it tries to read it as a string slice, like before, and logs a warning about it being deprecated.

j-fuentes: I'd favour simplicity over tidiness, but I am ok with that.

cmd/check.go
```diff
@@ -333,6 +347,8 @@ func check() {
 			log.Fatalf("Cannot evaluate package %q: %v", manifest.ID, err)
 		}
 
+		rc = results.FilterResultCollection(rc, enabledPackage.DisabledRuleIDs, enabledPackage.EnabledRuleIDs)
```
j-fuentes: sweet!

PR #68, opened 2020-01-28
Make command to fail if missing Rego definition

j-fuentes:
This makes the command fail if a rule existed in the PolicyManifest but it was not found in the Rego package.

Since the reports are still useful, it does not fail hard and finished the execution after generating the reports.

### DIFF
--- a/pkg/reports/reports.go
+++ b/pkg/reports/reports.go
@@ -2,6 +2,7 @@ package reports
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/jetstack/preflight/api"
 	"github.com/jetstack/preflight/pkg/packaging"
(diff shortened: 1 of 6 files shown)


pkg/reports/reports.go
```diff
@@ -148,5 +151,21 @@ func NewReport(pm *packaging.PolicyManifest, rc *results.ResultCollection) (api.
...
 }
+
+// MissingRegoDefinitionError error to be returned when a rule from the PolicyManifest was not found in Rego.
+type MissingRegoDefinitionError struct {
+	pkg string
+	ids []string
+}
+
+func (e *MissingRegoDefinitionError) Error() string {
+	return fmt.Sprintf("the following rules from the package %q are missing its Rego definition: %s", e.pkg, strings.Join(e.ids, ", "))
```
charlieegan3:
```suggestion
	return fmt.Sprintf("the following rules from the package %q are missing their Rego definitions: %s", e.pkg, strings.Join(e.ids, ", "))
```

charlieegan3:
All looking good to me - think it makes sense to error in this case as it's a misconfiguration case I think we'd always want to avoid.

Minor edit to the message, then I can lgtm

PR #70, opened 2020-01-29
Implement WriteIndex for cluster summaries

charlieegan3:
part of #54

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/jetstack/preflight/api"
 	"github.com/jetstack/preflight/pkg/datagatherer"
 	"github.com/jetstack/preflight/pkg/datagatherer/aks"
 	"github.com/jetstack/preflight/pkg/datagatherer/eks"
@@ -322,6 +323,7 @@ func check() {
 	}
 
 	missingRules := false
+	packageReports := []api.Report{}
 	for _, enabledPackage := range enabledPackages {
 		// Make sure we loaded the package for this.
 		pkg := packages[enabledPackage.ID]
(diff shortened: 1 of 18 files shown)


cmd/check.go
```diff
@@ -338,6 +341,13 @@ func check() {
 			log.Fatalf("Cannot marshal intermediate result: %v", err)
 		}
 
+		// build a report to build the updated context for the report index
+		report, err := reports.NewReport(manifest, rc)
```
j-fuentes: Possible optimization: We don't really need the whole report here. We could create here the `ReportSummary`.

But that's is not a reason to block this PR.

pkg/output/azblob/azblob.go
```diff
@@ -78,3 +78,26 @@ func (o *Output) Write(ctx context.Context, policyManifest *packaging.PolicyMani
...
+	blobURL := containerURL.NewBlockBlobURL(fmt.Sprintf("index/%s%s", cluster, o.exporter.FileExtension()))
+
+	_, err = azblob.UploadStreamToBlockBlob(ctx, buffer, blobURL, azblob.UploadStreamToBlockBlobOptions{
+		// values chosen arbitrarily
+		BufferSize: 2 * 1024 * 1024,
+		MaxBuffers: 3,
+	})
+	if err != nil {
+		return err
+	}
```
j-fuentes: This is very similar to what is done in `Write`, I'd create a common function.

cmd/check.go
charlieegan3: Yeah... I did wonder about this... it's not ideal.

pkg/output/azblob/azblob.go
charlieegan3: Yeah, that fair. I've 'DRYed' this up in 688e94ca4f091b2520d7eb074de875e78681e693

charlieegan3:
I had to rebase and fix a failed test...

PR #73, opened 2020-01-31
Remove missing pod rules

charlieegan3:
Needed after https://github.com/jetstack/preflight/pull/68

https://github.com/jetstack/preflight/issues/74 is a note about implementing these in the future so we don't forget them :smiley:

### DIFF
--- a/preflight-packages/jetstack.io/pods/policy-manifest.yaml
+++ b/preflight-packages/jetstack.io/pods/policy-manifest.yaml
@@ -120,26 +120,6 @@ sections:
       appropriately.
     links:
     - "https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/resource-qos.md"
-  - id: cpu_limits_disabled_for_latency_sensitive_workloads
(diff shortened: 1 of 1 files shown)


j-fuentes:
Those rules are making use of the `manual` property. The problem is that is not really taken into account.

As an alternative to deleting them, I suggest that, at the moment, we omit the rules with `manual: true` when the reports are being created.

charlieegan3:
Thanks for opening an issue for the manual rules. I think this is low priority but something we can keep in the backlog.

These are causing issues today as it's not something which is handled (hence the PR)

With #74 and #76 open to keep tabs on the issue, my feeling is that this can perhaps go in as is and we can return to the issue when (if) it's deemed a priority in the future?

wwwil:
This PR I submitted should resolve this issue without needing to remove any rules: https://github.com/jetstack/preflight/pull/77

charlieegan3:
I think https://github.com/jetstack/preflight/pull/77 does indeed fix this, so closing.

PR #75, opened 2020-01-31
Update preflight.yamls to new config

charlieegan3:
Updating after https://github.com/jetstack/preflight/pull/67 was merged

### DIFF
--- a/examples/aks.preflight.yaml
+++ b/examples/aks.preflight.yaml
@@ -13,7 +13,6 @@ package-sources:
 
 enabled-packages:
 - id: "examples.jetstack.io/aks_basic"
-- id: "jetstack.io/pods"
 
 # This configures how the results will be reported.
 outputs:
--- a/examples/aks.preflight.yaml
+++ b/examples/aks.preflight.yaml
@@ -12,7 +12,8 @@ package-sources:
   dir: preflight-packages/
 
 enabled-packages:
-  - "examples.jetstack.io/aks_basic"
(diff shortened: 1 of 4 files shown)


examples/aks.preflight.yaml
```diff
@@ -12,7 +12,8 @@ package-sources:
   dir: preflight-packages/
 
 enabled-packages:
-  - "examples.jetstack.io/aks_basic"
+- "examples.jetstack.io/aks_basic"
+- "jetstack.io/pods"
```
j-fuentes: This is going to fail since the `k8s/pods` is not configured.

Also, shouldn't it be:

```
- id: "examples.jetstack.io/aks_basic"
- id: "jetstack.io/pods"
```
?

charlieegan3: Ooops, this is a copypasta error... fixed in 8c6c3ecf7ba5dd558b2a74dc5f3fabfcbb38bae0

charlieegan3: > This is going to fail since the k8s/pods is not configured.
not sure I follow

wwwil: The `k8s/pods` data gatherer is not configured above in the `data-gatherers` section, and this is required for the `pods` package.

charlieegan3: Ahh, yup. I've opted to remove the pods from both the gke and aks examples in 3f44f1df41324d5f4646c38100b30d58044516fe

PR #77, opened 2020-02-03
Don't mark manual rules without Rego as missing

### LINKED ISSUE #76: Do not report rules marked as manual
In some packages, we have rules marked as `manual: true`.

That was an attempt to have rules that cannot be checked automatically, but require a human to check them manually (see #73).

The problem is that generated reports mark those rules as `missing: true`, which is true but can be confusing.

We should discuss ways to improve that.

wwwil:
Rules that are marked as manual should not be flagged as missing if they don't have Rego implementation as they are not supposed to.

Closes #76.


j-fuentes:
+1 to this change, but I don't think this closes #76.

Currently `api.Report` does not contain `Manual`, which means it is impossible to tell if a rule failed or it requires a manual check.

I suggest either filtering out manual rules from the report or surfacing the `Manual`.

wwwil:
> Currently `api.Report` does not contain `Manual`

The `api. ReportRule` has a `Manual` field: https://github.com/jetstack/preflight/blob/master/api/report.go#L81

I think the next issue is that we need to update the web UI to show manual rules differently, or at least highlight that they are manual.

Issue #78, opened 2020-02-03
Easier end-to-end testing


wwwil:
We should make it easier to do end-to-end testing of Preflight. While there are Go tests for components the only way to test the whole application currently is against a real cluster. As Preflight is designed to work with various platforms (GKE, AKS, etc.) testing support for all of them requires many clusters, which is not convenient.

wwwil:
My proposal is that we allow fake data to be fed into data gatherers from a file.

This would have to be implemented on a per data gatherer level but would be fairly simple. A JSON file would be read and marshalled into whatever struct the data gatherer used. From this point on Preflight would run as normal.

This would allow us to store a selection of test JSON files representing GKE, AKS, etc. clusters and Kubernetes Pods which could be used to test and demo Preflight without needing an actual cluster.

The downside of this approach would be that it doesn't test the data gatherers interaction with the API they fetch data from.

It may be best to first implement a local file data gatherer and share functions from that.

wwwil:
We now have a local data gatherer and the ability to feed 'fake' local data into Preflight.

As we're moving to use an agent and SaaS model we're now [building end to end testing in the private platform repository](https://github.com/jetstack/preflight-platform/issues/295) which will incorporate testing the agent in this repo so I feel this issue can be closed.

PR #80, opened 2020-02-03
Data gatherers can fetch data from local file

wwwil:
Possible solution for #78, currently only implemented for the AKS data gatherer.

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -16,6 +16,7 @@ import (
 	"github.com/jetstack/preflight/pkg/datagatherer/eks"
 	"github.com/jetstack/preflight/pkg/datagatherer/gke"
 	"github.com/jetstack/preflight/pkg/datagatherer/k8s"
+	localdatagatherer "github.com/jetstack/preflight/pkg/datagatherer/local"
 	"github.com/jetstack/preflight/pkg/output"
 	"github.com/jetstack/preflight/pkg/output/azblob"
 	"github.com/jetstack/preflight/pkg/output/gcs"
@@ -207,6 +208,13 @@ func check() {
 					log.Fatalf("Cannot create k8s client: %+v", err)
 				}
 				dg = k8s.NewPodsDataGatherer(k8sClient)
+			} else if name == "local" {
+				localConfig, ok := config.(map[string]interface{})
+				if !ok {
+					log.Fatal("Cannot parse 'data-gatherers.local' in config.")
+				}
+				dataPath, ok := localConfig["data-path"].(string)
+				dg = localdatagatherer.NewLocalDataGatherer(dataPath)
 			} else {
 				log.Fatalf("Found unsupported data-gatherer %q in config.", name)
 			}
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -178,7 +178,7 @@ func check() {
 					log.Fatal("Cannot parse 'data-gatherers.aks' in config.")
 				}
 				msg := "'data-gatherers.aks.%s' should be a non empty string."
-				var resourceGroup, clusterName, credentialsPath, dataPath string
+				var resourceGroup, clusterName, credentialsPath string
 				if resourceGroup, ok = aksConfig["resource-group"].(string); !ok {
 					log.Fatalf(msg, "resource-group")
 				}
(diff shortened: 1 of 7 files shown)


wwwil:
@j-fuentes @charlieegan3 What do you think of this approach?

By trying to run Preflight like this I was able to find and fix #81, so I think it's worth doing something to make it easier to just run locally with fake data. Not sure if this is the best way though.

j-fuentes:
I'd rather bypass the whole data-gathering process:

- add a command-line option like `--bypass-data-gathering-with-file` that points to a json file containing the raw input for the Rego evaluation. That raw input is:
```
{
  "data-gatherer ID 1": {... data-gatherer 1 data ...},
  "data-gatherer ID 2": {... data-gatherer 2 data ...},
  ...
}
```

- if that option is present, skip the execution of the data gatherers, and check that the enabled data-gatherers that are required by packages are present in the file.

- then change [the part that builds the input for the Rego evaluation](https://github.com/jetstack/preflight/blob/baa0bc73fb4d63761082fbd5494784c8c3ee7f6e/cmd/check.go#L345) so it uses the data from the JSON.

- Additionally, we could make schema validation on the data from the JSON, but that would require more work on the `DataGatherer` interface to provide methods for data validation.

charlieegan3:
I think if there's a way of doing this without making changes to individual gatherers that'd be preferred.

wwwil:
Thanks for the feedback!

I'd agree having to do this for every data gatherer is not ideal so I'm going to look for a more general approach.

I went for loading the data from the data gatherers as it maximises how 'end-to-end' the test is and keeps as much in common with a 'normal' run as possible. Data gatherers also use the relevant `struct` for data so this provides some validation of what's being loaded from file too.

I'd prefer to be able to bypass individual data gatherers. I'd also prefer to have each data gatherer's data in separate files. I think this will be easier to work with and is more like the format we would get it in anyway, for example from `intermediate` type output.

wwwil:
@j-fuentes @charlieegan3 I've updated the implementation.

This now adds a `local` data gatherer that to read a local file. This currently has no validation on the data it reads, it is just assumed to be JSON and passed direct to OPA for evaluation.

If any of the data gatherer's config specify a `data-path` then while creating data gatherers a `local` data gatherer will be created which 'impersonates' that gatherer.

docs/configuration.md
```diff
@@ -41,6 +41,7 @@ The following data gatherers are available:
 - [Google Kubernetes Engine](docs/datagatherers/gke.md)
 - [Amazon Elastic Kubernetes Service](docs/datagatherers/eks.md)
 - [Microsoft Azure Kubernetes Service](docs/datagatherers/aks.md)
+- [Local](docs/datagatheres/local.md)
```
charlieegan3: Todo?

examples/fake.preflight.yaml
```diff
@@ -0,0 +1,21 @@
+cluster-name: my-fake-cluster
```
charlieegan3: should this file be called `local.preflight.yaml`?

cmd/check.go
```diff
@@ -130,6 +131,19 @@ func check() {
...
 			// TODO: create gatherer from config in a more clever way. We need to read gatherer config from here and its schema depends on the gatherer itself.
 			var dg datagatherer.DataGatherer
+			dataGathererConfig, ok := config.(map[string]interface{})
+			if !ok {
+				log.Fatalf("Cannot parse %s data gatherer config.", name)
+			}
+			// Check if this data gatherer's config specifies a data-path.
+			// If it does create a LocalDataGatherer to load this data but keep
+			// the name of the data gatherer it is impersonating so it can
+			// provide fake data.
```
charlieegan3:
```suggestion
			// provide local data.
```

examples/fake.preflight.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+data-gatherers:
+  aks:
+    # Fake data for AKS.
```
charlieegan3:
```suggestion
    # Local data for AKS.
```

examples/fake.preflight.yaml
```diff
@@ -0,0 +1,21 @@
...
+    data-path: ./examples/data/aks.json
+  gke:
+    # Fake data for GKE.
```
charlieegan3:
```suggestion
    # Local data for GKE.
```

charlieegan3:
Closing in favor of #80 (since I don't have permissions from Wil for his branch)

PR #82, opened 2020-02-04
Improve config management

wwwil:
The objective of this PR is to improve how config is handled and reduce clutter in the `check.go` file.

### DIFF
--- a/pkg/datagatherer/aks/aks.go
+++ b/pkg/datagatherer/aks/aks.go
@@ -19,6 +19,13 @@ type AKSDataGatherer struct {
 	credentials   *AzureCredentials
 }
 
+type AKSDataGathererConfig struct {
+	Cluster       string `mapstructure:"cluster"`
+	ResourceGroup string `mapstructure:"resource-group"`
+	Credentials   string `mapstructure:"credentials"`
+	DataPath      string `mapstructure:"data-path"`
+}
+
 // AKSInfo contains the data retrieved from AKS.
 type AKSInfo struct {
 	Cluster *aks.ManagedCluster
@@ -55,16 +62,16 @@ func readCredentials(path string) (*AzureCredentials, error) {
 }
 
 // NewAKSDataGatherer creates a new AKSDataGatherer for a cluster.
-func NewAKSDataGatherer(ctx context.Context, resourceGroup, clusterName, credentialsPath string) (*AKSDataGatherer, error) {
-	credentials, err := readCredentials(credentialsPath)
+func NewAKSDataGatherer(ctx context.Context, config *AKSDataGathererConfig) (*AKSDataGatherer, error) {
+	credentials, err := readCredentials(config.Credentials)
 	if err != nil {
 		return nil, err
 	}
 
 	return &AKSDataGatherer{
 		ctx:           ctx,
-		resourceGroup: resourceGroup,
-		clusterName:   clusterName,
+		resourceGroup: config.ResourceGroup,
+		clusterName:   config.Cluster,
 		credentials:   credentials,
 	}, nil
 }
--- a/pkg/datagatherer/eks/eks.go
+++ b/pkg/datagatherer/eks/eks.go
@@ -17,6 +17,7 @@ type EKSDataGatherer struct {
 
 type EKSDataGathererConfig struct {
 	ClusterName string
+	DataPath    string
 }
 
 // EKSInfo contains the data retrieved from EKS.
--- a/pkg/datagatherer/eks/eks.go
+++ b/pkg/datagatherer/eks/eks.go
@@ -2,6 +2,8 @@
 package eks
 
 import (
+	"context"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/eks"
@@ -13,6 +15,10 @@ type EKSDataGatherer struct {
 	clusterName string
 }
 
+type EKSDataGathererConfig struct {
+	ClusterName string
+}
+
 // EKSInfo contains the data retrieved from EKS.
 type EKSInfo struct {
 	// https://docs.aws.amazon.com/sdk-for-go/api/service/eks/#Cluster
@@ -20,10 +26,10 @@ type EKSInfo struct {
 }
 
 // NewEKSDataGatherer creates a new EKSDataGatherer for a cluster.
-func NewEKSDataGatherer(clusterName string) *EKSDataGatherer {
+func NewEKSDataGatherer(ctx context.Context, config *EKSDataGathererConfig) *EKSDataGatherer {
 	return &EKSDataGatherer{
 		client:      eks.New(session.New()),
-		clusterName: clusterName,
+		clusterName: config.ClusterName,
 	}
 }
 
--- a/pkg/datagatherer/gke/gke.go
+++ b/pkg/datagatherer/gke/gke.go
@@ -18,6 +18,15 @@ type GKEDataGatherer struct {
 	credentialsPath string
 }
 
+type GKEDataGathererConfig struct {
+	Project     string `mapstructure:"project"`
+	Location    string `mapstructure:"location"`
+	Zone        string `mapstructure:"location"` // Deprecated
+	Cluster     string `mapstructure:"cluster"`
+	Credentials string `mapstructure:"credentials"`
+	DataPath    string `mapstructure:"data-path"`
+}
+
 // GKEInfo contains the data retrieved from GKE.
 type GKEInfo struct {
 	Cluster *container.Cluster
@@ -33,11 +42,16 @@ type Cluster struct {
 }
 
 // NewGKEDataGatherer creates a new GKEDataGatherer for a cluster.
-func NewGKEDataGatherer(ctx context.Context, cluster *Cluster, credsPath string) *GKEDataGatherer {
+func NewGKEDataGatherer(ctx context.Context, config *GKEDataGathererConfig) *GKEDataGatherer {
 	return &GKEDataGatherer{
-		ctx:             ctx,
-		cluster:         cluster,
-		credentialsPath: credsPath,
+		ctx: ctx,
+		cluster: &Cluster{
+			Project:  config.Project,
+			Zone:     config.Zone,
+			Name:     config.Cluster,
+			Location: config.Location,
+		},
+		credentialsPath: config.Credentials,
 	}
 }
 
--- a/pkg/datagatherer/k8s/pods.go
+++ b/pkg/datagatherer/k8s/pods.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"log"
 
-	"github.com/jetstack/preflight/pkg/homeutils"
+	"github.com/jetstack/preflight/pkg/pathutils"
 	"github.com/pkg/errors"
 	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -25,7 +25,7 @@ type PodsInfo *core.PodList
 
 // NewPodsDataGatherer creates a new PodsDataGatherer.
 func NewPodsDataGatherer(ctx context.Context, config *PodsDataGathererConfig) *PodsDataGatherer {
-	k8sClient, err := NewClient(homeutils.ExpandHome(config.KubeConfig))
+	k8sClient, err := NewClient(pathutils.ExpandHome(config.KubeConfig))
 	if err != nil {
 		log.Fatalf("Cannot create k8s client: %+v", err)
 	}
--- a/pkg/datagatherer/k8s/pods.go
+++ b/pkg/datagatherer/k8s/pods.go
@@ -1,6 +1,10 @@
 package k8s
 
 import (
+	"context"
+	"log"
+
+	"github.com/jetstack/preflight/pkg/homeutils"
 	"github.com/pkg/errors"
 	core "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -12,13 +16,21 @@ type PodsDataGatherer struct {
 	client *kubernetes.Clientset
 }
 
+type PodsDataGathererConfig struct {
+	KubeConfig string `mapstructure:"kubeconfig,omitempty"`
+}
+
 // PodsInfo contains Pods information retrieved from the Kubernetes API.
 type PodsInfo *core.PodList
 
 // NewPodsDataGatherer creates a new PodsDataGatherer.
-func NewPodsDataGatherer(client *kubernetes.Clientset) *PodsDataGatherer {
+func NewPodsDataGatherer(ctx context.Context, config *PodsDataGathererConfig) *PodsDataGatherer {
+	k8sClient, err := NewClient(homeutils.ExpandHome(config.KubeConfig))
+	if err != nil {
+		log.Fatalf("Cannot create k8s client: %+v", err)
+	}
 	return &PodsDataGatherer{
-		client: client,
+		client: k8sClient,
 	}
 }
 
--- a/pkg/homeutils/homeutils.go
+++ b/pkg/homeutils/homeutils.go
@@ -1,0 +1,30 @@
+package homeutils
+
+import (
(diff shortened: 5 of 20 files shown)


wwwil:
This is now ready for review and I've made all the changes I had planned. However it's now crashing when I run it with the stubbed data so I need to fix that.

wwwil:
> However it's now crashing

Fixed now!

pkg/homeutils/homeutils.go
```diff
@@ -0,0 +1,30 @@
+package homeutils
```
j-fuentes: This sounds too specific. I'd rather have `pathutils`. But that's not a blocker for this to land.

pkg/output/output.go
```diff
@@ -14,3 +17,73 @@ type Output interface {
...
+			if err != nil {
+				log.Fatalf("Could not create AZBlob output: %s", err)
+			}
+			outputs = append(outputs, output)
+		} else {
+			log.Fatalf("Output type not recognised: %s", outputConfig.Type)
+		}
+	}
+	return outputs
+}
```
j-fuentes: I think all of this belongs to another package. I'd call it `cmd/checkerconfig`, for instance.

`OutputConfig` is just the format of the specific config file for `preflight check`, so I'd encapsulate everything that is related to that under the same package, close to the cmd itself.

I don't think this specific package should be shared with the new agent configuration, for instance.

pkg/datagatherer/aks/aks.go
```diff
@@ -19,6 +19,13 @@ type AKSDataGatherer struct {
 	credentials   *AzureCredentials
 }
 
+type AKSDataGathererConfig struct {
+	Cluster       string `mapstructure:"cluster"`
+	ResourceGroup string `mapstructure:"resource-group"`
+	Credentials   string `mapstructure:"credentials"`
```
j-fuentes: Rename `Credentials` to `CredentialsPath`.

pkg/datagatherer/aks/aks.go
```diff
@@ -19,6 +19,13 @@ type AKSDataGatherer struct {
 	credentials   *AzureCredentials
 }
 
+type AKSDataGathererConfig struct {
+	Cluster       string `mapstructure:"cluster"`
```
j-fuentes: Rename `Cluster` to `ClusterName`.

j-fuentes: Please, avoid the use of `mapstructure`. Everything inside `pkg/datagatherer` should be independent from viper/cobra.

And anyway, I think it is not used for serialization anywhere.

pkg/datagatherer/aks/aks.go
```diff
@@ -19,6 +19,13 @@ type AKSDataGatherer struct {
...
+	ResourceGroup string `mapstructure:"resource-group"`
+	Credentials   string `mapstructure:"credentials"`
+	DataPath      string `mapstructure:"data-path"`
```
j-fuentes: Is this `DataPath` parameter necessary?

I think this was related to loading the data from a file instead of actually instantiating a data-gatherer.

pkg/datagatherer/aks/aks.go
```diff
@@ -19,6 +19,13 @@ type AKSDataGatherer struct {
 	credentials   *AzureCredentials
 }
 
+type AKSDataGathererConfig struct {
```
j-fuentes: Documentation on the fields of this struct would be very useful.

pkg/datagatherer/eks/eks.go
```diff
@@ -13,17 +15,22 @@ type EKSDataGatherer struct {
 	clusterName string
 }
 
+type EKSDataGathererConfig struct {
```
j-fuentes: Documentation on the fields of this struct would be very useful.

pkg/datagatherer/eks/eks.go
```diff
@@ -13,17 +15,22 @@ type EKSDataGatherer struct {
...
+	ClusterName string
+	DataPath    string
```
j-fuentes: I think `DataPath` should not be here.

pkg/datagatherer/gke/gke.go
```diff
@@ -18,6 +18,15 @@ type GKEDataGatherer struct {
 	credentialsPath string
 }
 
+type GKEDataGathererConfig struct {
```
j-fuentes: Documentation on the fields of this struct would be very useful.

pkg/datagatherer/gke/gke.go
```diff
@@ -18,6 +18,15 @@ type GKEDataGatherer struct {
...
+	Project     string `mapstructure:"project"`
+	Location    string `mapstructure:"location"`
+	Zone        string `mapstructure:"location"` // Deprecated
+	Cluster     string `mapstructure:"cluster"`
+	Credentials string `mapstructure:"credentials"`
+	DataPath    string `mapstructure:"data-path"`
```
j-fuentes: See this comment about  `mapstructure`: https://github.com/jetstack/preflight/pull/82#discussion_r377681362

pkg/output/output.go
wwwil: My intent with keeping the config and config checking for each package inside the package is that it makes it easier to reuse in other commands, so it's not specific to `preflight check`.

That way as we build new components if we want them to be able to write output we can just add the struct to their config struct and it will work the same as `check`.

I don't think this will be used in the `agent` command as it won't be writing output will it?

pkg/datagatherer/k8s/pods.go
```diff
@@ -12,13 +16,21 @@ type PodsDataGatherer struct {
...
+type PodsDataGathererConfig struct {
+	KubeConfig string `mapstructure:"kubeconfig,omitempty"`
+}
+
 // PodsInfo contains Pods information retrieved from the Kubernetes API.
 type PodsInfo *core.PodList
 
 // NewPodsDataGatherer creates a new PodsDataGatherer.
-func NewPodsDataGatherer(client *kubernetes.Clientset) *PodsDataGatherer {
+func NewPodsDataGatherer(ctx context.Context, config *PodsDataGathererConfig) *PodsDataGatherer {
```
j-fuentes: `ctx` is not used.

pkg/datagatherer/aks/aks.go
wwwil: `mapstucture` is used by Viper to unmarshal config, so as these structs are used as part of the config struct we sometimes need to specify the `mapstucture` annotation.

It actually isn't required in this case, so I could tidy up here. But for other stuff, like `data-path` which becomes `DataPath`, it is required.

pkg/output/azblob/azblob.go
```diff
@@ -16,26 +17,45 @@ import (
 )
 
 // Output writes to an Azure Blob Storage bucket
-type Output struct {
+type AZBlobOutput struct {
```
j-fuentes: It is idiomatic to use `azblob.Output` instead of `azblob.AZBlobOutput`.

pkg/datagatherer/aks/aks.go
wwwil: I think if we want to completely separate each Go package from Viper then we'll need to recreate these structs in `cmd.go` to load the config, then transfer values to the config structs in the packages. This seems like a lot of extra steps though. I will have a look around and see what other projects do...

pkg/datagatherer/aks/aks.go
wwwil: The `DataPath` can be specified with all data gatherers to make Preflight use a `local` data gatherer in place of the specific data gatherer (e.g. `gke`, `eks`, etc.) to give 'fake' data. As this struct is used to unmarshal the config it needs to be included. The alternative would be to use a different struct for unmarshalling.

pkg/output/output.go
j-fuentes: > My intent with keeping the config and config checking for each package inside the package is that it makes it easier to reuse in other commands, so it's not specific to preflight check.

Exactly, and that's fine. +1 to the config structs in the specific data-gatherers.

What I was trying to say is that this method that converts the schema of the config file to a slice of outputs fits more naturally inside the cmd itself, since it is coupled to the format of the config file.

> I don't think this will be used in the agent command as it won't be writing output will it?

The agent won't write outputs.

Neither the agent of the backend should use this `OutputConfig` struct. This is coupled to cmd/check.go and because of that it should not be in a shared package but inside `check.go` or an auxiliar package inside `cmd`.

pkg/datagatherer/aks/aks.go
j-fuentes: > I think if we want to completely separate each Go package from Viper then we'll need to recreate these structs in cmd.go to load the config, then transfer values to the config structs in the packages. This seems like a lot of extra steps though. I will have a look around and see what other projects do...

Aren't we doing that already in `NewOutputs`?

pkg/datagatherer/aks/aks.go
j-fuentes: >  As this struct is used to unmarshal the config it needs to be included.

I think I am missing something. Where are doing that?

j-fuentes: wait, I was confusing the datagatherers config with the outputs config. Let me have another look.

pkg/datagatherer/aks/aks.go
j-fuentes: Ok, I see what you meant with this. In that case, ok let's keep `mapstructure`.

j-fuentes: The `eks` data-gatherer is missing `mapstructure`.

pkg/datagatherer/aks/aks.go
j-fuentes: I see now why you put `DataPath` here.

I think it shouldn't be in the specific datagatherers config but in a general type of config.

As an example, you can create a generic type that embeds all the configuration types and adds the `DataPath`.

```
type DataGathererConfig struct {
  *gke.Config
  *eks.Config
  *aks.Config
  DataPath string
}
```

I think something like that might be needed for the changes we want to make for the cert-manager package and allowing several instances of the same data-gatherer, so I am ok to keep this as you wrote it to avoid adding more complexity to this PR.

pkg/datagatherer/eks/eks.go
j-fuentes: We can fix this later: https://github.com/jetstack/preflight/pull/82#discussion_r377745851

pkg/datagatherer/gke/gke.go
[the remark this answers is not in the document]
j-fuentes: We can fix this later: https://github.com/jetstack/preflight/pull/82#discussion_r377745851

charlieegan3:
As per discussion on slack, I'm trying to work out how best to get this in. There are a few things in play here and I'd like some input on what we want to do here.

There was a discussion on our call about datagathers being an array so we can have many of the same type (gke, k8s/pods) with different config (clustera, clusterb). This is a potential requirement of the cert-manager package: (internal: https://github.com/jetstack/preflight-private/issues/138#issuecomment-582602850)

While I'd hope to be able to demo the cert-manager with the saas, this might not be an option if we aren't fast enough. This leaves the question, is this PR a requirement of the cert-manager package?

How I understand it, the following datagatherers struct implemented here doesn't support this requirement as it only allows a single config for each type of data gatherer. For reference

```go
type DataGatherersConfig struct {
	GKE   *gke.GKEDataGathererConfig     `mapstructure:"gke,omitempty"`
	AKS   *aks.AKSDataGathererConfig     `mapstructure:"aks,omitempty"`
	EKS   *eks.EKSDataGathererConfig     `mapstructure:"eks,omitempty"`
	Pods  *k8s.PodsDataGathererConfig    `mapstructure:"k8s/pods,omitempty"`
	Local *local.LocalDataGathererConfig `mapstructure:"local,omitempty"`
}
```

In my agent PR, I implemented the start of a new config format for the agent file: (part of https://github.com/jetstack/preflight/pull/87). See [here](https://github.com/jetstack/preflight/blob/4fd19fe24e7991297ad08aaff1932d7fe9c27278/pkg/agent/config_test.go#L19-L29)

The validations for the datagatherer side was implemented in the follow up PR here: https://github.com/jetstack/preflight/pull/95 which is marked as on hold because of this PR.

(We also seem to have a legacy parsing option here. But from what I can tell there are no changes to the config file - only in how it's parsed? not sure if relevant but I was confused to see backwards compatibility without change in the config so I'm checking)

I find myself unsure how best to proceed for the following reasons:
- I'm not sure how to use the data gatherer part of this without the rest
- It seems this might soon need to change to support the multiple data gatherer requirement.

Sorry for the flapping but thought best to agree before continuing.

pkg/datagatherer/aks/aks.go
charlieegan3: I wonder if it might be cleaner if the config for a DataGatherer to never know about the DataPath?

Might we just run any datagatherer with a datapath as a local data gatherer? e.g. https://github.com/jetstack/preflight/pull/95/files#diff-d1aa418498ff92dc3def120d994c0379R45-R48

j-fuentes:
> As per discussion on slack, I'm trying to work out how best to get this in. There are a few things in play here and I'd like some input on what we want to do here.
> 
> There was a discussion on our call about datagathers being an array so we can have many of the same type (gke, k8s/pods) with different config (clustera, clusterb). This is a potential requirement of the cert-manager package: (internal: [jetstack/preflight-private#138 (comment)](https://github.com/jetstack/preflight-private/issues/138#issuecomment-582602850))
> 
> While I'd hope to be able to demo the cert-manager with the saas, this might not be an option if we aren't fast enough. This leaves the question, is this PR a requirement of the cert-manager package?
> 
> How I understand it, the following datagatherers struct implemented here doesn't support this requirement as it only allows a single config for each type of data gatherer. For reference
> 
> ```go
> type DataGatherersConfig struct {
> 	GKE   *gke.GKEDataGathererConfig     `mapstructure:"gke,omitempty"`
> 	AKS   *aks.AKSDataGathererConfig     `mapstructure:"aks,omitempty"`
> 	EKS   *eks.EKSDataGathererConfig     `mapstructure:"eks,omitempty"`
> 	Pods  *k8s.PodsDataGathererConfig    `mapstructure:"k8s/pods,omitempty"`
> 	Local *local.LocalDataGathererConfig `mapstructure:"local,omitempty"`
> }
> ```
> 
> In my agent PR, I implemented the start of a new config format for the agent file: (part of #87). See [here](https://github.com/jetstack/preflight/blob/4fd19fe24e7991297ad08aaff1932d7fe9c27278/pkg/agent/config_test.go#L19-L29)
> 
> The validations for the datagatherer side was implemented in the follow up PR here: #95 which is marked as on hold because of this PR.
> 
> (We also seem to have a legacy parsing option here. But from what I can tell there are no changes to the config file - only in how it's parsed? not sure if relevant but I was confused to see backwards compatibility without change in the config so I'm checking)
> 
> I find myself unsure how best to proceed for the following reasons:
> 
> * I'm not sure how to use the data gatherer part of this without the rest
> * It seems this might soon need to change to support the multiple data gatherer requirement.
> 
> Sorry for the flapping but thought best to agree before continuing.

When this PR was initiated, it was just a refactoring. So it wasn't intended to add support for multiple instances of the same data-gatherer.

I'd create a PR that refactors the configuration of the data-gatherers themselves, adding for instance the `aks.Config` struct, for instance. but keeping the config file as it is. This will unblock #95.

Then as part of #95, I'd create a config that supports multiple instances of data-gatherers.

charlieegan3:
> So it wasn't intended to add support for multiple instances of the same data-gatherer.

Sure, I just wanted to check that it still made sense to use this with that upcoming requirement in mind.

I think I can have a go at this though - thanks.

j-fuentes:
We are splitting out this PR into chunks.

So far we have merged #100, which creates uniform configuration and constructors for the data-gatherers.

@charlieegan3 and I talked this morning about if it is worth optimising those config structs on the data-gatherers packages for a direct unmarshal of the cmd/check config file. We think it is better to get #95 done and get the agent loading config in a cleaner and direct way and then adapt the config of cmd/agent if necessary.

wwwil:
We have agreed to implement this PR as multiple smaller PRs to make the work more manageable so I am going to close this.

PR #84, opened 2020-02-10
Add cluster and timestamp to summary

j-fuentes:
We are not storing the timestamp nor the cluster name to the cluster summary.

This fix is not the best way of doing it. A better solution will come with #1.

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -391,6 +391,9 @@ func check() {
 		if err != nil {
 			log.Fatalf("Cannot generate report for results: %v", err)
 		}
+		// at the moment, we need to add Cluster and Timestamp here because of #1.
+		report.Cluster = clusterName
+		report.Timestamp = api.Time{Time: checkTime}
 		packageReports = append(packageReports, report)
 
 		for _, output := range outputs {


charlieegan3:
Is this related to this: https://github.com/jetstack/preflight/blob/master/api/cluster.go#L22-L23

j-fuentes:
> Is this related to this: https://github.com/jetstack/preflight/blob/master/api/cluster.go#L22-L23

```
type ReportSet struct {
	Cluster      string           `json:"-"`
	Timestamp    Time             `json:"timestamp"`
	FailureCount int              `json:"failureCount"`
	SuccessCount int              `json:"successCount"`
	Reports      []*ReportSummary `json:"reports"`
}

type ReportSummary struct {
	ID           string `json:"id"`
	Package      string `json:"package"`
	Cluster      string `json:"-"`
	Timestamp    Time   `json:"-"`
	FailureCount int    `json:"failureCount"`
	SuccessCount int    `json:"successCount"`
}
```

`Cluster` and `Timestamp` are hidden in the JSON serialization to avoid redundancy, but they need to be in the struct so `reports.NewClusterSummary(packageReports)` can generate the `ReportSet`.

charlieegan3:
I see - just checking it wasn't related as I ran into this last week.

Issue #86, opened 2020-02-10
[agent] Implement minimal config file loading for agent


charlieegan3:
**Sub-tasks**
* [ ] implement a means of loading config - much the same as preflight core
* [ ] configure schedule variable
* [ ] configure endpoint variable
* [ ] configure identity token variable
* [ ] configure loaded data gather variables

This is just to load the vars from the file - nothing else.

Users of the agent need this to be able to control how the agents works when its running in their clusters. If we one day have a generated yaml, this might form the content of a configmap to be loaded into the running agent.

**Acceptance Criteria:**

- [ ] I can run the agent with different config
- [ ] for now, the agent just prints the config and exits

---
Internal: Created from: https://github.com/jetstack/preflight-private/issues/260 Needs: https://github.com/jetstack/preflight-private/issues/263 Enables: https://github.com/jetstack/preflight-private/issues/265

Issue #85, opened 2020-02-10
[agent] Create a new agent binary/command


charlieegan3:
We're going to cut back the scope of the cluster component to a more simple 'agent'.

This agent binary will have a reduced set of responsibilities:

- gathering data on a schedule
- uploading data for evaluation to some backend

This story is to *just build the binary* that we can start adding other features to afterwards.

This is being done to simplify the installation of preflight and support the remote evaluation of packages.

**Acceptance Criteria:**

- [ ] It is possible to build an agent binary from the makefile in the project

Internal Use: Created from https://github.com/jetstack/preflight-private/issues/260 Enables: https://github.com/jetstack/preflight-private/issues/264

charlieegan3:
Looking over this... I think I might just add it as a command to the existing preflight core.

charlieegan3:
Closing as I don't have permissions to assign this to a project.

PR #87, opened 2020-02-10
Add minimal version of the agent command

charlieegan3:
Fixes https://github.com/jetstack/preflight/issues/85

Fixes https://github.com/jetstack/preflight/issues/86

This is my go at the most simple implementation of the above issues. We needed an agent with some config. All the config parts are in place to run datagatherers and upload data.

Why did I not use the existing preflight yaml? Wil is working on this today - I would have needed to add a more fields to the config. I might well end up using what Wil refactors but this allows the work on the agent to continue either way.

### DIFF
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -1,0 +1,25 @@
+package cmd
+
+import (
+	"log"
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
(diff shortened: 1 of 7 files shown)


j-fuentes:
I have mixed feelings about having the agent baked into the same binary than the "normal" CLI. Let's do it like this for now as it seems to simplify things.

PR #88, opened 2020-02-10
Serialize cluster in ReportSet

j-fuentes:
I missed this in my previous PR (#84).


charlieegan3:
/lgtm
/approve

Right! Now this makes sense to me!

j-fuentes:
This is not needed after #89

PR #89, opened 2020-02-10
Custom unmarshaler for ClusterSummary

j-fuentes:
This is needed to a `ClusterSummary` can be unmarshaled correctly just by doing `json.Unmarshal`.

### DIFF
--- a/api/cluster.go
+++ b/api/cluster.go
@@ -1,5 +1,7 @@
 package api
 
+import "encoding/json"
+
 // ClusterSummary contains a summary of the most recent status of a cluster.
 type ClusterSummary struct {
 	Cluster         string     `json:"cluster"`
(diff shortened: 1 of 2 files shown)


api/cluster.go
```diff
@@ -24,3 +26,26 @@ type ReportSummary struct {
...
 }
+
+// UnmarshalJSON unmarshals a ClusterSummary.
+func (c *ClusterSummary) UnmarshalJSON(data []byte) error {
+	type Alias ClusterSummary
+	aux := &struct {
+		*Alias
+	}{
+		Alias: (*Alias)(c),
+	}
```
charlieegan3: I'm not sure exactly what this does. [(lines)](https://github.com/jetstack/preflight/pull/89/files#diff-bc8a656be21edce0cf2a74adf23d7aeaR32-R37). :thinking:

j-fuentes: We cannot do this:

```
var aux ClusterSummary
if err := json.Unmarshal(data, &aux); err != nil {
		return err
}
```

That would call the `UnmarshalJSON` method we are declaring recursively with no end, which leads to stack overflow.

We need to create that artificial alias so we can unmarshal to an intermediate variable using the native unmarshal logic and then apply our custom logic.

charlieegan3: Ahh, that makes sense. Thanks.

PR #93, opened 2020-02-10
Return error on closing gcs write

j-fuentes:


pkg/output/gcs/gcs.go
```diff
@@ -83,5 +83,5 @@ func (o *Output) WriteIndex(ctx context.Context, cluster string, timestamp time.
 		return err
 	}
 
-	return nil
+	return err
```
wwwil: Is this to keep a reference to the error in case something goes wrong in the `defer`? Could you add a comment to clarify this as its a bit strange.

j-fuentes: Yes, that's the reason.

I have added a comment.

PR #96, opened 2020-02-12
Make agent upload data

### LINKED ISSUE #91: [agent] Implement data upload from agent
Agents are not able to evaluate packages, only gather data. They need to be able to upload to a backend.

~Needs: https://github.com/jetstack/preflight-private/issues/265~ (or it can upload can upload an empty body for now?)

Enables: https://github.com/jetstack/preflight-private/issues/267

**Sub-tasks**
* [ ] implement an agent command that gathers data and posts it to an endpoint on a given path
* [ ] configure the endpoint and path in the config file

**Acceptance Criteria:**

- [ ] there is an agent command I can run that gathers data and uploads it to the supplied endpoint

https://webhook.site can be used to test that the data is sent correctly.

---
Internal: Created from: https://github.com/jetstack/preflight-private/issues/260

j-fuentes:
This closes #91

### DIFF
--- a/api/reading.go
+++ b/api/reading.go
@@ -1,0 +1,8 @@
+package api
+
+// Reading is the output of a datagatherer.
+type Reading struct {
+	DataGatherer string      `json:"data-gatherer"`
+	Timestamp    Time        `json:"timestamp"`
+	Data         interface{} `json:"data"`
+}
--- a/cmd/echo.go
+++ b/cmd/echo.go
@@ -46,7 +46,7 @@ func echoHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var readings []*api.Reading
+	var readings []*api.DataReading
 
 	err := json.NewDecoder(r.Body).Decode(&readings)
 	if err != nil {
@@ -60,7 +60,7 @@ func echoHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	fmt.Println("-----")
 
-	fmt.Fprintf(w, "{}")
+	fmt.Fprintf(w, `{ "status": "ok" }`)
 	w.Header().Set("Content-Type", "application/json")
 }
 
--- a/cmd/echo.go
+++ b/cmd/echo.go
@@ -1,0 +1,71 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http"
+
+	"github.com/jetstack/preflight/api"
+	"github.com/spf13/cobra"
+)
+
+var echoCmd = &cobra.Command{
+	Use:   "echo",
+	Short: "starts an echo server to test the agent",
+	Long: `The agent sends data to a server. This echo server
+can be used to act as the server part and echo the data received by the agent.`,
+	Run: echo,
+}
+
+var echoListen string
+
+func init() {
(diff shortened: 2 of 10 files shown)


api/reading.go
```diff
@@ -0,0 +1,8 @@
+package api
+
+// Reading is the output of a datagatherer.
+type Reading struct {
```
charlieegan3: Thoughts about DataReading - to match DataGatherer?

cmd/echo.go
```diff
@@ -0,0 +1,56 @@
...
+	http.HandleFunc("/", echoHandler)
+	fmt.Println("Listening to requests at ", echoListen)
+	http.ListenAndServe(echoListen, nil)
+}
+
+func echoHandler(w http.ResponseWriter, r *http.Request) {
+	// if r.Method != http.MethodPost {
+	// 	http.Error(w, fmt.Sprintf("invalid method. Expected POST, received %s", r.Method), http.StatusBadRequest)
+	// 	return
+	// }
```
charlieegan3: Any reason to keep this around? (seems like the echo server shouldn't be concerned with validating the agent's behaviour?)

pkg/agent/config.go
```diff
@@ -8,18 +8,21 @@ import (
...
+	Endpoint      Endpoint       `yaml:"endpoint"`
 	DataGatherers []dataGatherer `yaml:"data-gatherers"`
 }
 
-type endpoint struct {
-	Host string `yaml:"host"`
-	Path string `yaml:"path"`
+// Endpoint is the configuration of the server where to post the data.
+type Endpoint struct {
+	Protocol string `yaml:"protocol"`
```
charlieegan3: Should protocol default to http?

pkg/agent/run.go
```diff
@@ -27,14 +33,61 @@ func Run(cmd *cobra.Command, args []string) {
...
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	if code := resp.StatusCode; code < 200 || code >= 300 {
+		return fmt.Errorf("Received response with status code %d. Body: %s", code, string(body))
+	}
+
+	log.Println("Data send correctly. Server says: ", string(body))
```
charlieegan3: (Just a suggestion)

```suggestion
	log.Println("Data sent successfully. Server says: ", string(body))
```

charlieegan3:
https://github.com/j-fuentes/preflight/pull/1 should fix this failing test.

cmd/echo.go
j-fuentes: That was old, now it is not commented but actual code.

The agent should send a POST, so it makes sense that the echo server returns an error if it receives something that is not a POST.

And in fact, this has been already useful. I was building the URL wrong in the agent (`http://localhost:8080//`, with double trailing `/`) and that was causing the http client to send a GET instead of a POST, even though I was using `http.Post()`. ¯\_(ツ)_/¯

pkg/agent/config.go
j-fuentes: ok, let's make it default to http, since this is going to be used for local development mainly.

j-fuentes:
Thanks for the review!

PR #98, opened 2020-02-13
Switch k8s/pod data gatherer to use the dynamic client

munnerz:
This is the first stage of the work to support cert-manager resources in Preflight.

This changeset refactors the 'k8s' data gatherer to not require a `runtime.Scheme` with the type being validated registered. This should make it possible to fetch CRD resources and other unknown types in future 😄

/assign @j-fuentes

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -216,7 +216,7 @@ func check() {
 				if !ok {
 					log.Println("Didn't find 'kubeconfig' in 'data-gatherers.k8s/pods' configuration. Assuming it runs in-cluster.")
 				}
-				k8sClient, err := k8s.NewClient(expandHome(kubeconfigPath))
+				k8sClient, err := k8s.NewDynamicClient(expandHome(kubeconfigPath))
 				if err != nil {
 					log.Fatalf("Cannot create k8s client: %+v", err)
 				}
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,6 @@ require (
 	github.com/google/go-cmp v0.3.0
 	github.com/gookit/color v1.2.0
 	github.com/hashicorp/go-multierror v1.0.0
-	github.com/ianlancetaylor/demangle v0.0.0-20181102032728-5e5cf60278f6 // indirect
 	github.com/juju/errors v0.0.0-20190930114154-d42613fe1ab9
 	github.com/juju/loggo v0.0.0-20190526231331-6e530bcce5d8 // indirect
 	github.com/juju/testing v0.0.0-20191001232224-ce9dec17d28b // indirect
@@ -24,7 +23,6 @@ require (
 	github.com/open-policy-agent/opa v0.16.0
 	github.com/pkg/errors v0.8.1
 	github.com/spf13/cobra v0.0.5
-	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.5.0
 	github.com/yudai/gojsondiff v1.0.0
 	golang.org/x/net v0.0.0-20200114155413-6afb5195e5aa // indirect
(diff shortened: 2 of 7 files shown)


go.mod
```diff
@@ -33,9 +31,9 @@ require (
 	google.golang.org/api v0.15.0
 	gopkg.in/d4l3k/messagediff.v1 v1.2.1
 	gopkg.in/yaml.v2 v2.2.7
-	k8s.io/api v0.17.0
```
munnerz: :tada: we don't depend on anything in `k8s.io/api` now it seems 😄

pkg/datagatherer/k8s/generic.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Fetch will fetch the requested data from the apiserver, or return an error
+// if fetching the data fails.
+func (g *genericGatherer) Fetch() (interface{}, error) {
+	if g.groupVersionResource.Resource == "" {
+		return nil, fmt.Errorf("resource type must be specified")
+	}
+	resourceInterface := namespaceResourceInterface(g.cl.Resource(g.groupVersionResource), g.namespace)
+	list, err := resourceInterface.List(metav1.ListOptions{})
```
munnerz: We may want to consider allowing the ListOptions to be provided by the caller in future by adding a field to the `genericGatherer` struct. For now it isn't necessary though 😄

charlieegan3:
So the switch to dynamic is purely to support the fetching of Custom Resources? This isn't possible with the clientset in the kubernetes package I guess?

Presumably no downsides of using the dynamic client for our use case?

/approve

munnerz:
> So the switch to dynamic is purely to support the fetching of Custom Resources? This isn't possible with the clientset in the kubernetes package I guess?

Correct - the dynamic client decodes into `unstructured.Unstructured`, which is basically a `map[string]interface{}` that implements the `runtime.Object` interface.

The regular 'clientset' is a *typed* client, and it has to be used with a corresponding `runtime.Scheme` that has had the types it is handling registered with it, so that the clientset can decode into typed objects (e.g. `v1.Pod`) when fetching resources.

This should have a net-zero effect, although it is worth noting that the type of resource being output from the pod data gatherer will change from a `PodList` to an `UnstructuredList`, which is identical in all ways aside from the internal representation in Go's type system.

Because the Rego rules accept input as an arbitrary blob of JSON, this shouldn't make any difference as there's no point where we attempt to access this data is a `PodList` (as evidenced by the fact we no longer even depend on `k8s.io/api` at all!)

munnerz:
I don't think there's any integration testing capability in Preflight right now, so I can't show you this in the form of a test case. Hopefully the unit test cases provided here can provide some level of confidence! I've not actually attempted to run any packages after writing this code, so it may be worth getting a cluster up and running and running the pods checks against it...

pkg/datagatherer/k8s/generic_test.go
```diff
@@ -0,0 +1,110 @@
...
+			"metadata": map[string]interface{}{
+				"resourceVersion": "",
+			},
+		},
+		Items: itemsNonPtr,
+	}
+}
+
+func TestGenericGatherer_Fetch(t *testing.T) {
+	emptyScheme := runtime.NewScheme()
```
munnerz: Used an empty scheme here to try and demonstrate it doesn't require anything to be registered/recognised in order to work 😄

charlieegan3:
Ok cool - thanks for the extra details. I'll try this out just now.

cmd/check.go
```diff
@@ -216,7 +216,7 @@ func check() {
 				if !ok {
 					log.Println("Didn't find 'kubeconfig' in 'data-gatherers.k8s/pods' configuration. Assuming it runs in-cluster.")
 				}
-				k8sClient, err := k8s.NewClient(expandHome(kubeconfigPath))
+				k8sClient, err := k8s.NewDynamicClient(expandHome(kubeconfigPath))
```
j-fuentes: This is a nice trick. I didn't know about the existence of that `DynamicClient`.

pkg/datagatherer/k8s/generic.go
j-fuentes: Agree, it is an interesting feature for the future.

PR #99, opened 2020-02-13
Expose generic data gatherer through custom gatherer name format

munnerz:
This PR adds two data gatherers utilising the new 'generic' data gatherer.

In future, we can refactor the way the `k8s` gatherer is instantiated so that we don't need to explicitly add new config key entries for each type into the preflight binary.

This could be done by either expanding the gatherer config in the package config, or otherwise encoding the GroupVersionResource into the name of the data gatherer (e.g. `k8s/certificates.v1alpha2.cert-manager.io`, `k8s/pods.v1.core` etc).

For now, this allows users to write Preflight packages that make use of (some of the) cert-manager API types 🎉.

This PR is based on top of #98.

/assign @j-fuentes


munnerz:
We could probably even simplify this PR now to just expose `NewGenericGatherer` which accepts a GVR to save having a struct for each type of gatherer too... happy to do that as part of this PR if you'd prefer 😄

munnerz:
> We could probably even simplify this PR now to just expose NewGenericGatherer which accepts a GVR to save having a struct for each type of gatherer too... happy to do that as part of this PR if you'd prefer 😄

Nerd-sniped myself - I made this change 😄

charlieegan3:
/approve 

This looks good to me but I'll wait for @j-fuentes to lgtm. 

I think my preference would be to get #99 and #98 in before we address the config refactor in https://github.com/jetstack/preflight/pull/82.

This allows us to begin work on & testing building the cert-manager rules.

pkg/datagatherer/k8s/generic_test.go
```diff
@@ -90,9 +90,9 @@ func TestGenericGatherer_Fetch(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			cl := fake.NewSimpleDynamicClient(emptyScheme, test.objects...)
 			g := genericGatherer{
-				cl: cl,
+				cl:                   cl,
 				groupVersionResource: test.gvr,
-				namespace: test.namespace,
+				namespace:            test.namespace,
```
munnerz: Forgot to run `go fmt` before the last PR (#98)!

cmd/check.go
```diff
@@ -220,7 +221,39 @@ func check() {
...
+				trimmed := strings.TrimPrefix(name, "k8s/")
+				nameOnDots := strings.SplitN(trimmed, ".", 3)
+				// if a user has used for example `k8s/pods.v1`, we should
+				// handle this case and set the group name to empty.
+				if len(nameOnDots) == 2 {
+					nameOnDots = append(nameOnDots, "")
+				}
+				if len(nameOnDots) != 3 {
+					log.Fatal("Failed to parse generic k8s plugin configuration. Expected data gatherer name of the form k8s/{resource-name}.{api-version}.{api-group}")
+				}
```
j-fuentes: Doesn't this break compatibility with existing configurations using `k8s/pods`?

charlieegan3: I think it should be possible to keep this with a manual case for `k8s/pod` -> `k8s/pods.v1` as the output will still be grouped under the datagatherer name (https://github.com/jetstack/preflight/pull/99/files#diff-f9c4c955b1d6c71769486619f542fef1R267)

j-fuentes: yes, but that "mapping" is needed here.

munnerz: No, I haven't removed the old code - `k8s/pod` is the block just above this.

munnerz: https://github.com/jetstack/preflight/blob/4423174dc2013aa09b1434296999291fc3119124/cmd/check.go#L211-L228

j-fuentes: ups, I missed that.

PR #95, opened 2020-02-11
Implement data gathering functionality of agent

charlieegan3:
Fixes https://github.com/jetstack/preflight/issues/90

PR #100, opened 2020-02-13
Data gatherers config

j-fuentes:
This offloads from #82 the formalization of the configuration for the datagatherers.

It also moves the responsibility of the validation to the data-gatherer package itself. If we finally end up refactoring the whole cmd/check config file parsing to rely on Viper's `mapstructure` annotations, we can make validation errors use the mapstructure tag ([see example](https://stackoverflow.com/questions/40864840/how-to-get-the-json-field-names-of-a-struct-in-golang)).

This is needed to unblock #95, although additional changes might be needed.

### DIFF
--- a/pkg/datagatherer/aks/aks.go
+++ b/pkg/datagatherer/aks/aks.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 
 	aks "github.com/Azure/aks-engine/pkg/api/agentPoolOnlyApi/v20180331"
 )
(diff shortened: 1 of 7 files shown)


pkg/datagatherer/aks/aks.go
```diff
@@ -11,19 +10,31 @@ import (
...
-// AKSDataGatherer is a DataGatherer for AKS.
-type AKSDataGatherer struct {
-	ctx           context.Context
-	resourceGroup string
-	clusterName   string
-	credentials   *AzureCredentials
+// Config is the configuration for an AKS DataGatherer.
+type Config struct {
+	// ClusterID is the ID of the cluster in AKS.
+	ClusterID string `mapstructure:"cluster-id"`
```
charlieegan3: I think in AKS they appear to use clustername rather than id https://docs.microsoft.com/en-us/rest/api/aks/managedclusters/get

ID seems to be the 'full qualified' id, e.g. `/subscriptions/subid1/resourcegroups/rg1/providers/Microsoft.ContainerService/managedClusters/clustername1`

j-fuentes: good point

charlieegan3:
/approve

This looks good to me, pending AKS field name change.

I think this will make https://github.com/jetstack/preflight/pull/95 nicer.

j-fuentes:
I have renamed ClusterID -> ClusterName for aks and eks, since both seem to be using that terminology.

charlieegan3:
Sounds good.

/lgtm

Issue #92, opened 2020-02-10
[agent] Make agent share identity with backend


charlieegan3:
Users of the agent need to to be able to prove their ID so that they have access to all the features of the backend. When uploading data - share the identity of the agent.

**Sub-tasks**
* [x] share the configured token as a bearer token header or similar
* [x] handle 401 errors gracefully with a good error message

**Acceptance Criteria:**

- [x] Agent token shows in request header (which would allow a backend to determine its ID)

**Notes**

The use of token auth doesn't scale in the long term as we add many agents tokens to the backend configuration. In the future it'd likely we'll offload the authentication to some third party. I use Auth0 as an example.

In this version of the future agents would gain an access token from Auth0 to talk to the backend using '[machine to machine](https://auth0.com/machine-to-machine/)' auth. We'd create client credentials to give to installers of agents automatically, perhaps using an [API endpoint like this](https://auth0.com/docs/api/management/v2#!/Clients/post_clients).

Our use of simple static auth doesn't make this option harder to reach in future, it looks almost identical to the installer of the agent. It'd also be easy to run both Auth0 + static token auth at the same time if needed.

---
Internal: Created from: https://github.com/jetstack/preflight-private/issues/260 Pre-req: https://github.com/jetstack/preflight-private/issues/266

PR #110, opened 2020-02-18
Echo server to accept auth token

j-fuentes:
In preparation for #92, this makes the echo server accept a token.


charlieegan3:
/approve
/lgtm

looks good!

PR #111, opened 2020-02-18
Fix refactor config decoding in check function

wwwil:
Change dataGatherersConfig variable creation to instantiate map

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -163,7 +163,7 @@ func check() {
 	}
 
 	// Decode data gatherer config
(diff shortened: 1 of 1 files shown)


charlieegan3:
/approve
/lgtm

shame there's three ways to do this in go (at least that I know of...)

wwwil:
> shame there's three ways to do this in go (at least that I know of...)

Yeah, I was gonna look into which one was best but for now felt it was worth just getting this fixed so it doesn't trip us up later!

charlieegan3:
I think the `make` way is only 'better' if you know the size - fine by me.

PR #112, opened 2020-02-18
Move check config loading to own package

wwwil:
This PR moves the loading and decoding of config for a Preflight check to its own package. This is then used with the `LoadConfig` function which returns a config struct.


charlieegan3:
Am I right in saying the context for this is: "we want to have a lighter check command" e.g. move out some of the logic from that mega function. (or is there more to this?)

wwwil:
@charlieegan3 > Am I right in saying the context for this is: "we want to have a lighter check command" e.g. move out some of the logic from that mega function. (or is there more to this?)

Yeah pretty much. I find the amount of code for handling config loading in `check.go` distracts from what it's actually doing and makes it cluttered.

In the long run I'd like to improve the config loading and move to relying more on `validate` functions on different config structs. But as we're focusing on the SaaS stuff now this seems a good place to leave this.

pkg/checkconfig/checkconfig.go
```diff
@@ -0,0 +1,316 @@
+package checkconfig
```
j-fuentes:
```suggestion
// Package checkconfig provides all the necessary for the `check` command to load its configuration.
package checkconfig
```

j-fuentes:
This new package should not go into the `/pkg` directory but into `/internal`.
Other than that, it is ok.

wwwil:
> This new package should not go into the /pkg directory but into /internal.

Ah cool! We should probably also move `pkg/agent` to `/internal` too.

j-fuentes:
> > This new package should not go into the /pkg directory but into /internal.
> 
> Ah cool! We should probably also move `pkg/agent` to `/internal` too.

Indeed, we might do that. I am not sure if that would be used without the `cmd/agent` CLI. It probably won't.

PR #114, opened 2020-02-19
Embed ReportMetadata into Report

j-fuentes:
This avoids some repetition and makes it easy to retrieve the metadata of a report without implementing new helpers.

### DIFF
--- a/pkg/reports/reports_test.go
+++ b/pkg/reports/reports_test.go
@@ -14,17 +14,19 @@ import (
 )
 
 var fixtureReport1 = api.Report{
-	ID:               "fixtureReport1",
-	PreflightVersion: "version.PreflightVersion",
-	Package:          "examplePackage.ID",
-	PackageInformation: api.PackageInformation{
-		Namespace: "examplePackage.Namespace",
-		ID:        "examplePackage.ID",
-		Version:   "examplePackage.PackageVersion",
+	ReportMetadata: api.ReportMetadata{
+		ID:               "fixtureReport1",
+		PreflightVersion: "version.PreflightVersion",
+		Package:          "examplePackage.ID",
+		PackageInformation: api.PackageInformation{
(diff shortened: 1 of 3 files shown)


j-fuentes:
I have mixed feelings about this because of the barrier of what is metadata and what is not it a bit diffuse.

If at some point we decide to make the Description also part of the metadata, it will require to change all places where we create reports.

Maybe the `ReportMetadata` struct should be kept separate from what an actual report is, so we have more freedom to change both types without affecting too many things.

WDYT @charlieegan3 ?

j-fuentes:
> I have mixed feelings about this because of the barrier of what is metadata and what is not it a bit diffuse.
> 
> If at some point we decide to make the Description also part of the metadata, it will require to change all places where we create reports.
> 
> Maybe the `ReportMetadata` struct should be kept separate from what an actual report is, so we have more freedom to change both types without affecting too many things.
> 
> WDYT @charlieegan3 ?

#115 is an alternative to this.

pkg/reports/reports_test.go
```diff
@@ -14,17 +14,19 @@ import (
...
+		Package:          "examplePackage.ID",
+		PackageInformation: api.PackageInformation{
+			Namespace: "examplePackage.Namespace",
+			ID:        "examplePackage.ID",
+			Version:   "examplePackage.PackageVersion",
+		},
+		Cluster: "exampleCluster",
 	},
 	Name:        "examplePackage.Name",
 	Description: "examplePackage.Description",
```
charlieegan3: I think I prefer #115 because as you say it's hard to decide what is and isn't metadata. e.g. for me, I'd expect name and ID to be at the same level.

j-fuentes:
Closing this in favour of #114.

charlieegan3:
*See #115

PR #115, opened 2020-02-19
Add GenReportMetadata function to Report

j-fuentes:


charlieegan3:
I think this might be preferred over #114. I might prefer to include it alongside some callers of the new function to give more validation/context.

PR #119, opened 2020-02-20
Fix data-gatherers config loading

j-fuentes:
This is needed so we load the config of the particular data-gatherers.


pkg/agent/config.go
```diff
@@ -25,10 +31,76 @@ type Endpoint struct {
...
+		}{}
+		err = unmarshal(&cfg)
+		dg.Config = cfg.Config
+	// dummy dataGatherer is just used for testing
+	case "dummy":
+		cfg := struct {
+			Config *dummyConfig `yaml:"config"`
+		}{}
+		err = unmarshal(&cfg)
+		dg.Config = cfg.Config
```
j-fuentes: I am sure there is a better way of doing this with `reflect`, but I am too tired at the moment.

I will revisit this tomorrow.

j-fuentes: I have reworked this part. Hopefully, it is better now.

pkg/agent/config.go
```diff
@@ -25,10 +31,72 @@ type Endpoint struct {
...
+	err = yaml.Unmarshal(bb, config)
+	if err != nil {
+		return nil
+	}
+
+	return nil
+}
+
+// UnmarshalYAML unmarshals a dataGatherer resolving the type according to Kind.
+func (dg *dataGatherer) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
charlieegan3: This is called implicitly when loading yaml data into a dg right?

pkg/agent/config.go
charlieegan3: The new version looks :+1:

charlieegan3:
I'm going to lgtm - but would be nice to confirm this question here: https://github.com/jetstack/preflight/pull/119#discussion_r383210502

/lgtm
/approve

pkg/agent/config.go
j-fuentes: Exactly: https://godoc.org/gopkg.in/yaml.v2#Unmarshaler

charlieegan3: :+1: thanks.

PR #124, opened 2020-02-24
k8s: allow setting KUBECONFIG to use via environment variable

munnerz:
This allows users to omit the `kubeconfig` field from their config files, instead opting to use their default configured kubeconfig file (as per typical kubectl loading rules).

I've also added a very basic test that exercises these two codepaths. Without breaking out the config loading piece into its own function, this seemed like a nice and quick way to check things work (albeit it *doesn't* validate the the *expected* config file is being loaded, only that one _is_ loaded).

### DIFF
--- a/pkg/datagatherer/k8s/client.go
+++ b/pkg/datagatherer/k8s/client.go
@@ -4,21 +4,47 @@ package k8s
 import (
 	"github.com/pkg/errors"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 )
 
 // NewDynamicClient creates a new 'dynamic' clientset using the provided kubeconfig.
-// If kubeconfigPath is not set/empty, it will attempt to load the InClusterConfig.
+// If kubeconfigPath is not set/empty, it will attempt to load configuration using
+// the default loading rules.
 func NewDynamicClient(kubeconfigPath string) (dynamic.Interface, error) {
-	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	cfg, err := loadRESTConfig(kubeconfigPath)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
-
-	cl, err := dynamic.NewForConfig(config)
+	cl, err := dynamic.NewForConfig(cfg)
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
-
 	return cl, nil
+}
+
+func loadRESTConfig(path string) (*rest.Config, error) {
+	switch path {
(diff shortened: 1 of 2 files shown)


pkg/datagatherer/k8s/client.go
```diff
@@ -4,21 +4,39 @@ package k8s
...
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+	// Otherwise use the explicitly named kubeconfig file.
+	default:
+		cfg, err = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+			&clientcmd.ClientConfigLoadingRules{ExplicitPath: kubeconfigPath},
+			&clientcmd.ConfigOverrides{}).ClientConfig()
+		if err != nil {
+			return nil, errors.WithStack(err)
```
munnerz: I'm not too sure if we _really_ want `errors.WithStack(err)` all over the place, and feel like it may be more sensible to only print this stack trace as and when it may be useful. But I didn't want to break convention here 😄

pkg/datagatherer/k8s/client.go
```diff
@@ -4,21 +4,47 @@ package k8s
...
+		cfg, err := clientcmd.NewDefaultClientConfig(*apicfg, &clientcmd.ConfigOverrides{}).ClientConfig()
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+		return cfg, nil
+	// Otherwise use the explicitly named kubeconfig file.
+	default:
+		cfg, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+			&clientcmd.ClientConfigLoadingRules{ExplicitPath: path},
+			&clientcmd.ConfigOverrides{}).ClientConfig()
```
munnerz: This is the code that `clientcmd.BuildConfigFromFlags` used to call, minus the `masterURL` value being set. I opted to do it this way simply to make it clearer what's being set where, but 🤷‍♂

pkg/datagatherer/k8s/client_test.go
```diff
@@ -0,0 +1,77 @@
...
+	defer f.Close()
+	if err := clientcmdlatest.Codec.Encode(&cfg, f); err != nil {
+		t.Fatal(err)
+	}
+	return f.Name()
+}
+
+func createValidTestConfig() clientcmdapi.Config {
+	const (
+		server = "https://anything.com:8080"
```
charlieegan3: nit but can we use example.com?

munnerz: sure, this was taken straight from client-go so didn't pay much attention :)

charlieegan3:
Thanks for this James - very nice!

/lgtm
/approve

Issue #118, opened 2020-02-20
Update datagatherer documentation with example data they send


charlieegan3:
It's important that we have examples of the data sent by each data gatherer to the backend (when running the agent).

Update each datagatherer to show an example json payload - perhaps in the readme of the datagatherer.

This is useful for reviewing the data sent but also when making new packages.

PR #125, opened 2020-02-24
Explain data and permissions for all datagatherers

charlieegan3:
Fixes https://github.com/jetstack/preflight/issues/118

What this does:

    - documents all the dg data and permissions
    - updates to the new k8s generic gatherer
    - references the new agent config format

What this does not do:

    - update all the examples since I think they need to be revisited and
      removed/update for the new architecture.


docs/datagatherers/eks.md
```diff
@@ -0,0 +1,46 @@
+# EKS Data Gatherer
+
+The EKS *data gatherer* fetches information about a cluster from the AWS
+Kubernetes Service API.
```
wwwil:
```suggestion
Elastic Kubernetes Service API.
```

docs/datagatherers/aks.md
```diff
@@ -0,0 +1,36 @@
+# AKS Data Gatherer
+
+The AKS *data gatherer* fetches information about a cluster from the Azure
+Kubernetes Service API.
+
+## Data
```
wwwil: We also have an example of the data collected here: https://github.com/jetstack/preflight/tree/master/examples/data

docs/datagatherers/gke.md
```diff
@@ -1,68 +1,62 @@
 # GKE Data Gatherer
 
-The GKE *data gatherer* fetches information about a cluster 
-from the Google Kubernetes Engine API.
+The GKE *data gatherer* fetches information about a cluster from the Google
```
wwwil: Should we standardise a markdown line length?

docs/datagatherers/gke.md
```diff
@@ -1,68 +1,62 @@
...
 
-The gathered data looks like this:
+It's comparable to the output from:
 
-```json
-{
-  "Cluster": {...}
-}
+```bash
+gcloud container clusters describe main --format=json
```
wwwil:
```suggestion
gcloud container clusters describe my-cluster --format=json
```

charlieegan3: fixed in 6068de10553fed71f81bb8b64a350c0913776478

docs/datagatherers/gke.md
j-fuentes: ¯\_(ツ)_/¯

I don't care.

docs/datagatherers/aks.md
charlieegan3: Hmm, forgot about those. My preference would be, for now, to keep all this information in one place - or at least local to the data gatherer folder (and not some other dir in the project).

Main reason being I'm keen to keep the number of places changes to dg and other config down.

If it were up to me, I'd delete that examples/data folder since we have links to the data here in the readmes now.

docs/datagatherers/gke.md
wwwil: Yeah that's what I've been going for, I also try to use semantic line breaks.

docs/datagatherers/eks.md
charlieegan3: Fixed in 1319d079eda23ae69d48ae1e8f25d4fd8292ece1

docs/datagatherers/aks.md
charlieegan3: Removed the examples in 527ff171fef87e3f4cfe8312ea3a2dc4977abe82 - reasoning repeated in the commit message.

examples/data/aks.json
```diff
@@ -1,45 +0,0 @@
...
-        "kubernetesVersion": "1.14.8",
-        "provisioningState": "Succeeded"
-      },
-      "type": "Microsoft.ContainerService/ManagedClusters",
-      "name": "preflight-test",
-      "location": "uksouth",
-      "id": "/subscriptions/*****************/resourcegroups/preflight/providers/Microsoft.ContainerService/managedClusters/preflight-test"
-    }
-  }
-}
```
wwwil: Why is this being removed?

I've found it quite useful to run the [stubbed config](https://github.com/jetstack/preflight/blob/master/examples/stubbed.preflight.yaml) as a basic test of the CLI, and I imagine it could be used with the agent too.

wwwil: Ah I missed your comment in the commit message! I still feel this is useful to have...

docs/datagatherers/aks.md
```diff
@@ -0,0 +1,36 @@
...
+## Configuration
+
+To use the AKS data gatherer add an `aks` entry to the `data-gatherers`
+configuration. For example:
+
+```
+data-gatherers:
+- kind: "aks"
+  name: "aks"
+  config:
```
wwwil: Something this highlights is that how we configure data gatherers is now different in the CLI and the agent.

j-fuentes: That's ok. It had to change because ultimately we aim to support multiple instances of every data-gatherer in the agent.

PR #126, opened 2020-02-25
Move sub-tasks to the end

j-fuentes:
Move sub-tasks to the end, since it is a more natural order when writng the issue. Remove checklists where it does not apply.


.github/ISSUE_TEMPLATE/user-story.md
```diff
@@ -29,21 +22,28 @@ Acceptance Criteria is a set of test scenarios specific to this particular issue
...
 
 **Risks:**
-- [ ] ...
-- [ ] ...
-- [ ] ...
-- [ ] ...
+- ...
+- ...
+- ...
+- ...
```
charlieegan3: While we're here... would it make sense to make it only one list item? 4 seems like more to delete since we usually only have 1 or 2...

j-fuentes: Done.

PR #130, opened 2020-03-02
Kubernetes data gatherer exclude some namespaces

### LINKED ISSUE #128: Limit namespaces used in k8s datagatherer
It's a request to be able to ignore pods in the kube-system namespace as these are often not under the control of the users in managed clusters.

**Sub-tasks:**
* [ ] support excluding namespacs in a k8s datagatherer config

**Acceptance Criteria:**

- [ ] I should be able to run the pods datagatherer without sending the pod data from the kube-system ns

**Assumptions:**
- [ ] this is a good first port of call, and label selectors and other more advanced filters are not needed at this time.

This task relates to this [milestone](https://github.com/jetstack/preflight-private/milestone/9)

wwwil:
Closes #128

### DIFF
--- a/internal/checkconfig/checkconfig.go
+++ b/internal/checkconfig/checkconfig.go
@@ -170,6 +170,7 @@ func LoadConfig(configPath string) (*Config, error) {
 				if !ok {
 					log.Println("Didn't find 'kubeconfig' in 'data-gatherers.k8s/pods' configuration. Assuming it runs in-cluster.")
 				}
+				excludedNamespaces := dataGathererConfigMap["excluded-namespaces"].([]string)
 				dataGathererConfig = &k8s.Config{
 					KubeConfigPath: pathutils.ExpandHome(kubeconfigPath),
 					GroupVersionResource: schema.GroupVersionResource{
@@ -177,6 +178,7 @@ func LoadConfig(configPath string) (*Config, error) {
 						Version:  "v1",
 						Resource: "pods",
 					},
+					ExcludeNamespaces: excludedNamespaces,
 				}
 			} else if strings.HasPrefix(name, "k8s/") {
 				trimmed := strings.TrimPrefix(name, "k8s/")
--- a/internal/checkconfig/checkconfig.go
+++ b/internal/checkconfig/checkconfig.go
@@ -170,14 +170,21 @@ func LoadConfig(configPath string) (*Config, error) {
 				if !ok {
 					log.Println("Didn't find 'kubeconfig' in 'data-gatherers.k8s/pods' configuration. Assuming it runs in-cluster.")
 				}
-				excludedNamespaces := []string{}
-				excludedNamespacesFromFile, ok := dataGathererConfigMap["exclude-namespaces"].([]interface{})
-				if ok {
-					for _, excludedNamespaceFromFile := range excludedNamespacesFromFile {
-						excludedNamespace := excludedNamespaceFromFile.(string)
-						if ok {
-							excludedNamespaces = append(excludedNamespaces, excludedNamespace)
+				excludeNamespaces := []string{}
+				// The exclude-namespaces field is optional, only try to decode it if present.
+				if excludeNamespacesFromFile, ok := dataGathererConfigMap["exclude-namespaces"]; ok {
+					// Cast to a list of interfaces.
+					excludeNamespacesListFromFile, ok := excludeNamespacesFromFile.([]interface{})
+					if !ok {
+						return nil, fmt.Errorf("Could not parse exclude-namespaces list from configuration: %+v", excludeNamespacesFromFile)
+					}
+					// Cast each value in the list to a string.
+					for _, excludeNamespaceFromFile := range excludeNamespacesListFromFile {
+						excludeNamespace, ok := excludeNamespaceFromFile.(string)
+						if !ok {
+							return nil, fmt.Errorf("Could not parse exclude-namespaces value from configuration: %+v", excludeNamespaceFromFile)
 						}
+						excludeNamespaces = append(excludeNamespaces, excludeNamespace)
 					}
 				}
 				dataGathererConfig = &k8s.Config{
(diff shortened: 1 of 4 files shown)


wwwil:
This PR implements the solution discussed in #128 and is tested and working so now ready for review.

There are some caveats to this approach that I found:

Firstly we can't test the field selector in our local tests as the fake client we use does not implement field selector filtering.

Secondly namespaces won't be excluded if `data-path` is used to stub in local data from a file. As this feature is for internal testing it's not a huge issue, but may catch people out.

If we view these as significant problems we are probably best to choose a new implementation.

pkg/datagatherer/k8s/generic.go
```diff
@@ -110,3 +121,16 @@ func namespaceResourceInterface(iface dynamic.NamespaceableResourceInterface, na
 	}
 	return iface.Namespace(namespace)
 }
+
+// generateFieldSelector creates a field selector string from a list of
+// namespaces to exclude.
+func generateFieldSelector(excludeNamespaces []string) string {
```
j-fuentes: Since we could use `FieldSelectors` for other type of filtering, I'd rename it to `excludeNamespacesToFieldSelector` or something that makes it explicit it is for excluding namespaces.

wwwil: I imagined that if we added more things to the field selector they would also be handled by this function. Then we could split stuff out later to other functions if it gets too big.

internal/checkconfig/checkconfig.go
```diff
@@ -170,13 +170,24 @@ func LoadConfig(configPath string) (*Config, error) {
 				if !ok {
 					log.Println("Didn't find 'kubeconfig' in 'data-gatherers.k8s/pods' configuration. Assuming it runs in-cluster.")
 				}
+				excludedNamespaces := []string{}
+				excludedNamespacesFromFile, ok := dataGathererConfigMap["exclude-namespaces"].([]interface{})
```
munnerz: 🤔 do we not have any typing on this config file we are decoding into...? It may not be specific to this PR, but it seems totally wrong that we're casting to a slice of interfaces here - is it not possible for us to decode into the structure we actually earlier on than this, and actually make use of the type system?

internal/checkconfig/checkconfig.go
```diff
@@ -170,13 +170,24 @@ func LoadConfig(configPath string) (*Config, error) {
...
+				if ok {
```
munnerz: Probably best to invert this to check for `if !ok` and throw an unexpected error if it is not okay, which also saves the extra indenting here for the 'happy path'?

pkg/datagatherer/k8s/generic.go
```diff
@@ -110,3 +121,16 @@ func namespaceResourceInterface(iface dynamic.NamespaceableResourceInterface, na
...
+	var fieldSelector string
+	for _, excludeNamespace := range excludeNamespaces {
+		if excludeNamespace == "" {
+			continue
+		}
+		fieldSelector = fmt.Sprintf("%smetadata.namespace!=%s,", fieldSelector, excludeNamespace)
```
munnerz: This works for now, but you really should consider using `field.Selector` and constructing this in a safe way, instead of directly constructing a string to be sent as a selector to the server 😄

pkg/datagatherer/k8s/generic_test.go
```diff
@@ -86,6 +86,9 @@ func TestGenericGatherer_Fetch(t *testing.T) {
 				getObject("foobar/v1", "Foo", "testfoo", "nottestns"),
 			),
 		},
+		// Note that we can't test use of fieldSelector to exclude namespaces
+		// here as the as the fake client does not implement it.
+		// See go/pkg/mod/k8s.io/client-go@v0.17.0/dynamic/fake/simple.go:291
```
munnerz: I think it _is_ possibly to define a custom reactor however, that should be able to access request parameters like this in order to ensure that the expected selector is actually being sent to the server when this request is made.

Alternatively/additionally, we could look at a light-weight integration test for this once there's a few more places that need it. Perhaps worth adding a TODO for the integration test, and opening an issue to track adding an integration test framework?

internal/checkconfig/checkconfig.go
wwwil: So this is just the config loading for `check` which is being deprecated. It's not great but we decided not to work on it as it's going away. The `agent` part which replaces it has much better config loading.

internal/checkconfig/checkconfig.go
wwwil: This isn't a required field so we don't need to error. Though I should probably error on the inner `if ok`  as that implies someone has tried to specify a namespace for exclusion but it can't be parsed.

pkg/datagatherer/k8s/generic.go
wwwil: Ah, didn't know that was a thing, will look into it!

internal/checkconfig/checkconfig.go
munnerz: Yeah, `!ok` indicates that the typing of the field is wrong, i.e. the user has specified: `excludedNamespaces: "sensitive-ns"` or something. Probably safest to error in these cases as we may do something unexpected (i.e. not exclude the `sensitive-ns` namespace like the use attempted to do)

wwwil: I've updated this to work better.

pkg/datagatherer/k8s/generic.go
wwwil: Have switched to using `fields.Selector` functions to build the selector. Am I correct in thinking it still needs to be converted to string with `.String()` to be used in the `ListOptions`?

pkg/datagatherer/k8s/generic_test.go
wwwil: Yeah I've spent a bit of time looking into this but I think we're best to open a separate issue for it.

PR #133, opened 2020-03-09
Redact Kubernetes Secret data

### LINKED ISSUE #129: Implement a k8s secret gatherer
We have recently implemented a generic k8s data gatherer https://github.com/jetstack/preflight/pull/99. This is being used to fetch resources for the private certmanager package. To make this package ready for real use, we need to make sure that when secrets are gathered, only their metadata is sent.

**Acceptance Criteria:**
- [ ] when I run a k8s/secrets.v1 datagatherer - only the metadata is sent
- [ ] it is not possible to configure the agent to send secret data to the backend

**Risks:**
- [ ] people have secret data in non k8s-secret resources 

**Assumptions:**
- [ ] we're not going to be running policy on secret content

This task relates to this [milestone](https://github.com/jetstack/preflight-private/milestone/9)

wwwil:
This PR changes the Kubernetes data gatherer so that when Secret resources are gathered their data is removed.

Closes #129

### DIFF
--- a/pkg/datagatherer/k8s/generic.go
+++ b/pkg/datagatherer/k8s/generic.go
@@ -131,17 +131,13 @@ func redactList(list *unstructured.UnstructuredList) error {
 		if err != nil {
 			return errors.WithStack(err)
 		}
-		redactItem := false
 		for _, gvk := range gvks {
 			// If this item is a Secret then we need to redact it.
 			if gvk.Kind == "Secret" {
-				redactItem = true
+				// Redact the Secret by overwriting its data.
+				list.Items[i].Object["data"] = map[string]interface{}{}
 				break
 			}
-		}
-		// Redcat the Secret by overwriting its data.
-		if redactItem {
-			list.Items[i].Object["data"] = map[string]interface{}{}
 		}
 	}
 	return nil
--- a/pkg/datagatherer/k8s/generic.go
+++ b/pkg/datagatherer/k8s/generic.go
@@ -3,7 +3,6 @@ package k8s
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	"github.com/jetstack/preflight/pkg/datagatherer"
 	"github.com/pkg/errors"
@@ -11,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes/scheme"
 )
 
 // Config contains the configuration for the data-gatherer.
@@ -112,9 +112,19 @@ func (g *DataGatherer) Fetch() (interface{}, error) {
 		return nil, errors.WithStack(err)
 	}
 	// Redact Secret data
-	if strings.ToLower(g.groupVersionResource.Resource) == "secrets" {
-		for i := range list.Items {
-			list.Items[i].Object["data"] = map[string]interface{}{}
+	for i := range list.Items {
+		gvks, _, err := scheme.Scheme.ObjectKinds(list.Items[i].DeepCopyObject())
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+		for _, gvk := range gvks {
+			redact := false
+			if gvk.Kind == "Secret" {
+				redact = true
+			}
+			if redact {
+				list.Items[i].Object["data"] = map[string]interface{}{}
+			}
 		}
 	}
 	return list, nil
--- a/pkg/datagatherer/k8s/generic.go
+++ b/pkg/datagatherer/k8s/generic.go
@@ -112,7 +112,7 @@ func (g *DataGatherer) Fetch() (interface{}, error) {
 		return nil, errors.WithStack(err)
 	}
 	// Redact Secret data
-	if strings.ToLower(g.groupVersionResource.Resource) == "secret" {
+	if strings.ToLower(g.groupVersionResource.Resource) == "secrets" {
 		for i := range list.Items {
 			list.Items[i].Object["data"] = map[string]interface{}{}
 		}
--- a/pkg/datagatherer/k8s/generic.go
+++ b/pkg/datagatherer/k8s/generic.go
@@ -127,7 +127,7 @@ func redactList(list *unstructured.UnstructuredList) error {
 	// Iterate over the items in the list.
 	for i := range list.Items {
 		// Determine the kind of items in case this is a generic 'mixed' list.
-		gvks, _, err := scheme.Scheme.ObjectKinds(list.Items[i].DeepCopyObject())
+		gvks, _, err := scheme.Scheme.ObjectKinds(&list.Items[i])
 		if err != nil {
 			return errors.WithStack(err)
 		}
--- a/pkg/datagatherer/k8s/generic.go
+++ b/pkg/datagatherer/k8s/generic.go
@@ -133,7 +133,7 @@ func redactList(list *unstructured.UnstructuredList) error {
 		}
 		for _, gvk := range gvks {
 			// If this item is a Secret then we need to redact it.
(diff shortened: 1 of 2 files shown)


pkg/datagatherer/k8s/generic.go
```diff
@@ -110,6 +111,12 @@ func (g *DataGatherer) Fetch() (interface{}, error) {
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
+	// Redact Secret data
+	if strings.ToLower(g.groupVersionResource.Resource) == "secret" {
```
munnerz: I think you also need to match `secrets` here too, as technically `secrets` with an 's' is the *resource name* (as part of the REST path).

I imagine there's also a clever way we can use the `runtime.Scheme` to inspect the list type to check if it == `SecretList` (i.e. the resource *kind*), which is more concrete.

pkg/datagatherer/k8s/generic_test.go
```diff
@@ -86,6 +92,21 @@ func TestGenericGatherer_Fetch(t *testing.T) {
 				getObject("foobar/v1", "Foo", "testfoo", "nottestns"),
 			),
 		},
+		"Secret resources should have data removed": {
+			gvr: schema.GroupVersionResource{Group: "", Version: "v1", Resource: "Secret"},
```
munnerz: `Secret` should be `secrets`

pkg/datagatherer/k8s/generic.go
wwwil: Yeah, changing to `secrets` with an 's' has fixed things!

I did try doing `if list.Object["kind"] == "SecretList" {` here, however while that worked against a real cluster is didn't work with the test suite.

pkg/datagatherer/k8s/generic.go
```diff
@@ -110,6 +111,12 @@ func (g *DataGatherer) Fetch() (interface{}, error) {
...
+		for i := range list.Items {
+			list.Items[i].Object["data"] = map[string]interface{}{}
```
wwwil: Also is it adequate to just work with the unstructured objects, or should they be properly converted to `Secrets`?

pkg/datagatherer/k8s/generic.go
munnerz: So I think a better approach is to use a runtime.Scheme to *identify* the type of resource, instead of relying on things like the `kind` or `apiVersion` being set. A runtime.Scheme will be able to use the type information registered with it to identify what kind it is. Let me have a dig now and see what method there is for this...

pkg/datagatherer/k8s/generic.go
munnerz: Unstructured is fine, assuming the question/thread above is resolved IMO

pkg/datagatherer/k8s/generic.go
munnerz: https://godoc.org/k8s.io/apimachinery/pkg/runtime#Scheme.ObjectKinds this should do it 😄 from there you can iterate over all the returned GVKs and if any of them have a `kind` of `Secret` and a `group` of either `core` or `` (empty string) then redact it

munnerz: Ah. you'll actually need to match on `SecretList` come to think.. but same difference 😄

munnerz: You will need to use a 'scheme' that has the 'Secret' and 'SecretList' kind registered with it - Kubernetes provides this as `k8s.io/client-go/kubernetes/scheme.Scheme`, which you should be safe to directly reference here

pkg/datagatherer/k8s/generic.go
```diff
@@ -110,9 +112,41 @@ func (g *DataGatherer) Fetch() (interface{}, error) {
...
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
 	return list, nil
 }
 
+func redactList(list *unstructured.UnstructuredList) error {
+	// In principal we could only redact the list if it's kind is SecretList or
+	// a generic mixed List, however the test suite does not set the list kind
+	// and it is safer to always check for Secrets.
```
wwwil: @munnerz Does this make sense? Or should I look into the issues with the test suite further?

pkg/datagatherer/k8s/generic.go
```diff
@@ -110,9 +112,41 @@ func (g *DataGatherer) Fetch() (interface{}, error) {
...
+	// Iterate over the items in the list.
+	for i := range list.Items {
+		// Determine the kind of items in case this is a generic 'mixed' list.
+		gvks, _, err := scheme.Scheme.ObjectKinds(list.Items[i].DeepCopyObject())
```
munnerz: You don't need to deep copy this here 😄 (i.e. you can replace `list.Items[i].DeepCopyObject()` with `list.Items[i]`). The `ObjectKinds` method doesn't modify the object, so it is safe to do 😄

pkg/datagatherer/k8s/generic.go
```diff
@@ -110,9 +112,41 @@ func (g *DataGatherer) Fetch() (interface{}, error) {
...
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		redactItem := false
+		for _, gvk := range gvks {
+			// If this item is a Secret then we need to redact it.
+			if gvk.Kind == "Secret" {
```
munnerz: Need to check the 'group' as well (check for `core` or `` (empty))

pkg/datagatherer/k8s/generic.go
```diff
@@ -110,9 +112,41 @@ func (g *DataGatherer) Fetch() (interface{}, error) {
...
+				redactItem = true
+				break
+			}
+		}
+		// Redcat the Secret by overwriting its data.
+		if redactItem {
```
munnerz: Why do this outside of the loop? Can this be done with `redactItem` gets set to true above instead?

pkg/datagatherer/k8s/generic.go
```diff
@@ -110,9 +112,41 @@ func (g *DataGatherer) Fetch() (interface{}, error) {
...
+			list.Items[i].Object["data"] = map[string]interface{}{}
```
munnerz: Hm - there's a few different ways this could be redacted...

1) Removing the `data` field altogether
2) Making `data` be `nil`
3) Setting `data` to an empty map (what currently happens here now).

I think (3) _does_ make sense to do, but thought I'd call out the various options here as there _are_ options :)

pkg/datagatherer/k8s/generic.go
wwwil: Ah yeah, not sure why I did it that way...

charlieegan3:
This looks perfect

/lgtm

pkg/datagatherer/k8s/generic.go
```diff
@@ -110,9 +112,37 @@ func (g *DataGatherer) Fetch() (interface{}, error) {
...
+	// Iterate over the items in the list.
+	for i := range list.Items {
+		// Determine the kind of items in case this is a generic 'mixed' list.
+		gvks, _, err := scheme.Scheme.ObjectKinds(&list.Items[i])
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		for _, gvk := range gvks {
+			// If this item is a Secret then we need to redact it.
+			if gvk.Kind == "Secret" && (gvk.Group == "Core" || gvk.Group == "") {
```
munnerz: `Core` should be `core`

munnerz: (groups are always lowercase)

wwwil: 😬 Fixing now!

wwwil: https://github.com/jetstack/preflight/pull/136

Issue #127, opened 2020-02-26
Create manifests to deploy preflight agent


charlieegan3:
In order to help users install Preflight in their clusters, we should provide a guide and some manifests to do this.

**Sub-tasks:**
* [x] Create a readme explaining the installation of the agent
* [ ] Explain authorization requirements of datagathers, perhaps linking to updated dg docs for things like k8s rbac etc.
* [x] Create a deployment yaml to run the agent (1 replica)

**Acceptance Criteria:**

- [ ] There is a clear process outlined to install preflight with the k8s/pods and gke dg configured.

**Assumptions:**
- [ ] we're only worrying about k8s/pods and gke for now.
- [ ] users want to run preflight in a k8s cluster

This task relates to this [milestone ](https://github.com/jetstack/preflight-private/milestone/9)

PR #134, opened 2020-03-10
Use non interactive client loading rules

charlieegan3:
Without this, agents running in pods will still look for a kubeconfig
file.

Encountered during: https://github.com/jetstack/preflight/issues/127


munnerz:
I haven't actually pulled this and tested it but it looks good to me generally. It is a shame we do not have any unit tests for this function so we can automatically check 😄 

If you're happy to merge this as-is and explore unit tests later, feel free to comment `/unhold` and this should merge automatically!

/lgtm
/hold

charlieegan3:
/unhold 


?

PR #135, opened 2020-03-10
Agent installation docs

charlieegan3:
Fixes https://github.com/jetstack/preflight/issues/127

This is a basic write up of how to install the agent. I think it might also be worth including an example YAML for gathering GKE data?


README.md
```diff
@@ -121,26 +126,142 @@ packages](./docs/how_to_write_packages.md).
...
+
+### Agent
+
+The following instructions walk through the installation of the Preflight agent
+to gather data about cluster pods and send them to the backend for analysis.
+
+**To complete the secret manifest below, you will need to have a Preflight
+token.**
+
+First create a namespace for the preflight components:
```
wwwil: I feel like these manifests should be in files in `examples/` or `deploy/` so they are easier for user to consume, though this might be an issue for a future PR.

README.md
```diff
@@ -121,26 +126,142 @@ packages](./docs/how_to_write_packages.md).
...
+        resources:
+          requests:
+            memory: "200Mi"
+            cpu: "200m"
+          limits:
+            memory: "200Mi"
+            cpu: "200m"
+```
+
+### Preflight V1 (check)
```
wwwil: Maybe to make things cleaner all the info about check should be moved out to another file in `docs/` and just mentioned and linked once above. Might also be best for another PR.

charlieegan3: I can move them out if that's the popular vote.

My reasoning here was that we have documentation in a lot of places - this is becoming harder to keep in sync as we deprecate check. 

Until then, I wanted to keep the changes minimal and quite self contained until we're ready to remove check references completely...

README.md
charlieegan3: I think my other comment touches on this.

README.md
wwwil: Yeah I can see the reasoning there. I think I just don't like the cluttered feel in the mean time but I can always just get over it.

My thinking was that we actually move all check specific stuff into `docs/check.md`, refer to this once in `README.md` and otherwise everything can focus on the agent and new SaaS way of working.

charlieegan3: Sounds reasonable, I'll see what that looks like.

charlieegan3: ffe6c83351954fec31df77d8c638f3584b4fbcb7 is an attempt at this.

charlieegan3: I've tried to move all the check docs and examples to one place - leaving the new readme as a clear process for installing the agent.

README.md
```diff
@@ -18,29 +18,22 @@ checks using [Open Policy Agent (OPA)](https://www.openpolicyagent.org/).
 **[Preflight _agent_](#agent).**
 While this is happening the docs may be inconsistent and the repo structure may
 change.
+[Docs for this component can still be found here](./docs/installation_manual_in_cluster.md).
```
wwwil:
```suggestion
[Docs for the check tool can still be found here](./docs/installation_manual_in_cluster.md).
```

Just to clarify as we mentioned two components in the previous sentence.

wwwil:
```suggestion
[Docs for this component can still be found here](./docs/check.md).
```

docs/check.md
```diff
@@ -0,0 +1,439 @@
...
+:warning: :construction:
+
+The Preflight _check_ tool also uses _data gatherers_ to collect required data,
+but unlike the _agent_ it evaluates this data and produces a report locally.
+
+**This functionality is now deprecated in favour of the _agent_.** Previous
+versions of the _check_ tool can still be downloaded and used, however it is no
+longer being maintained and will be removed from this repository.
+
+## Packages
```
wwwil: The packages are still in use so I think this section should either stay in `README.md` or move to `docs/packages.md`

charlieegan3: But they're not in use by users of the agent - and will likely go before they come back from the public repo.

docs/check.md
```diff
@@ -0,0 +1,439 @@
...
+Create your `preflight.yaml` configuration file. There is full [configuration
+documentation](./docs/configuration.md) available, as well as several example
+files in [`./examples`](./examples).
+
+### Locally
+
+By default Preflight looks for a configuration at `./preflight.yaml`. Once this
+is set up, run a Preflight check like so:
+
+``` preflight check ```
```
wwwil:
```suggestion
``` 
preflight check
```
```

README.md
charlieegan3: fixed in aff5c9841c73a6736816db01edf4552ed1edf1df

charlieegan3: fixed in aff5c9841c73a6736816db01edf4552ed1edf1df

docs/check.md
wwwil: Ah yeah! I might PR this paragraph back into the private repo's `README.md` then.

wwwil: Actually I think it would still be good to briefly mention packages. Maybe just:

```
## Packages

Policies for cluster configuration are encoded into *Preflight packages*. 
Each package focuses on a different aspect of the cluster, for example the
`gke` package provides rules for the configuration of a GKE cluster.

Preflight packages are implemented using
[Open Policy Agent](https://www.openpolicyagent.org) with evaluation
taking place in the SaaS backend.
```

charlieegan3: Yeah, good idea, added in: 835d92fd9e963624e9bbe8ec67839624424578d0

PR #138, opened 2020-03-18
Support providing auth token in cli flag

j-fuentes:
This is convenient in some cases.

For instance, now it is easier to inject an env. variable with `kubectl set env` and user it in as a flag.

### DIFF
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -20,6 +20,13 @@ func init() {
 		"agent-config-file",
 		"c",
 		"./agent.yaml",
-		"Config file location, default is `agent.yaml` in the current working directory",
+		"Config file location, default is `agent.yaml` in the current working directory.",
+	)
+	agentCmd.PersistentFlags().StringVarP(
+		&agent.AuthToken,
+		"auth-token",
+		"t",
+		"",
+		"Authorization token. If used, it will override the authorization token in the configuration file.",
 	)
 }
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -20,6 +20,9 @@ import (
 // ConfigFilePath is where the agent will try to load the configuration from
 var ConfigFilePath string
 
+// AuthToken is the authorization token that will be used for API calls
+var AuthToken string
+
 // Run starts the agent process
 func Run(cmd *cobra.Command, args []string) {
 	ctx := context.Background()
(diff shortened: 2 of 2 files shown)


james-w:
/approve

There's a little package called `flagenv` that allows reading flag values from env vars that may make some of this easier.

j-fuentes:
> /approve
> 
> There's a little package called `flagenv` that allows reading flag values from env vars that may make some of this easier.

I tried that, but I didn't figure out a way to integrate with `cobra`.

james-w:
Ah, right. I think this can work in that case:

```
func SetFlagsFromEnv(prefix string, fs *pflag.FlagSet) {
    set := map[string]bool{}
    fs.Visit(func(f *pflag.Flag) {
        set[f.Name] = true
    })
    fs.VisitAll(func(f *pflag.Flag) {
        // ignore flags set from the commandline
        if set[f.Name] {
            return
        }
        // remove trailing _ to reduce common errors with the prefix, i.e. people setting it to MY_PROG_
        cleanPrefix := strings.TrimSuffix(prefix, "_")
        name := fmt.Sprintf("%s_%s", cleanPrefix, strings.Replace(strings.ToUpper(f.Name), "-", "_", -1))
        if e, ok := os.LookupEnv(name); ok {
            _ = f.Value.Set(e)
        }
    })
}
```

where the flagset is `agentCmd.Flags()`

j-fuentes:
> Ah, right. I think this can work in that case:
> 
> ```
> func SetFlagsFromEnv(prefix string, fs *pflag.FlagSet) {
>     set := map[string]bool{}
>     fs.Visit(func(f *pflag.Flag) {
>         set[f.Name] = true
>     })
>     fs.VisitAll(func(f *pflag.Flag) {
>         // ignore flags set from the commandline
>         if set[f.Name] {
>             return
>         }
>         // remove trailing _ to reduce common errors with the prefix, i.e. people setting it to MY_PROG_
>         cleanPrefix := strings.TrimSuffix(prefix, "_")
>         name := fmt.Sprintf("%s_%s", cleanPrefix, strings.Replace(strings.ToUpper(f.Name), "-", "_", -1))
>         if e, ok := os.LookupEnv(name); ok {
>             _ = f.Value.Set(e)
>         }
>     })
> }
> ```
> 
> where the flagset is `agentCmd.Flags()`

Thanks! Done here: https://github.com/jetstack/preflight/pull/141

PR #141, opened 2020-03-18
Bind env variables to config flags

j-fuentes:
This binds environment variables to cli flags which makes configuration easier for containers.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -3,8 +3,10 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 // rootCmd represents the base command when called without any subcommands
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -17,6 +17,12 @@ configuration checks using Open Policy Agent (OPA).
 Preflight checks are bundled into Packages`,
 }
 
+func init() {
+	for _, command := range rootCmd.Commands() {
+		setFlagsFromEnv("PREFLIGHT_", command.PersistentFlags())
```
james-w: I think this wants to be `.Flags()` instead, so that it picks up any local flags for that command.

cmd/agent.go
```diff
@@ -13,6 +18,27 @@ var agentCmd = &cobra.Command{
 	Run: agent.Run,
 }
 
+func setFlagsFromEnv(prefix string, fs *pflag.FlagSet) {
+	fmt.Println("setflags")
```
charlieegan3: should, this be a log line? or removed?

cmd/root.go
j-fuentes: I am not sure about the difference, but it does not work with `Flags`

cmd/agent.go
j-fuentes: ups, it should be removed.

PR #144, opened 2020-03-19
Add period flag

j-fuentes:
Add a period flag with a default period of 1 hour.


james-w:
/approve

I wonder if there should be some server-side control over this, either telling the client when to send data, or just dropping some data rather than generating a report every time. Something for later for sure.

j-fuentes:
> I wonder if there should be some server-side control over this, either telling the client when to send data, or just dropping some data rather than generating a report every time. Something for later for sure.

Yes. We plan to make the agent retrieve the period, and other configuration, from the server.

Issue #155, opened 2020-04-14
Scrape whitelisted namespaces to allow more restricted RBAC


charlieegan3:
**Acceptance Criteria:**

Acceptance Criteria is a set of test scenarios specific to this particular issue. It should capture functional and non-functional requirements.

- [ ] it is possible to run preflight with permissions to scrape a single namespace without knowing the other namespaces in the cluster to exclude

**Assumptions:**
- Some users will not be able to have cluster-viewer to run preflight.

**Sub-tasks:**
* [ ] extend the generic datagatherer here https://github.com/jetstack/preflight/blob/764397f8ba1aad718eb9d3a05893ae213c7680ee/pkg/datagatherer/k8s/generic.go

charlieegan3:
:point_up:  was merged.

PR #156, opened 2020-04-14
Scrape only Included namespaces when set to allow more restricted RBAC

charlieegan3:
This implements https://github.com/jetstack/preflight/issues/155

Turns our that --field-selector in the API does not support set operations like label selectors :cry: I instead make a request for each namespace.

### DIFF
--- a/pkg/datagatherer/k8s/generic.go
+++ b/pkg/datagatherer/k8s/generic.go
@@ -3,6 +3,7 @@ package k8s
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/jetstack/preflight/pkg/datagatherer"
 	"github.com/pkg/errors"
@@ -22,6 +23,8 @@ type Config struct {
 	GroupVersionResource schema.GroupVersionResource
 	// ExcludeNamespaces is a list of namespaces to exclude.
 	ExcludeNamespaces []string `yaml:"exclude-namespaces"`
+	// IncludeNamespaces is a list of namespaces to include.
+	IncludeNamespaces []string `yaml:"include-namespaces"`
 }
 
 // UnmarshalYAML unmarshals the Config resolving GroupVersionResource.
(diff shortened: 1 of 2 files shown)


pkg/datagatherer/k8s/generic_test.go
```diff
@@ -60,14 +61,39 @@ func asUnstructuredList(items ...*unstructured.Unstructured) *unstructured.Unstr
...
+	if err != nil {
+		t.Errorf("expected no error but got: %v", err)
+	}
+
+	expected := &DataGatherer{
+		cl:                   cl,
+		groupVersionResource: config.GroupVersionResource,
+		// it's important that the namespaces are set as the IncludeNamespaces
+		// during initialization
+		namespaces: config.IncludeNamespaces,
```
charlieegan3, line 81: I'm not sure this test is a good idea. It feels a bit much tbh.

However, it's the only place we validate that the include-namespaces make it to the dg configuration...

I thought it worth adding, but I'm indifferent as to it staying. `newDataGathererWithClient` isn't ideal.

pkg/datagatherer/k8s/generic.go
```diff
@@ -51,8 +54,17 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 // validate validates the configuration.
 func (c *Config) validate() error {
+	var errors []string
+	if len(c.ExcludeNamespaces) > 0 && len(c.IncludeNamespaces) > 0 {
+		errors = append(errors, "cannot set excluded an included namespaces")
```
j-fuentes, line 59:
```suggestion
		errors = append(errors, "cannot set excluded and included namespaces")
```

charlieegan3: Fixed in d12572780845107463c22448986dd000130147e4 (manually to add commit sign-off)

PR #159, opened 2020-04-20
Change agent to support OAuth2 auth

j-fuentes:
This makes the agent support OAuth2.


j-fuentes:
I have just realised I haven't changed the Makefile to add LDFLAGS to define `clientID`, `clientSecret` and `authServer`.

Will do tomorrow.

j-fuentes:
This is ready for review :)

agentmanigest.yaml
```diff
@@ -0,0 +1,91 @@
+apiVersion: v1
```
james-w, line 1: The filename has a typo: -> `agentmanifest.yaml`.

What are these files for?

agentconfig.yaml
```diff
@@ -0,0 +1,13 @@
+schedule: "* * * * *"
```
charlieegan3, line 1: We already have agent.yaml - might be worth removing that file?

agentmanigest.yaml
```diff
@@ -0,0 +1,91 @@
...
+      kind: "k8s"
+      config:
+        resource-type:
+          resource: pods
+          version: v1
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: agent
```
charlieegan3, line 31: There is a reference to this service account here: https://github.com/jetstack/preflight/blame/master/docs/datagatherers/gke.md#L104

Now seems as good a time as any to correct it from `default`

agentmanigest.yaml
charlieegan3, line 1: At the moment, these files are documented in the readme: https://github.com/jetstack/preflight#installation

pkg/client/client.go
```diff
@@ -0,0 +1,122 @@
...
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/jetstack/preflight/api"
+)
+
+// This variables are injected at build time.
```
charlieegan3, line 13:
```suggestion
// These variables are injected at build time.
```

agentmanigest.yaml
j-fuentes: Ooops, `agentmanigest.yaml` and `agentconfig.yaml` should not be here. They are part of a manual test I was doing and I added them by mistake. Going to delete them right now.

j-fuentes: I added this file by mistake.

agentconfig.yaml
j-fuentes: I added that file by mistake. It should be deleted now.

j-fuentes:
> This mostly all looks ready to go. ~Do~ Should we update the example manifests to mount in service account credentials as part of this PR?

Not yet, as the production auth server still needs to be configured. Will do later.

pkg/client/client.go
j-fuentes: Changed in a new commit (required signoff).

PR #161, opened 2020-04-23
Fix bug with new preflight client and basic auth

j-fuentes:
We introduced a regression in #159 as we were misconfiguring the preflight client with `"(redacted)"` instead of the actual token.

This contains the fix.

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -44,6 +44,13 @@ func Run(cmd *cobra.Command, args []string) {
 		log.Fatalf("Failed to parse config file: %s", err)
 	}
 
+	// AuthToken flag takes preference over token in configuration file.
+	if AuthToken == "" {
+		AuthToken = config.Token
+	} else {
+		log.Printf("Using authorization token from flag.")
+	}
+
 	if config.Token != "" {
 		config.Token = "(redacted)"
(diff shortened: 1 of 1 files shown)


pkg/agent/run.go
```diff
@@ -44,6 +44,13 @@ func Run(cmd *cobra.Command, args []string) {
 		log.Fatalf("Failed to parse config file: %s", err)
 	}
 
+	// AuthToken flag takes preference over token in configuration file.
+	if AuthToken == "" {
+		AuthToken = config.Token
+	} else {
+		log.Printf("Using authorization token from flag.")
```
charlieegan3, line 51: Is the issue here that it'll print both `Using authorization`  then potentially `A credentials file was specified. Using OAuth2 authentication` later?

charlieegan3: I guess this is a minor issue as authtoken will be gone soon.

j-fuentes: If someone uses both the token and the credentials file, it is going to use the credentials file and yes, print both messages.

It is not ideal but I didn't want to fail hard because I am often reusing the same config file with both auth methods.

I agree it might be confusing but this is going away soon.

PR #163, opened 2020-04-24
Make agent support new endpoint for gathering data

j-fuentes:
This adds support for the new form of the endpoint that gathers data.

It introduces significant changes in the configuration, but everything should be backward compatible.

These changes are:

- The `Schedule` field is not optional, as we are using the period flag at the moment.
- The `Token` string is also optional, as it should not be specified when using OAuth2.
- The `Endpoint` is being deprecated, as in the future you won't configure the endpoint for uploading data, just preflight server. At the moment it is still being used if specified, but ignored and defaulting to `/api/v1/datareadings` otherwise.
- Added a new `Server` property that points to the Preflight server. Defaults to `https://preflight.jetstack.io` if not specified.
- Added `OrganizationID` which will be used with the new endpoint.
- Added `ClusterID`. To be left empty if using agent tokens (as they already point to a cluster).

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -1,7 +1,9 @@
 package agent
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -12,6 +14,7 @@ import (
 	"github.com/jetstack/preflight/api"
 	"github.com/jetstack/preflight/pkg/client"
 	"github.com/jetstack/preflight/pkg/datagatherer"
+	"github.com/jetstack/preflight/pkg/version"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/agent/run.go
```diff
@@ -138,14 +150,35 @@ func Run(cmd *cobra.Command, args []string) {
...
+			data, err := json.Marshal(readings)
+			if err != nil {
+				log.Fatalf("Cannot marshal readings: %+v", err)
+			}
+			path := config.Endpoint.Path
+			if path == "" {
+				path = "/api/v1/datareadings"
+			}
+			res, err := preflightClient.Post(path, bytes.NewBuffer(data))
+			if code := res.StatusCode; code < 200 || code >= 300 {
```
charlieegan3, line 164: Should 3xx responses not be ok?

charlieegan3: I guess the client will follow redirects but if we get another 300 response here we don't know what to do so what you have is good I think.

pkg/client/http.go
```diff
@@ -0,0 +1,46 @@
...
+
+func (c *PreflightClient) fullURL(path string) string {
+	base := c.baseURL
+	for strings.HasSuffix(base, "/") {
+		base = strings.TrimSuffix(base, "/")
+	}
+	for strings.HasPrefix(path, "/") {
+		path = strings.TrimPrefix(path, "/")
+	}
+	return fmt.Sprintf("%s/%s", base, path)
```
charlieegan3, line 45: can this use filepath join? like this? https://github.com/jetstack/preflight/pull/163/files#diff-fb87476fa66849c8514ba700e715b4b6R86

charlieegan3: this was only an idea so I've lgtm'd

shame prow doesn't have an option of /lgtm but let the author merge it... maybe it does...

j-fuentes: in this case, `filepath` does not work because the string contains `https://` and it assumes it is a file path, so the result ends up being `http:/domain/path` (with a single `/`).

pkg/agent/run.go
j-fuentes: I don't think so.

Maybe we can make the client handle 3xx responses. The common cases (as the redirect) are already handled by the go `http` client by default.

PR #165, opened 2020-05-06
Implementation of k8s-discovery datagatherer

charlieegan3:
This makes use of the discovery client in client-go to get the server
version information. We need this to use in upcoming packages.

https://godoc.org/k8s.io/client-go/discovery

Sadly, the fake discovery client doesn't seem to work in the same way as
the dynamic one and I've not spent the time to write tests around it.
However, it's operation is much simpler. Open to suggestions on how to
test this in the same way as the dynamic client.

This PR also updates all docs and renames the k8s datagatherer to
k8s-dynamic in a backwards compatible manner.

(Internal Issue: https://github.com/jetstack/preflight-platform/issues/350)

### DIFF
--- a/pkg/datagatherer/k8s/discovery.go
+++ b/pkg/datagatherer/k8s/discovery.go
@@ -54,6 +54,7 @@ func (g *DataGathererDiscovery) Fetch() (interface{}, error) {
 	}
 
 	response := map[string]interface{}{
+		// data has type Info: https://godoc.org/k8s.io/apimachinery/pkg/version#Info
 		"server_version": data,
 	}
 
--- a/pkg/datagatherer/k8s/discovery.go
+++ b/pkg/datagatherer/k8s/discovery.go
@@ -1,0 +1,61 @@
+package k8s
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/jetstack/preflight/pkg/datagatherer"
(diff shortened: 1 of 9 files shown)


pkg/datagatherer/k8s/discovery.go
```diff
@@ -0,0 +1,61 @@
...
+
+// Fetch will fetch discovery data from the apiserver, or return an error
+func (g *DataGathererDiscovery) Fetch() (interface{}, error) {
+	data, err := g.cl.ServerVersion()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get server version: %v", err)
+	}
+
+	response := map[string]interface{}{
+		"server_version": data,
```
j-fuentes, line 57:
```suggestion
                // data has type Info: https://godoc.org/k8s.io/apimachinery/pkg/version#Info
		"server_version": data,
```

j-fuentes:
It looks good.

Consider adding a comment pointing to the type that `ServerVersion` returns.

/lgtm
/hold

pkg/datagatherer/k8s/discovery.go
charlieegan3: :+1: fixed in e7f5c040d2ccbc31c1002d39c9d835babe58512b

PR #166, opened 2020-05-07
Fix some issues for oauth2 client

j-fuentes:
This fixes some issues related to the oauth2 client.

Also, I have moved `Credentials` to the `client` package and now it supports loading `ClientID`, `ClientSecret` and `AuthServerDomain` from the credentials file, which is useful for development.

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -75,7 +75,7 @@ func Run(cmd *cobra.Command, args []string) {
 
 	log.Printf("Loaded config: \n%s", dump)
 
-	var credentials *Credentials
+	var credentials *client.Credentials
 	if CredentialsPath != "" {
 		file, err = os.Open(CredentialsPath)
 		if err != nil {
@@ -85,7 +85,7 @@ func Run(cmd *cobra.Command, args []string) {
 
 		b, err = ioutil.ReadAll(file)
 
-		credentials, err = ParseCredentials(b)
+		credentials, err = client.ParseCredentials(b)
 		if err != nil {
 			log.Fatalf("Failed to parse credentials file: %s", err)
 		}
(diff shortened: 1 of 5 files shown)


pkg/agent/run.go
```diff
@@ -85,7 +85,7 @@ func Run(cmd *cobra.Command, args []string) {
 
 		b, err = ioutil.ReadAll(file)
 
-		credentials, err = ParseCredentials(b)
+		credentials, err = client.ParseCredentials(b)
```
charlieegan3, line 88: Perhaps also validate credentials at this point?

pkg/client/client.go
```diff
@@ -47,29 +45,34 @@ func NewWithBasicAuth(agentMetadata *api.AgentMetadata, authToken, baseURL strin
 }
 
 // New creates a new client that uses OAuth2.
-func New(agentMetadata *api.AgentMetadata, userID, userSecret, baseURL string) (*PreflightClient, error) {
-	if userID == "" || userSecret == "" {
+func New(agentMetadata *api.AgentMetadata, credentials *Credentials, baseURL string) (*PreflightClient, error) {
+	if credentials == nil || credentials.UserID == "" || credentials.UserSecret == "" {
```
charlieegan3, line 49: Is it worth validating the userID and secret are set earlier than this function? (see note above)

charlieegan3: Shouldn't be nil here because of https://github.com/jetstack/preflight/pull/166/files#diff-d1aa418498ff92dc3def120d994c0379R98 right?

pkg/agent/run.go
j-fuentes: `ParseCredentials` does validation.

charlieegan3: But not `nil || credentials.UserID == "" || credentials.UserSecret == ""`?

charlieegan3: (my concern was that we appeared to be validating the credentials when creating the client instead - but perhaps a non issue)

pkg/client/client.go
j-fuentes: Oh, I forgot we had a validation method for `Credentials`.

In that case, I am going to change this so `New` checks `credentials` is not `nil` and it is indeed valid credentials.

Having `New` do that sanity check is a good thing, as this package could be used somewhere else.

charlieegan3: :+1: sounds good.

PR #169, opened 2020-05-08
Pass oauth flags to docker build

j-fuentes:


charlieegan3:
After our discussion this morning, I think we're not going this route any more?

j-fuentes:
> After our discussion this morning, I think we're not going this route any more?

Finally, we have opted to continue down this path and add the OAuth clientID and clientSecret to the Makefile. So this PR still applies.

PR #176, opened 2020-06-17
Add flag for number periods gathering/posting data

JammyL:
Number of periods is integer.
Takes -1 as default (runs perpetually) - Any -ve will do this

### DIFF
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -57,12 +57,12 @@ func init() {
 		"",
 		"Location of the credentials file. For OAuth2 based authentication.",
 	)
-	agentCmd.PersistentFlags().IntVarP(
-		&agent.NumberPeriods,
-		"number-periods",
-		"n",
-		-1,
-		"Number of interations for data collection. Default is '-1', runs perpetually.",
+	agentCmd.PersistentFlags().BoolVarP(
+		&agent.OneShot,
+		"one-shot",
+		"o",
+		false,
+		"Runs agent a single time if true, or continously if false",
 	)
 
 }
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -57,4 +57,12 @@ func init() {
 		"",
 		"Location of the credentials file. For OAuth2 based authentication.",
 	)
+	agentCmd.PersistentFlags().IntVarP(
+		&agent.NumberPeriods,
+		"number-periods",
+		"n",
+		-1,
+		"Number of interations for data collection. Default is '-1', runs perpetually.",
+	)
+
 }
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -60,7 +60,7 @@ func init() {
 	agentCmd.PersistentFlags().BoolVarP(
 		&agent.OneShot,
 		"one-shot",
-		"o",
+		"",
 		false,
 		"Runs agent a single time if true, or continously if false",
 	)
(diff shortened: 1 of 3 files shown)


pkg/agent/run.go
```diff
@@ -28,16 +28,30 @@ var AuthToken string
...
+		if i == NumberPeriods - 1 {
+			break
+		}
+
 		time.Sleep(time.Duration(Period) * time.Second)
+
+		// Progress loop if a positive number of periods is given
+		if NumberPeriods > 0 {
+			i++
+		}
```
james-w, line 54: I don't think this is necessary, except that it avoids the agent stopping if it runs for long enough to overflow?

charlieegan3:
I've written up an issue here: https://github.com/jetstack/preflight/issues/177

I'm aware it might be slightly different from the task, however I think it's worth having the discussion there about the requirements as I think there might be a simpler implementation.

pkg/agent/run.go
JammyL: Yep, if that's not an issue I can remove it :)

james-w: I think that's not a bad thing to have.

The only reasons not to have it would be:

1) We don't have a counter of the number of loops in case we ever wanted to use that. If we add a counter then that can overflow and maybe cause problems.

2) It's extra lines of code, and fewer lines of code is preferable, all else being equal

`i` is an `int` here, which is either a 32 or 64 bit signed int, depending on the system.
In the worst case of it being a 32 bit signed the max is `2^31 -1`. By my calculations
if we sleep for 1 minute per loop that would mean it would overflow after 4000 years.

Therefore I think it's never really going to happen, and I would err on leaving out
the check in order to have less code, but it's hard to explain that in the code and
that there could be a problem lurking there if the context of this code changes.

JammyL: Good point! In line with this and Charlie's comment I'll switch over to a 'one-shot' flag

cmd/agent.go
```diff
@@ -57,4 +57,12 @@ func init() {
 		"",
 		"Location of the credentials file. For OAuth2 based authentication.",
 	)
+	agentCmd.PersistentFlags().BoolVarP(
+		&agent.OneShot,
+		"one-shot",
+		"o",
```
james-w, line 63: `-o` is often used for `--output`. I don't know what `--output` would do for the agent, but if we add it later
then it would be unexpected for it not to be `-o`.

I don't think that we need a short option for `--one-shot`, so I would just have the long option here, with
no short option.

pkg/agent/run.go
```diff
@@ -28,16 +28,20 @@ var AuthToken string
...
 // Run starts the agent process
 func Run(cmd *cobra.Command, args []string) {
 	ctx := context.Background()
-
-	for {
-		gatherAndPostData(ctx)
+	gatherAndPostData(ctx)
+	for OneShot != true {
 		time.Sleep(time.Duration(Period) * time.Second)
+		gatherAndPostData(ctx)
```
james-w, line 44: This is fine as it is, however I prefer to not repeat the "meat" of the loop. It's only one function
call today, but we might want to change that later. Therefore I would choose to use a for loop
with `if OneShot { break }` in it. You don't have to change anything though.

charlieegan3:
Looks good to me!

/lgtm
/approve

PR #180, opened 2020-06-29
Adds write-to-file flag to store data locally

JammyL:
Flag --write-to-file allows for a local path to be given to which data is stored.
Only stores data locally if no organization is specified. - This can be added for when organisation is specified if needed.

Signed-off-by: Jamie Leppard <JammyL@users.noreply.github.com>

### DIFF
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -64,5 +64,12 @@ func init() {
 		false,
 		"Runs agent a single time if true, or continously if false",
 	)
+	agentCmd.PersistentFlags().StringVarP(
+		&agent.DataFilePath,
+		"write-to-file",
+		"",
+		"",
+		"Data file path, if used will write data to a local file instead of posting",
+	)
 
 }
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -65,11 +65,11 @@ func init() {
 		"Runs agent a single time if true, or continously if false",
 	)
 	agentCmd.PersistentFlags().StringVarP(
-		&agent.DataFilePath,
-		"write-to-file",
+		&agent.OutputPath,
+		"output-path",
 		"",
 		"",
-		"Data file path, if used will write data to a local file instead of posting",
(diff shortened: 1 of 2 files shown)


cmd/agent.go
```diff
@@ -64,5 +64,12 @@ func init() {
 		false,
 		"Runs agent a single time if true, or continously if false",
 	)
+	agentCmd.PersistentFlags().StringVarP(
+		&agent.DataFilePath,
+		"write-to-file",
```
charlieegan3, line 69: Do you think `--output-path` might be more descriptive?

cmd/agent.go
```diff
@@ -64,5 +64,12 @@ func init() {
...
+		"",
+		"",
+		"Data file path, if used will write data to a local file instead of posting",
```
charlieegan3, line 72: `...instead of posting` -. `...instead of uploading to the preflight server`

pkg/agent/run.go
```diff
@@ -176,27 +178,35 @@ func gatherAndPostData(ctx context.Context) {
 	log.Println("Running Agent...")
 	log.Println("Posting data to ", baseURL)
 	if config.OrganizationID == "" {
```
charlieegan3, line 180: I think we might need to handle the case when the org is set too?

pkg/agent/run.go
```diff
@@ -176,27 +178,35 @@ func gatherAndPostData(ctx context.Context) {
...
-		data, err := json.Marshal(readings)
+		data, err := json.MarshalIndent(readings, " ", " ")
```
charlieegan3, line 181: Ideally, we only indent the json when we're writing to disk. Data sent over the wire should be as small as possible.

cmd/agent.go
JammyL: Yes, that's a better name

pkg/agent/run.go
JammyL: Yes, I think this should be a part of the larger changes made with the load-from-file addition

PR #185, opened 2020-07-06
Edits data gathering process to retry dataGatherer

JammyL:
A ticker manages an exponential backoff algorithm
This uses: https://pkg.go.dev/github.com/cenkalti/backoff/

The variables e.g. maxtime or intervals can be changed through backOff.

Successful dataGatherers are tracked through a map.
Only those not in the map are retried.

WIP: Wrap this in a function which can be tested more rigorously

Signed-off-by: Jamie Leppard <JammyL@users.noreply.github.com>

### DIFF
--- a/agent.yaml
+++ b/agent.yaml
@@ -9,3 +9,8 @@ data-gatherers:
   name: "dummy"
   config:
     param-1: "bar"
+    failed-attempts: 5
+- kind: "dummy"
+  name: "dummy-fail"
+  config:
+    param-1: "foo"
--- a/agent.yaml
+++ b/agent.yaml
@@ -8,9 +8,8 @@ data-gatherers:
 - kind: "dummy"
   name: "dummy"
   config:
-    param-1: "bar"
     failed-attempts: 5
 - kind: "dummy"
   name: "dummy-fail"
   config:
-    param-1: "foo"
+    always-fail: true
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/jetstack/preflight/pkg/agent"
 	"github.com/spf13/cobra"
@@ -43,12 +44,12 @@ func init() {
 		"",
 		"Authorization token. If used, it will override the authorization token in the configuration file.",
 	)
-	agentCmd.PersistentFlags().UintVarP(
+	agentCmd.PersistentFlags().DurationVarP(
 		&agent.Period,
 		"period",
 		"p",
-		3600,
-		"Time between scans, in seconds.",
+		time.Hour,
+		"Time between scans (given as XhYmZs).",
 	)
 	agentCmd.PersistentFlags().StringVarP(
 		&agent.CredentialsPath,
@@ -77,6 +78,13 @@ func init() {
 		"",
 		"",
 		"Input file path, if used, it will read data from a local file instead of gathering data from clusters",
+	)
+	agentCmd.PersistentFlags().DurationVarP(
+		&agent.BackoffMaxTime,
+		"backoff-max-time",
+		"",
+		5*time.Minute,
+		"Max time for retrying failed data gatherers (given as XhYmZs).",
 	)
 
 }
--- a/pkg/agent/dummy_data_gatherer.go
+++ b/pkg/agent/dummy_data_gatherer.go
@@ -8,8 +8,8 @@ import (
 )
 
 type dummyConfig struct {
-	Param1            string `yaml:"param-1"`
-	FailedAttempts    int    `yaml:"failed-attempts"`
+	AlwaysFail        bool `yaml:"always-fail"`
+	FailedAttempts    int  `yaml:"failed-attempts"`
 	wantOnCreationErr bool
 }
 
@@ -18,13 +18,13 @@ func (c *dummyConfig) NewDataGatherer(ctx context.Context) (datagatherer.DataGat
 		return nil, fmt.Errorf("an error")
 	}
 	return &dummyDataGatherer{
-		Param1:         c.Param1,
+		AlwaysFail:     c.AlwaysFail,
 		FailedAttempts: c.FailedAttempts,
 	}, nil
 }
 
 type dummyDataGatherer struct {
-	Param1         string
+	AlwaysFail     bool
 	attemptNumber  int
 	FailedAttempts int
 }
@@ -34,8 +34,8 @@ func (c *dummyDataGatherer) Fetch() (interface{}, error) {
 	if c.attemptNumber < c.FailedAttempts {
 		err = fmt.Errorf("First %d attempts will fail", c.FailedAttempts)
 	}
-	if c.Param1 == "foo" {
-		err = fmt.Errorf("Param1 cannot be foo")
+	if c.AlwaysFail {
+		err = fmt.Errorf("This data gatherer will always fail")
 	}
 	c.attemptNumber++
 	return nil, err
--- a/pkg/agent/dummy_data_gatherer.go
+++ b/pkg/agent/dummy_data_gatherer.go
@@ -9,6 +9,7 @@ import (
 
 type dummyConfig struct {
 	Param1            string `yaml:"param-1"`
+	FailedAttempts    int    `yaml:"failed-attempts"`
 	wantOnCreationErr bool
 }
 
@@ -17,14 +18,25 @@ func (c *dummyConfig) NewDataGatherer(ctx context.Context) (datagatherer.DataGat
 		return nil, fmt.Errorf("an error")
 	}
 	return &dummyDataGatherer{
-		Param1: c.Param1,
+		Param1:         c.Param1,
+		FailedAttempts: c.FailedAttempts,
 	}, nil
 }
 
 type dummyDataGatherer struct {
-	Param1 string
+	Param1         string
+	attemptNumber  int
+	FailedAttempts int
 }
 
 func (c *dummyDataGatherer) Fetch() (interface{}, error) {
-	return nil, nil
+	var err error
+	if c.attemptNumber < c.FailedAttempts {
+		err = fmt.Errorf("First %d attempts will fail", c.FailedAttempts)
+	}
+	if c.Param1 == "foo" {
+		err = fmt.Errorf("Param1 cannot be foo")
+	}
+	c.attemptNumber++
+	return nil, err
 }
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -12,6 +12,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cenkalti/backoff"
+	"github.com/hashicorp/go-multierror"
 	"github.com/jetstack/preflight/api"
 	"github.com/jetstack/preflight/pkg/client"
 	"github.com/jetstack/preflight/pkg/datagatherer"
@@ -25,8 +27,8 @@ var ConfigFilePath string
 // AuthToken is the authorization token that will be used for API calls
 var AuthToken string
 
-// Period is the number of seconds between scans
-var Period uint
+// Period is the time waited between scans
+var Period time.Duration
 
 // OneShot flag causes agent to run once
 var OneShot bool
@@ -40,6 +42,9 @@ var OutputPath string
 // InputPath is where the agent will read data from instead of gathering from clusters if specified
 var InputPath string
 
+// BackoffMaxTime is the maximum time for which data gatherers will be retried
+var BackoffMaxTime time.Duration
+
 // Run starts the agent process
 func Run(cmd *cobra.Command, args []string) {
 	ctx := context.Background()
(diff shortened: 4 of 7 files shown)


pkg/agent/dummy_data_gatherer.go
```diff
@@ -22,9 +22,18 @@ func (c *dummyConfig) NewDataGatherer(ctx context.Context) (datagatherer.DataGat
...
 type dummyDataGatherer struct {
-	Param1 string
+	Param1        string
+	attemptNumber int
 }
 
 func (c *dummyDataGatherer) Fetch() (interface{}, error) {
-	return nil, nil
+	var err error
+	if c.attemptNumber < 2 {
```
j-fuentes, line 31: This should be configurable, not hardcoded.

I suggest that you add a config parameter to `dummyConfig` that let's us configure the behaviour from the configuration file.

pkg/agent/run.go
```diff
@@ -184,37 +184,61 @@ func gatherData(ctx context.Context, config Config) []*api.DataReading {
 		dataGatherers[dgConfig.Name] = dg
 	}
 
-	// Fetch from all datagatherers
-	now := time.Now()
+	//TODO Change backoff parameters to those desired
+	backOff := backoff.NewExponentialBackOff()
+	backOff.MaxElapsedTime = 30 * time.Second
```
j-fuentes, line 189: This should come from a CLI flag with a default value.

pkg/agent/run.go
```diff
@@ -184,37 +184,61 @@ func gatherData(ctx context.Context, config Config) []*api.DataReading {
...
-	for k, dg := range dataGatherers {
-		i, err := dg.Fetch()
-		if err != nil {
-			log.Printf("Error fetching with DataGatherer %q: %s", k, err)
-			failedDataGatherers = append(failedDataGatherers, k)
-			continue
-		}
-
-		log.Printf("Gathered data for %q:\n", k)
+	completedDataGatherers := make(map[string]bool)
```
j-fuentes, line 191: Since you already know the final size of the map (`len(dataGatherers)` in an ideal world), instantiate it with that size `make(map[string]bool, len(dataGatherers))`.

pkg/agent/run.go
```diff
@@ -184,37 +184,61 @@ func gatherData(ctx context.Context, config Config) []*api.DataReading {
...
+			if completedDataGatherers[k] {
+				continue
+			}
+			dgData, err := dg.Fetch()
+			if err != nil {
+				if fullError == nil {
+					fullError = fmt.Errorf("%s: %v", k, err)
+				} else {
+					fullError = fmt.Errorf("%v\n%s: %v", fullError, k, err)
+				}
```
j-fuentes, line 206: Instead of building the string in this complicated way, I suggest that you replace `fullError error` with `dgErrors map[string]error`, where the key is the different `k`s.

So in case of error you just need to add a new element to the map.

Then, in the end, you can create a new error and format the message as you prefer.

j-fuentes: I see we are already using `go-multierror` in the project. In that case, it's best to just use it: https://github.com/hashicorp/go-multierror.

pkg/agent/dummy_data_gatherer.go
JammyL: That's a good point! I hadn't thought about the most general use case there. I can add that to the data gatherers parameters. i.e. `failedAttempts int`

pkg/agent/run.go
JammyL: Good idea, what would you suggest for a default max elapsed time, and also default interval?

pkg/agent/run.go
JammyL: Will do :)

pkg/agent/run.go
JammyL: I've set the default to 60 seconds for now

cmd/agent.go
```diff
@@ -78,5 +78,12 @@ func init() {
 		"",
 		"Input file path, if used, it will read data from a local file instead of gathering data from clusters",
 	)
+	agentCmd.PersistentFlags().UintVarP(
+		&agent.BackoffMaxTime,
+		"backoff-max-time",
+		"",
+		60,
+		"Max time for retrying failed data gatherers. ",
```
james-w, line 86: This should indicate the unit of the time, e.g.
```suggestion
		"Max time for retrying failed data gatherers (in seconds).",
```

pkg/agent/run.go
```diff
@@ -170,7 +175,18 @@ func gatherAndOutputData(ctx context.Context, config Config, preflightClient *cl
 		}
 		log.Println("Data saved locally to", OutputPath)
 	} else {
-		postData(config, preflightClient, readings)
+		//TODO Change backoff parameters to those desired
+		backOff := backoff.NewExponentialBackOff()
+		backOff.MaxElapsedTime = time.Duration(BackoffMaxTime) * time.Second
```
james-w, line 180: I don't think you need `time.Duration()` here, I think that you can just multiply `time.Second` by the number of seconds.

pkg/agent/run.go
```diff
@@ -40,6 +42,9 @@ var OutputPath string
 // InputPath is where the agent will read data from instead of gathering from clusters if specified
 var InputPath string
 
+// BackoffMaxTime is the maximum time for which data gatherers will be retried
+var BackoffMaxTime uint
```
james-w, line 46: Just an idea, but this could be a `time.Duration` so that it can be used wherever without worrying about converting it, and there's no ambiguity as to the units.

pkg/agent/run.go
```diff
@@ -170,7 +175,18 @@ func gatherAndOutputData(ctx context.Context, config Config, preflightClient *cl
 		}
 		log.Println("Data saved locally to", OutputPath)
 	} else {
-		postData(config, preflightClient, readings)
+		//TODO Change backoff parameters to those desired
```
james-w, line 178: What does this mean?

pkg/agent/run.go
```diff
@@ -170,7 +175,18 @@ func gatherAndOutputData(ctx context.Context, config Config, preflightClient *cl
...
+		backOff := backoff.NewExponentialBackOff()
+		backOff.MaxElapsedTime = time.Duration(BackoffMaxTime) * time.Second
+		post := func() error {
+			err := postData(config, preflightClient, readings)
+			return err
```
james-w, line 183: This could work as just

```suggestion
			return postData(config, preflightClient, readings)
```

but it doesn't really make a difference.

pkg/agent/run.go
```diff
@@ -193,39 +209,67 @@ func gatherData(ctx context.Context, config Config) []*api.DataReading {
...
 
-	if len(failedDataGatherers) > 0 {
-		log.Printf(
-			"Warning, the following DataGatherers failed, %s. Their data is not being sent.",
-			strings.Join(failedDataGatherers, ", "),
-		)
+	err := backoff.RetryNotify(getReadings, backOff, notify)
+	if err != nil {
+		log.Println(err)
+		log.Printf("This will not be retried")
```
james-w, line 259: Definitely not for this change, but it feels like we want to extend the datareadings protocol to be able to include errors for particular gatherers. That would then allow us to expose that to the user in some way.

pkg/agent/dummy_data_gatherer.go
```diff
@@ -17,14 +18,25 @@ func (c *dummyConfig) NewDataGatherer(ctx context.Context) (datagatherer.DataGat
...
+	FailedAttempts int
 }
 
 func (c *dummyDataGatherer) Fetch() (interface{}, error) {
-	return nil, nil
+	var err error
+	if c.attemptNumber < c.FailedAttempts {
+		err = fmt.Errorf("First %d attempts will fail", c.FailedAttempts)
+	}
+	if c.Param1 == "foo" {
```
j-fuentes, line 37: do you really need `Param1` now? You can recreate the failure just with `FailedAttemps`, don't you?

pkg/agent/run.go
j-fuentes, line 183: For simplicity:
```
return postData(config, preflightClient, readings)
```

pkg/agent/run.go
j-fuentes, line 178: What do you mean with this TODO?

pkg/agent/run.go
JammyL: I tried it as an integer and Go didn't like it. I think your other comment is the way to go with initialising as `time.Duration()` is the way to go

pkg/agent/run.go
JammyL: The `*exponentialBackoff` takes certain parameters such as initial interval, and max elapsed time. We can make them all configurable (I think the config file would be the best way of doing that). But I think its best we agree on the optimal default values for the agent

pkg/agent/run.go
JammyL: I'm not sure what you mean by this? At the moment the errors are reported for each failed data gatherer.

pkg/agent/dummy_data_gatherer.go
JammyL: Yes, you could set failedAttempts to a very large number and Param1 would become redundant. I didn't add it in however, so I'm not sure if its there for some other reason

pkg/agent/run.go
JammyL: The *exponentialBackoff takes certain parameters such as initial interval, and max elapsed time. We can make them all configurable (I think the config file would be the best way of doing that). But I think its best we agree on the optimal default values for the agent

pkg/agent/run.go
JammyL: Ah yes, that looks nicer

JammyL: That does look nicer

pkg/agent/run.go
JammyL: The same has been done for the `period` command line argument. I'll update that too

pkg/agent/run.go
james-w: Sorry, I mean reported to the server.

JammyL: Ah I see! That's a good point. So something along the lines of including errors as a reading itself?

james-w: Yes, something like that. Then the server can tell the difference between "the agent didn't read this" and "the agent wanted to read this and failed" and that will be useful to us in improving the information available to the users about which reports they can view.

pkg/agent/run.go
james-w: I would suggest initial interval of 5s, with max elapsed time 5m.

cmd/agent.go
```diff
@@ -78,5 +79,12 @@ func init() {
 		"",
 		"Input file path, if used, it will read data from a local file instead of gathering data from clusters",
 	)
+	agentCmd.PersistentFlags().DurationVarP(
+		&agent.BackoffMaxTime,
+		"backoff-max-time",
+		"",
+		time.Minute,
+		"Max time for retrying failed data gatherers (in seconds).",
```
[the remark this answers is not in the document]
JammyL: Yep, so the flag is now formatted as 'X'h'Y'm'Z's 
You can leave any of them out. 
If you just say 10 cobra gives you back an error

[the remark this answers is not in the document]
JammyL: Oh yes of course, I should change that comment to say XhYmZs

cmd/agent.go
```diff
@@ -43,11 +44,11 @@ func init() {
...
 		"Authorization token. If used, it will override the authorization token in the configuration file.",
 	)
-	agentCmd.PersistentFlags().UintVarP(
+	agentCmd.PersistentFlags().DurationVarP(
 		&agent.Period,
 		"period",
 		"p",
-		3600,
+		time.Hour,
 		"Time between scans, in seconds.",
```
james-w, line 52: Ah, this has the same issue in the help now.

JammyL: Good spot!

pkg/agent/run.go
```diff
@@ -170,7 +175,17 @@ func gatherAndOutputData(ctx context.Context, config Config, preflightClient *cl
 		}
 		log.Println("Data saved locally to", OutputPath)
 	} else {
-		postData(config, preflightClient, readings)
+		//TODO Change backoff parameters to those desired
```
james-w, line 178: What do we want to do here? I think it's better to set something than to have a TODO, unless the defaults are fine, then I would just remove the TODO.

pkg/agent/dummy_data_gatherer.go
j-fuentes: if we want a "fail always" option, I would rename `Param1` to `FailAlways`

agent.yaml
```diff
@@ -9,3 +9,8 @@ data-gatherers:
   name: "dummy"
   config:
     param-1: "bar"
```
j-fuentes, line 11: this line can go now I think

JammyL: Oh yeah, missed that one!

PR #189, opened 2020-07-15
Retain certs when posting tls secrets

charlieegan3:
This allows crt content to be used in package rules.


charlieegan3:
I've made this a draft until I work out more how the server side of this works. I'm not sure opa can parse PEM encoded certs.

charlieegan3:
https://github.com/open-policy-agent/opa/pull/2564 has been merged so we can move forward with this now.

PR #195, opened 2020-08-13
Remove basic auth from agent codebase

tfadeyi:
Removes basic auth from the preflight client inialization.
Still allows preflight client to be not auth.

Related #414

Signed-off-by: Oluwole Fadeyi <tfadeyi@users.noreply.github.com>

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -18,9 +18,6 @@ import (
 // Config wraps the options for a run of the agent.
 type Config struct {
 	Schedule string `yaml:"schedule"`
-	// Token is the agent token if using basic authentication.
-	// If not provided it will assume OAuth2 authentication.
-	Token string `yaml:"token"`
 	// Deprecated: Endpoint is being replaced with Server.
 	Endpoint Endpoint `yaml:"endpoint"`
 	// Server is the base url for the Preflight server.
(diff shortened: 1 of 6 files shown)


tfadeyi:
@j-fuentes should we remove the token field from the agent's `config` yaml ?

j-fuentes:
> @j-fuentes should we remove the token field from the agent's `config` yaml ?

Yes, we should, as it is useless now.

pkg/agent/config.go
```diff
@@ -130,15 +127,6 @@ func (c *Config) Dump() (string, error) {
...
 
-	if c.Token == "" {
-		if c.OrganizationID == "" {
-			result = multierror.Append(result, fmt.Errorf("organization_id is required"))
-		}
-		if c.ClusterID == "" {
-			result = multierror.Append(result, fmt.Errorf("cluster_id is required"))
-		}
-	}
-
```
j-fuentes, line 141: The checks inside the `if` should be applied all the times now.

pkg/agent/config_test.go
```diff
@@ -103,36 +99,12 @@ func TestInvalidConfigError(t *testing.T) {
 	}
 }
 
-func TestMissingConfigError(t *testing.T) {
```
j-fuentes, line 106: This test is still a valid test after deleting the token, isn't it?

pkg/agent/config.go
tfadeyi: Oh Right! I messed up there, I'll add the conditions back

pkg/agent/config_test.go
tfadeyi: Yeah, for some reason I thought that org and cluster ids conditionals weren't needed. But the other conditions in the test don't really throw any error if they are missing

tfadeyi: I'll add it back in the next commit since it's valid

j-fuentes:
:+1: 
/lgtm

PR #196, opened 2020-08-19
Add strict flag to agent

tfadeyi:
Agent will hard fail if a data gatherer is defined for a missing CRD.

Related [#510](https://github.com/jetstack/preflight-platform/issues/510)

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -42,6 +42,9 @@ var InputPath string
 // BackoffMaxTime is the maximum time for which data gatherers will be retried
 var BackoffMaxTime time.Duration
 
+// StrictMode flag causes the agent to fail at the first attempt
+var StrictMode bool
+
 // Run starts the agent process
 func Run(cmd *cobra.Command, args []string) {
 	ctx := context.Background()
@@ -199,7 +202,7 @@ func gatherData(ctx context.Context, config Config) []*api.DataReading {
 	completedDataGatherers := make(map[string]bool, len(dataGatherers))
 
 	// Fetch from all datagatherers
-	getReadings := func() error {
+	getReadings := func() *multierror.Error {
 		var dgError *multierror.Error
 		for k, dg := range dataGatherers {
 			if completedDataGatherers[k] {
@@ -235,16 +238,30 @@ func gatherData(ctx context.Context, config Config) []*api.DataReading {
 					len(es), strings.Join(points, "\n\t"))
 			}
 		}
-		return dgError.ErrorOrNil()
+		return dgError
 	}
 
-	err := backoff.RetryNotify(getReadings, backOff, notify)
-	if err != nil {
-		log.Println(err)
-		log.Printf("This will not be retried")
+	if StrictMode {
+		multiError := getReadings()
+		// check all the errors and find if it's due to a missing dg or not
+		for _, err := range multiError.WrappedErrors() {
+			if !strings.Contains(err.Error(), "the server could not find the requested resource") {
+				log.Fatalf("%v", err)
+			}
+		}
+		if err := multiError.ErrorOrNil(); err != nil {
+			log.Printf("%v", err)
+		}
 	} else {
-		log.Printf("All data gatherers successfull")
+		err := backoff.RetryNotify(func() error { return getReadings().ErrorOrNil() }, backOff, notify)
+		if err != nil {
+			log.Println(err)
+			log.Printf("This will not be retried")
+		} else {
+			log.Printf("All data gatherers successfull")
+		}
 	}
+
 	return readings
 }
 
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -17,7 +17,7 @@ import (
 	"github.com/jetstack/preflight/api"
 	"github.com/jetstack/preflight/pkg/client"
 	"github.com/jetstack/preflight/pkg/datagatherer"
-	"github.com/jetstack/preflight/pkg/datagatherer/local"
+	dgerror "github.com/jetstack/preflight/pkg/datagatherer/error"
 	"github.com/jetstack/preflight/pkg/version"
 	"github.com/spf13/cobra"
 )
@@ -211,7 +211,7 @@ func gatherData(ctx context.Context, config Config) []*api.DataReading {
 			}
 			dgData, err := dg.Fetch()
 			if err != nil {
-				if _, ok := err.(*local.ConfigError); ok {
+				if _, ok := err.(*dgerror.ConfigError); ok {
 					if StrictMode {
 						err = fmt.Errorf("%s: %v", k, err)
 						dgError = multierror.Append(dgError, err)
@@ -262,7 +262,7 @@ func gatherData(ctx context.Context, config Config) []*api.DataReading {
 			log.Println(err)
 			log.Printf("This will not be retried")
 		} else {
-			log.Printf("All data gatherers successfull")
+			log.Printf("Finished gathering data")
 		}
 	}
 
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -17,7 +17,7 @@ import (
 	"github.com/jetstack/preflight/api"
 	"github.com/jetstack/preflight/pkg/client"
 	"github.com/jetstack/preflight/pkg/datagatherer"
-	"github.com/jetstack/preflight/pkg/datagatherer/k8s"
+	"github.com/jetstack/preflight/pkg/datagatherer/local"
 	"github.com/jetstack/preflight/pkg/version"
 	"github.com/spf13/cobra"
 )
@@ -211,7 +211,7 @@ func gatherData(ctx context.Context, config Config) []*api.DataReading {
 			}
 			dgData, err := dg.Fetch()
 			if err != nil {
-				if _, ok := err.(*k8s.CRDNotFoundError); ok {
+				if _, ok := err.(*local.ConfigError); ok {
 					if StrictMode {
 						err = fmt.Errorf("%s: %v", k, err)
 						dgError = multierror.Append(dgError, err)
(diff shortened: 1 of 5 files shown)


tfadeyi:
Agent will hard fail if a data gatherer is defined for a missing CRD.

i.e: I've added a data gatherer for an incorrect CRD. `certificrequests`  instead of `certificaterequests`
```
    - kind: "k8s-dynamic"
      name: "k8s/certificrequests"
      config:
        resource-type:
          group: cert-manager.io
          version: v1alpha2
          resource: certificrequests
```


<details>
  <summary>without `--strict`</summary>

```
2020/08/20 13:28:09 The following 1 data gatherer(s) have failed:
        * k8s/certificrequests: certificrequests.cert-manager.io is forbidden: User "system:serviceaccount:preflight:agent" cannot list resource "certificrequests" in API group "cert-manager.io" at the cluster scope 
Retrying...
2020/08/20 13:28:10 The following 1 data gatherer(s) have failed:
        * k8s/certificrequests: certificrequests.cert-manager.io is forbidden: User "system:serviceaccount:preflight:agent" cannot list resource "certificrequests" in API group "cert-manager.io" at the cluster scope 
Retrying...
2020/08/20 13:28:10 The following 1 data gatherer(s) have failed:
        * k8s/certificrequests: certificrequests.cert-manager.io is forbidden: User "system:serviceaccount:preflight:agent" cannot list resource "certificrequests" in API group "cert-manager.io" at the cluster scope 
Retrying...
2020/08/20 13:28:12 The following 1 data gatherer(s) have failed:
        * k8s/certificrequests: certificrequests.cert-manager.io is forbidden: User "system:serviceaccount:preflight:agent" cannot list resource "certificrequests" in API group "cert-manager.io" at the cluster scope 
Retrying...
2020/08/20 13:28:13 The following 1 data gatherer(s) have failed:
        * k8s/certificrequests: certificrequests.cert-manager.io is forbidden: User "system:serviceaccount:preflight:agent" cannot list resource "certificrequests" in API group "cert-manager.io" at the cluster scope 
Retrying...
2020/08/20 13:28:15 The following 1 data gatherer(s) have failed:
        * k8s/certificrequests: certificrequests.cert-manager.io is forbidden: User "system:serviceaccount:preflight:agent" cannot list resource "certificrequests" in API group "cert-manager.io" at the cluster scope 
Retrying...
2020/08/20 13:28:18 The following 1 data gatherer(s) have failed:
        * k8s/certificrequests: certificrequests.cert-manager.io is forbidden: User "system:serviceaccount:preflight:agent" cannot list resource "certificrequests" in API group "cert-manager.io" at the cluster scope 
Retrying...
2020/08/20 13:28:23 The following 1 data gatherer(s) have failed:
        * k8s/certificrequests: certificrequests.cert-manager.io is forbidden: User "system:serviceaccount:preflight:agent" cannot list resource "certificrequests" in API group "cert-manager.io" at the cluster scope 
Retrying...
2020/08/20 13:28:33 The following 1 data gatherer(s) have failed:
        * k8s/certificrequests: certificrequests.cert-manager.io is forbidden: User "system:serviceaccount:preflight:agent" cannot list resource "certificrequests" in API group "cert-manager.io" at the cluster scope 
Retrying...
2020/08/20 13:28:41 The following 1 data gatherer(s) have failed:
        * k8s/certificrequests: certificrequests.cert-manager.io is forbidden: User "system:serviceaccount:preflight:agent" cannot list resource "certificrequests" in API group "cert-manager.io" at the cluster scope 
Retrying...
```

</details>

<details>
  <summary>with `--strict`</summary>

```
2020/08/20 13:29:56 A credentials file was specified. Using OAuth2 authentication...
2020/08/20 13:29:56 Successfully gathered data for "k8s/secrets"
2020/08/20 13:29:56 Successfully gathered data for "k8s/certificates"
2020/08/20 13:29:56 Successfully gathered data for "k8s/ingresses"
2020/08/20 13:29:56 The following 1 data gatherer(s) have failed:
        * k8s/certificrequests: certificrequests.cert-manager.io is forbidden: User "system:serviceaccount:preflight:agent" cannot list resource "certificrequests" in API group "cert-manager.io" at the cluster scope

NAME                     READY   STATUS    RESTARTS   AGE
agent-84f6fdf49c-k7qpf   0/1     Error     2          22s
```


Related [#510](https://github.com/jetstack/preflight-platform/issues/510)

j-fuentes:
This solves one problem, which is how to verify in a one-off run if there is something wrong gathering the data.

There is an additional problem we want to solve, which is: the default configuration will make the agent try to gather custom CRDs. We don't want the agent to fail hard, because we want the default configuration to work on as many cases as possible, but at the same time, we don't want the agent to retry that specific data-gatherer if it is clear that the CRD is not present.

So, what it is missing is to actually handle as a special error the case of a CRD not being present.

tfadeyi:
new logs if a dg is missing
<details>
<summary>with `--strict`</summary>

```
input-path: ""
output-path: ""
2020/08/21 08:31:58 A credentials file was specified. Using OAuth2 authentication...
2020/08/21 08:31:58 Successfully gathered data for "k8s/secrets"
2020/08/21 08:31:58 Successfully gathered data for "k8s/ingresses"
2020/08/21 08:31:58 The following 2 data gatherer(s) have failed:
        * k8s/certificates: the server could not find the requested resource
        * k8s/certificaterequests: the server could not find the requested resource
2020/08/21 08:31:58 Running Agent...
2020/08/21 08:31:58 Posting data to  http://a0779f5054d6.ngrok.io
```

pkg/agent/run.go
```diff
@@ -235,16 +238,30 @@ func gatherData(ctx context.Context, config Config) []*api.DataReading {
...
 
-	err := backoff.RetryNotify(getReadings, backOff, notify)
-	if err != nil {
-		log.Println(err)
-		log.Printf("This will not be retried")
+	if StrictMode {
+		multiError := getReadings()
+		// check all the errors and find if it's due to a missing dg or not
+		for _, err := range multiError.WrappedErrors() {
+			if !strings.Contains(err.Error(), "the server could not find the requested resource") {
```
j-fuentes, line 248: This part of the code should not worry about specific text on the errors.

[The data-gatherer](https://github.com/jetstack/preflight/blob/c62dbc4b54642565ceff374d8cc9826980d703e6/pkg/datagatherer/k8s/dynamic.go#L138) should return an appropriate error type.

Also, in general, we should not be checking the text in error messages, as that is fragile. First I would check if there is an error type we can check for in the upstream k8s client. If not, the more robust solution would be to check if the CRD is defined or not. If that turns out to be too complicated, then checking the text of the error might be a good choice. But again, that should be a responsibility of the data-gatherer, not `agent/run.go`.

j-fuentes: Yes, that is what I meant. Ultimately, the decision of "failing hard vs retrying" will be made on `agent/run.go`. But I would not expect that code to compare the error text. Ideally, it would just check an error type.

pkg/agent/run.go
```diff
@@ -210,8 +211,20 @@ func gatherData(ctx context.Context, config Config) []*api.DataReading {
...
 			dgData, err := dg.Fetch()
 			if err != nil {
-				err = fmt.Errorf("%s: %v", k, err)
-				dgError = multierror.Append(dgError, err)
+				if _, ok := err.(*k8s.CRDNotFoundError); ok {
+					err := k8s.CRDNotFoundError{
+						Err:          err.Error(),
+						DataGatherer: k,
+					}
+					if StrictMode {
```
JammyL, line 219: @charlieegan3 @j-fuentes 

I'm not a fan of this part. I think the `getReadings()` function should not be responsible for handling the error edge cases. Ideally this would be handled by the backoff loop. But I haven't been able to find anything that could "ignore" a certain type of error in the [backoff package](https://pkg.go.dev/github.com/cenkalti/backoff/v4?tab=doc). 

Do you have any suggestions?

pkg/agent/run.go
```diff
@@ -199,16 +203,28 @@ func gatherData(ctx context.Context, config Config) []*api.DataReading {
...
 		var dgError *multierror.Error
 		for k, dg := range dataGatherers {
 			if completedDataGatherers[k] {
 				continue
 			}
 			dgData, err := dg.Fetch()
 			if err != nil {
-				err = fmt.Errorf("%s: %v", k, err)
-				dgError = multierror.Append(dgError, err)
+				if _, ok := err.(*k8s.CRDNotFoundError); ok {
```
charlieegan3, line 214: I think `_` here is the CRD error you want. e.g. this bit hopefully isn't needed.

```
err := k8s.CRDNotFoundError{
  Err:          err.Error(),
  DataGatherer: k,
}
```

pkg/agent/run.go
charlieegan3: I think that what you have looks ok - but as you say, it's a shame that we have know about the k8s error here in `Run`.

Perhaps datagathers can return an error or a permanent error. If they return a permanent error, then there will be no retries.

This way the dynamic dg can make a crd missing error a more generic 'PermanentError' signalling not to retry?

pkg/agent/run.go
JammyL: The purpose of this is to add the DataGatherer into the error. This could also be handled by giving `k` as an input to `dg.Fetch()` 
i.e. `func (g *DataGathererDynamic) Fetch(Datagatherer string) (interface{}, error) {`

Or perhaps adding the name of the data gatherer to the `DataGathererDynamic` structure is also a valid way to go about it

pkg/agent/run.go
JammyL: I don't think using the permanent error would work. This would run into the problem of cancelling the retries if even if there are other errors that require a retry.

j-fuentes: > This would run into the problem of cancelling the retries if even if there are other errors that require a retry.

I don't understand this. If a data gatherer returns a `PermanentError`, then this should not continue processing that data gatherer and continue to the next (or fail hard if running with strict mode).

pkg/datagatherer/local/local.go
```diff
@@ -46,3 +46,12 @@ func (g *DataGatherer) Fetch() (interface{}, error) {
 	}
 	return dataBytes, nil
 }
+
+// ConfigError is the error type for a misconfiguration. e.g. A missing CRD
```
j-fuentes, line 50: `local` is not the right package for this. I suggest moving this to `datagatherer`.

PR #197, opened 2020-09-15
Update k8s client

j-fuentes:
Update k8s dependencies.


pkg/datagatherer/k8s/dynamic_test.go
```diff
@@ -129,50 +132,50 @@ func TestDynamicGatherer_Fetch(t *testing.T) {
 		"Foos in different namespaces should be returned if no namespace field is set": {
 			gvr: schema.GroupVersionResource{Group: "foobar", Version: "v1", Resource: "foos"},
 			objects: []runtime.Object{
-				getObject("foobar/v1", "Foo", "testfoo", "testns"),
-				getObject("foobar/v1", "Foo", "testfoo", "nottestns"),
+				getObject("foobar/v1", "Foo", "testfoo", "testns1"),
```
j-fuentes, line 135: For some reason, the new version of the client returns the objects ordered alphabetically by namespace.

pkg/datagatherer/k8s/dynamic_test.go
```diff
@@ -67,18 +68,20 @@ func asUnstructuredList(items ...*unstructured.Unstructured) *unstructured.Unstr
 }
 
 func TestNewDataGathererWithClient(t *testing.T) {
+	ctx := context.TODO()
```
charlieegan3, line 71: I'm not sure why this is TODO rather than just an empty context? (`context.Context`)

My understanding was that TODO was for when the a higher context was to be supplied from higher up at a later date.

j-fuentes: you are right! I have changed to [context.Background()](https://golang.org/pkg/context/#Background).

PR #198, opened 2020-10-06
Update go-client and apimachinery to the same version (v0.19.2)

j-fuentes:
go-client wasn't pinned to an actual release.

Ideally, go-client and apimachinery should use the same version.

I found this was a problem while reusing the k8s dynamic datagatherer in another project.


charlieegan3:
We also downgraded `github.com/Azure/aks-engine`?

/lgtm 
/hold

j-fuentes:
> We also downgraded `github.com/Azure/aks-engine`?
> 
> /lgtm
> /hold

good point. I missed that.

It is kind of a mess with aks-engine because it does not use releases for the k8s libraries (see https://github.com/Azure/aks-engine/blob/v0.56.0/go.mod#L61).

I am going to try and find a combination that works and is more modern.

maelvls:
In the `go.mod` I can see the pinned version `v11.0.0+incompatible`:

```go
require (
	k8s.io/client-go v11.0.0+incompatible
)

replace k8s.io/client-go => k8s.io/client-go v0.19.2
```

Did you have to `replace` due to transitive dependencies?

maelvls:
Ah, I just tried removing the `replace` directive and I can see why this was added:

```
go: finding module for package k8s.io/api/auditregistration/v1alpha1
github.com/jetstack/preflight-platform/backend/internal/backend imports
	github.com/jetstack/preflight/pkg/datagatherer/k8s imports
	k8s.io/client-go/kubernetes/scheme imports
	k8s.io/api/auditregistration/v1alpha1: module k8s.io/api@latest found (v0.20.4), but does not contain package k8s.io/api/auditregistration/v1alpha1
```

PR #199, opened 2020-10-13
Send gather time and cluster ID in agent upload

charlieegan3:
This will allow agent data to be categorized and stored in preparation
for processing async on the backend.


charlieegan3:
It's probably time we did a release. Last one was in July and getting https://github.com/jetstack/preflight/pull/194 out would be good too.

/unhold

PR #200, opened 2020-11-02
Add Version Checker datagatherer

charlieegan3:
This creates a datagatherer based on version checker. The new datagatherer wraps the dynamic client and [version-checker](https://github.com/jetstack/version-checker). It returns a list of pods and the results of their image check from version checker.

Fixes https://github.com/jetstack/preflight-platform/issues/608


pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -0,0 +1,256 @@
...
+	versionCheckerOptions vcapi.Options
+}
+
+// PodResult wraps a pod and a version checker result for an image found in one
+// of the containers for that pod. Exported so the backend can destructure
+// json.
+type PodResult struct {
+	Pod    v1.Pod            `json:"pod"`
+	Result *vcchecker.Result `json:"result"`
+}
```
JoshVanL, line 217: Are we duplicating the pod for every container we use? Could we improve this by wrapping the results with every paired containes?

```yaml
result:
  pod: {}
  results:
    containerName: &Result{}
```

pkg/datagatherer/versionchecker/versionchecker_test.go
```diff
@@ -0,0 +1,357 @@
...
+		t.Errorf("GCR token does not match: got=%+v want=%+v", got, want)
+	}
+
+	if got, want := cfg.VersionCheckerClientOptions.Selfhosted["selfhosted-6"].Password, "pa55w0rd"; got != want {
+		t.Errorf("Selfhosted 6 password does not match: got=%+v want=%+v", got, want)
+	}
+
+	if got, want := cfg.VersionCheckerClientOptions.Selfhosted["selfhosted-7"].Bearer, "pa55w0rd"; got != want {
+		t.Errorf("Selfhosted 7 bearer does not match: got=%+v want=%+v", got, want)
+	}
```
JoshVanL, line 109: Maybe structure these tests into a table based:

```go
tests := map[string]struct {
  configTarget string
  expTarget string
} {
  "expect Kubeconfig path": {
  configTarget: cfg.Dynamic.KubeConfigPath,
  expTarget: "/home/someone/.kube/config",
},
.......
}

for name, test := range tests {
    	if test.configTarget != test.expTarget {
		t.Errorf("unexpected config target: got=%q; want=%q", test.configTarget, test.expTarget)
	}
  })
}
```

Bit easier to see each individual test case, and names the tests.

pkg/datagatherer/versionchecker/versionchecker_test.go
```diff
@@ -0,0 +1,357 @@
...
+		} else if r.URL.Path == "/v2/jetstack/example/manifests/v1.0.1" {
+			// this is a partial response, but it's all version checker needs
+			responseContent = []byte(`{
+			  "schemaVersion": 1,
+			  "name": "jetstack/example",
+			  "tag": "v1.0.1"
+			}`)
+		} else {
+			t.Fatalf("Unexpected URL was called: %s", r.URL.Path)
+		}
```
JoshVanL, line 351: Switch is a good candidate here

```go
switch r.URL.Path {
  case "/api/v1/pods":
...
}
```

pkg/datagatherer/versionchecker/versionchecker_test.go
```diff
@@ -0,0 +1,357 @@
...
+			file, err := os.Open("fixtures/tags.json")
+			if err != nil {
+				t.Fatalf("failed to open tags fixture: %s", err)
+			}
+			defer file.Close()
+
+			responseContent, err = ioutil.ReadAll(file)
+			if err != nil {
+				t.Fatalf("failed to read tags fixture: %s", err)
+			}
```
JoshVanL, line 334: ioutil.ReadFile

pkg/datagatherer/versionchecker/versionchecker_test.go
```diff
@@ -0,0 +1,357 @@
...
+      }
+    },
+    "result": {
+      "CurrentVersion": "v1.0.0",
+      "LatestVersion": "v1.0.1",
+      "IsLatest": false,
+      "ImageURL": "%s/jetstack/example"
+    }
+  }
+]`, parsedURL.Host, parsedURL.Host)
```
JoshVanL, line 233: Maybe push the json to a const at the bottom of the file so it doesn't interrupt the flow of the code?

pkg/datagatherer/versionchecker/versionchecker_test.go
```diff
@@ -0,0 +1,357 @@
...
+	if err := tmpfile.Close(); err != nil {
+		return nil, fmt.Errorf("failed to close tmp host file after writing")
+	}
+
+	return tmpfile, nil
+}
+
+// creates a kubeconfig file on disk with a reference to the local server
+// allowing requests to be mocked
+func createKubeConfigWithServer(server string) (*os.File, error) {
```
JoshVanL, line 260: Think it makes sense to return just the filename here since the file is not longer open, and we only really need the name anyway.

pkg/datagatherer/versionchecker/versionchecker_test.go
```diff
@@ -0,0 +1,357 @@
...
+    namespace: default
+    user: test
+  name: test
+current-context: test
+users:
+- name: test
+  user:
+    username: test
+    password: test`, server)
+	tmpfile, err := ioutil.TempFile("", "example")
```
JoshVanL, line 280: Perhaps we create a testing directory for this testing suite, and then remove the directory at the end of the test?

The creation of the dir, and deletion can be done in the test main, and then we don't have to remember cleaning up per test.

pkg/datagatherer/versionchecker/versionchecker_test.go
```diff
@@ -0,0 +1,357 @@
...
+			responseContent = []byte(`{
+			  "schemaVersion": 1,
+			  "name": "jetstack/example",
+			  "tag": "v1.0.1"
+			}`)
+		} else {
+			t.Fatalf("Unexpected URL was called: %s", r.URL.Path)
+		}
+
+		fmt.Fprint(w, string(responseContent))
```
JoshVanL, line 353: We can use `w.Write`, and don't need to put the content in a `string()`

JoshVanL (changes requested):
Made a couple comments, hope they make sense :smile: 

/assign @charlieegan3

pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -0,0 +1,255 @@
...
+	c.Dynamic.ExcludeNamespaces = aux.Dynamic.ExcludeNamespaces
+	c.Dynamic.IncludeNamespaces = aux.Dynamic.IncludeNamespaces
+	// gvr must be pods for the version checker dg
+	c.Dynamic.GroupVersionResource.Group = ""
+	c.Dynamic.GroupVersionResource.Version = "v1"
+	c.Dynamic.GroupVersionResource.Resource = "pods"
+
+	c.VersionCheckerClientOptions.Selfhosted = map[string]*selfhosted.Options{}
+	registryKindCounts := map[string]int{}
+	for i, v := range aux.Registries {
```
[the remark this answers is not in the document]
charlieegan3: I'm keen to leave it open to having multiple registries of the same kind in the future (as is possible with the self hosted configs).

I'd have done as you said if it were easier to update the agent configs, but when it's installed in many customer clusters I felt it was better to go with something that worked in both cases.

pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -0,0 +1,255 @@
...
+			}
+
+			opts := selfhosted.Options{
+				Username: data["username"],
+				Password: data["password"],
+				Bearer:   data["bearer"],
+				Host:     data["host"],
+			}
+
+			c.VersionCheckerClientOptions.Selfhosted[fmt.Sprintf("selfhosted-%d", i+1)] = &opts
```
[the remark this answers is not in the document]
charlieegan3: I'm not sure I follow. Likely related to https://github.com/jetstack/preflight/pull/200#discussion_r516796910

pkg/datagatherer/versionchecker/versionchecker_test.go
charlieegan3: We need to delete the file at the end of the test case.

pkg/datagatherer/versionchecker/versionchecker_test.go
charlieegan3: Fixed in 6b6ab818c1e4f74b3e7809c0b487eedeecb23bdc

pkg/datagatherer/versionchecker/versionchecker_test.go
charlieegan3: Fixed in 6b6ab818c1e4f74b3e7809c0b487eedeecb23bdc

pkg/datagatherer/versionchecker/versionchecker_test.go
charlieegan3: Fixed in 87384d2b9bee22da0a1c2ed3d5868fc1074eafe2

pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -0,0 +1,255 @@
...
+		file, err := os.Open(path)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load file for %s at %s: %s", k, path, err)
+		}
+		defer file.Close()
+
+		b, err := ioutil.ReadAll(file)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read file for %s at %s: %s", k, path, err)
+		}
```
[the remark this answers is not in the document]
charlieegan3: Fixed in 87384d2b9bee22da0a1c2ed3d5868fc1074eafe2

pkg/datagatherer/versionchecker/versionchecker.go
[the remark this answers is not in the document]
JoshVanL: I'm a bit dubious of this being a real use case, and would have thought a single sufficiently scoped identity for each cluster would be enough. I haven't come across a customer whose needed separate service accounts for the same registry provider, or indeed wasn't in a position to create a new one which could be used just for version-checker.

>  but when it's installed in many customer clusters I felt it was better to go with something that worked in both cases.

Would this not be a case of having different value files for different clusters, which is fairly standard anyway.

I don't plan on adding this functionality to version-checker unless there was a real call for it in an issue by multiple people. It is also a nightmare to get it to work programmatically- you would at worst, try each client until one worked, and at best make the user need to define exactly which images are for which client, even though they share the same host.

pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -0,0 +1,255 @@
...
+				return fmt.Errorf("failed to load params for registry %d/%d: %s", i+1, len(aux.Registries), err)
+			}
+
+			c.VersionCheckerClientOptions.ACR.Username = data["username"]
+			c.VersionCheckerClientOptions.ACR.Password = data["password"]
+			c.VersionCheckerClientOptions.ACR.RefreshToken = data["refresh_token"]
+		case "ecr":
+			data, err := loadKeysFromPaths([]string{"access_key_id", "secret_access_key", "session_token"}, v.Params)
+			if err != nil {
+				return fmt.Errorf("failed to load params for registry %d/%d: %s", i+1, len(aux.Registries), err)
```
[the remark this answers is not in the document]
charlieegan3: Fixed in 578be1c87edba21edff3d14c18aa900171f7c019

pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -0,0 +1,255 @@
...
+			// don't try to load unset secrets. version-checker will fail if
+			// config is missing
+			continue
+		}
+
+		file, err := os.Open(path)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load file for %s at %s: %s", k, path, err)
+		}
+		defer file.Close()
```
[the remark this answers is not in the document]
charlieegan3: Fixed in 87384d2

pkg/datagatherer/versionchecker/versionchecker.go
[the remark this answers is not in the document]
JoshVanL: This is basically for logging really, not too important..

In version-checker, selfhosted clients are defined as environment variables which are given a name:

```yaml
selfhosted:
  - name: REGISTRY
    host: http://registry:5000
    username:
    password:
    token:
```
This just means that this client is labelled "-registry" or some such in the logs.

pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -0,0 +1,255 @@
...
+			return fmt.Errorf("registry %d/%d was an unknown kind (%s)", i+1, len(aux.Registries), v.Kind)
+		}
+	}
+
+	for k, v := range registryKindCounts {
+		if v > 1 && k != "selfhosted" {
+			return fmt.Errorf("found %d registries of kind %s, only 1 is supported", v, k)
+		}
+	}
+
```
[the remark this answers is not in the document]
charlieegan3: See comment: https://github.com/jetstack/preflight/pull/200#discussion_r516796910

I see this as a temporary check. 

I added a comment to that effect in 455d035980a282cdd824b5d49152faece38eef9e

pkg/datagatherer/versionchecker/versionchecker.go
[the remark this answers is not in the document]
charlieegan3: I added a comment about this in 455d035980a282cdd824b5d49152faece38eef9e

pkg/datagatherer/versionchecker/versionchecker.go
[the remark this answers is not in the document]
charlieegan3: Resolving, will continue discussion here: https://github.com/jetstack/preflight/pull/200#discussion_r516770999

pkg/datagatherer/versionchecker/versionchecker_test.go
JoshVanL: Indeed, but it is safer to only return file.Name(), rather than the *os.File which has already been closed. Taking a minimal footprint too, we only need the name string, rather than the entire os.File struct.

pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -0,0 +1,255 @@
...
+	vcsearch "github.com/jetstack/version-checker/pkg/controller/search"
+	vcversion "github.com/jetstack/version-checker/pkg/version"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// Config is the configuration for a VersionChecker DataGatherer.
+type Config struct {
```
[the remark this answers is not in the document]
charlieegan3: I guess the main bit is here: https://github.com/jetstack/preflight/blob/master/pkg/agent/config.go#L65 with some custom unmarshalling.

Which reminded me to add our new dg to the list: 1d3231c78eec9e590eb82337e72577961b4cfd98

pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -0,0 +1,255 @@
...
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/jetstack/preflight/pkg/datagatherer"
+	"github.com/jetstack/preflight/pkg/datagatherer/k8s"
```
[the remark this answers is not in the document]
charlieegan3: Like this? e25a81c5fcece68b75763f26eb4932766e81fa2d

pkg/datagatherer/versionchecker/versionchecker_test.go
charlieegan3: There are only two files written during the test at the moment so this is acceptable to me. If we add more, we can come up with a better way to do it.

I made the prefixes more descriptive in 1508c19592d2847d97559c2dcd610515be6f3738.

pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -0,0 +1,255 @@
...
+		case "quay":
+			data, err := loadKeysFromPaths([]string{"token"}, v.Params)
+			if err != nil {
+				return fmt.Errorf("failed to load params for registry %d/%d: %s", i+1, len(aux.Registries), err)
+			}
+
+			c.VersionCheckerClientOptions.Quay.Token = data["token"]
+		case "selfhosted":
+			// currently, version checker only supports multiple selfhosted registries
+			data, err := loadKeysFromPaths([]string{"host", "username", "password", "bearer"}, v.Params)
```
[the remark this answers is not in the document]
charlieegan3: Fixed in 93c92a53e1446ccfae16ba974368fb8b81651098

pkg/datagatherer/versionchecker/versionchecker.go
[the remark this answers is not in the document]
charlieegan3: Ahh I see. I guess I could add name to the self hosted registry config. Another option would be to use the hostname.

pkg/datagatherer/versionchecker/versionchecker_test.go
charlieegan3: Ahh, my bad. I thought we needed that for os.Remove... 

Fixed in c354eb7a10592e5347082bbe9e13e49f4ca62255

pkg/datagatherer/versionchecker/versionchecker.go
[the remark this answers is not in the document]
JoshVanL: I mean change the grouping into [this convention](https://github.com/jetstack/cert-manager/blob/8127f0ad4228c58f37f0c0e4f5a2ef90c04f5082/pkg/controller/util.go#L19):

```go
import (
  "standard/library"
   
  "external"

  "project/internal"
)

File: pkg/datagatherer/versionchecker/versionchecker.go (hunk 9 above)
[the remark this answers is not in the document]
Comment (charlieegan3) [reply pkg/datagatherer/versionchecker/versionchecker.go:90]: I'm mostly doing this as we've been burned by it before in preflight. We used to only allow one datagather of each kind for example. Since then, I've generally preferred to have a list rather than a more rigid map for to make it more flexible with little overhead. I think the overhead is worth it to me when the alternative is updating agent config which slows us down a lot as we need to get users to update and have no means of automating it yet.

w.r.t not adding it to VC. Fair enough. My only comment on would be related to this:

> Would this not be a case of having different value files for different clusters, which is fairly standard anyway.

At the moment our model is one agent per cluster and that'd the route we guide users down too. However, everything else in preflight has been designed and built to support multiple clusters and environments if the user wanted to do that. At one point there was talk about running the agent in cloud run for example (talking to many clusters/envs).

Perhaps that's something we need to revisit, but it's an assumption I continue to make in preflight for the time being.

File: pkg/datagatherer/versionchecker/versionchecker.go (hunk 10 above)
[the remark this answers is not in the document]
Comment (JoshVanL) [reply pkg/datagatherer/versionchecker/versionchecker.go:148]: Yeah, all works, just makes debugging using the logs a little easier

File: pkg/datagatherer/versionchecker/versionchecker.go (hunk 1 above)
Comment (charlieegan3) [reply pkg/datagatherer/versionchecker/versionchecker.go]: I've added something similar in 22023da7da642e2e50a1668fce87517e56938a7b but it also tracks if the container is an init container. (I don't hate maps, honest :joy: )

File: pkg/datagatherer/versionchecker/versionchecker.go (hunk 16 above)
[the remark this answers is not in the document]
Comment (charlieegan3) [reply pkg/datagatherer/versionchecker/versionchecker.go]: Ahh, thanks. I have fixed in 25b1117bb4943941eb87eea85f86f57fb9a2975f

File: pkg/datagatherer/versionchecker/versionchecker.go (hunk 10 above)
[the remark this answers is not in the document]
Comment (charlieegan3) [reply pkg/datagatherer/versionchecker/versionchecker.go]: Sounds good, I've used the host in 42cd668291c5532a5e9f135105bf030cb3e1ec03

File: pkg/datagatherer/versionchecker/versionchecker_test.go (hunk 2 above)
Comment (charlieegan3) [reply pkg/datagatherer/versionchecker/versionchecker_test.go]: Agreed, fixed in f1551f8076f28d3666c295a4c16170ba14348fc9

File: pkg/datagatherer/versionchecker/versionchecker_test.go (hunk 5 above)
Comment (charlieegan3) [reply pkg/datagatherer/versionchecker/versionchecker_test.go]: Call me mad, but I like this :joy: 

I'm going to leave this comment open in case others agree.

File: pkg/datagatherer/versionchecker/versionchecker.go [hunk 18]
```diff
@@ -0,0 +1,310 @@
...
+			}
+
+			c.VersionCheckerClientOptions.GCR.Token = data["token"]
+		case "acr":
+			data, err := loadKeysFromPaths([]string{acrUsernameKey, acrPasswordKey, acrRefreshTokenKey}, v.Params)
+			if err != nil {
+				return fmt.Errorf("failed to load params for %s registry %d/%d: %s", v.Kind, i+1, len(aux.Registries), err)
+			}
+
+			c.VersionCheckerClientOptions.ACR.Username = data["username"]
```
tfadeyi, line 106: this is super minor but since we have the consts should we just use them in the keys?

tfadeyi:
It lgtm to me :+1:

pkg/datagatherer/versionchecker/versionchecker.go
charlieegan3: Oops, good spot, fixed in 75a168cce1ed172014521ff1f58c84e123483f91

j-fuentes (approved):

PR #202, opened 2020-11-10
Drop managed fields from objects

charlieegan3:
This adds weight to the agent uploads that we don't need.

Also includes changes from a go mod tidy.

### DIFF
--- a/pkg/datagatherer/k8s/dynamic.go
+++ b/pkg/datagatherer/k8s/dynamic.go
@@ -200,15 +200,21 @@ func redactList(list *unstructured.UnstructuredList) error {
 					secret.Object["data"] = map[string]interface{}{}
 				}
 
-				// Redact last-applied-configuration annotation if set
-				annotations, present := secret.Object["annotations"].(map[string]interface{})
-				if present {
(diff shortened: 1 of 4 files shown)


pkg/datagatherer/k8s/dynamic.go
```diff
@@ -211,12 +213,25 @@ func redactList(list *unstructured.UnstructuredList) error {
...
+			metadata, metadataPresent := object.Object["metadata"].(map[string]interface{})
+			if metadataPresent {
+				// Drop managed fields if set
+				if _, present := metadata["managedFields"]; present {
+					delete(metadata, "managedFields")
+				}
+			}
+
+			// update the metadata after an modifications
+			object.Object["metadata"] = metadata
```
tfadeyi, line 234: Is this needed?

charlieegan3: Huh, we don't. I just copied it from above. Removed in 2ba21c0092016487bce2a1bdba2525e60291cd5b

charlieegan3:
Closed in favor of https://github.com/jetstack/preflight/pull/203

PR #204, opened 2020-11-18
Implement more generic means of filtering uploaded objects

charlieegan3:
This can also be imported by the backend.

Also changes uploaded secret data to only contain a fixed allowlist of keys within the object.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,8 @@ require (
 	github.com/Azure/go-autorest/autorest v0.11.8 // indirect
 	github.com/Azure/go-autorest/autorest/to v0.4.0 // indirect
 	github.com/Azure/go-autorest/autorest/validation v0.3.0 // indirect
+	github.com/Jeffail/gabs v1.1.1
+	github.com/Jeffail/gabs/v2 v2.6.0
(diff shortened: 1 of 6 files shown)


j-fuentes:
It looks good to me.

I assume the part that you want to reuse is `Select` and `Redact`.

It would be great to reuse the list of fields to select and redact for secrets in the backend as well, but it is not a strong requirement.

/lgtm
/hold

charlieegan3:
> It would be great to reuse the list of fields to select and redact for secrets in the backend as well, but it is not a strong requirement.

Yeah, I did wonder if this would be an idea...

They could be package variables - how about like this: 0521da0fa4622b2d3a4129f6ccd1fa19c3c1e75f

j-fuentes:
> > It would be great to reuse the list of fields to select and redact for secrets in the backend as well, but it is not a strong requirement.
> 
> Yeah, I did wonder if this would be an idea...
> 
> They could be package variables - how about like this: [0521da0](https://github.com/jetstack/preflight/commit/0521da0fa4622b2d3a4129f6ccd1fa19c3c1e75f)

Looks good.

PR #205, opened 2020-11-18
Correct last-applied-configuration annotations key

charlieegan3:
This isn't essential since we're now selecting fields but it should
still be fixed.


charlieegan3:
Thanks

PR #207, opened 2020-11-19
Make more prominent that only secrets are redacted

j-fuentes:


charlieegan3:
/lgtm
/hold

One addition might be mentioning of managed fields. This is now removed from all resources but not really that interesting....

j-fuentes:
> /lgtm
> /hold
> 
> One addition might be mentioning of managed fields. This is now removed from all resources but not really that interesting....

We redact both `managedFields` and `last-applied-configuration` but I am intentionally avoiding being too specific in this doc, as it is so easy to forget to update it if we change the code that filters fields.

charlieegan3:
> intentionally avoiding being too specific
Makes sense

Issue #58, opened 2020-01-23
Add docs for AKS and EKS data gatherer


wwwil:
There is no documentation for the AKS and EKS data gatherers. This should be added so that users know how to set them up.

charlieegan3:
We have basic versions of these now, more work to happen in https://github.com/jetstack/preflight-private/issues/120

charlieegan3:
I am working on this in https://github.com/jetstack/preflight/issues/118

PR #210, opened 2020-11-24
Update version checker data gatherer

tfadeyi:
Updating version checker data gatherer to support multi-arch images.
This PR uses the packages from the https://github.com/jetstack/version-checker/tree/58-add-arch-and-os-checks
branch of version-checker.
With this change checks will be performed on the image for the correct architecture, so if node is amd64 the check is done on images with amd64. This won't fix the issues seen with fat manifests [#58](https://github.com/jetstack/version-checker/issues/58).
image

image

Signed-off-by: oluwole.fadeyi <oluwole.fadeyi@jetstack.io>

### DIFF
--- a/pkg/datagatherer/versionchecker/versionchecker.go
+++ b/pkg/datagatherer/versionchecker/versionchecker.go
@@ -10,11 +10,12 @@ import (
 	"time"
 
 	vcapi "github.com/jetstack/version-checker/pkg/api"
+	vcchecker "github.com/jetstack/version-checker/pkg/checker"
+	vcarchitecture "github.com/jetstack/version-checker/pkg/checker/architecture"
+	vcsearch "github.com/jetstack/version-checker/pkg/checker/search"
+	vcversion "github.com/jetstack/version-checker/pkg/checker/version"
 	vcclient "github.com/jetstack/version-checker/pkg/client"
 	vcselfhosted "github.com/jetstack/version-checker/pkg/client/selfhosted"
-	vcchecker "github.com/jetstack/version-checker/pkg/controller/checker"
-	vcsearch "github.com/jetstack/version-checker/pkg/controller/search"
-	vcversion "github.com/jetstack/version-checker/pkg/version"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -53,7 +54,9 @@ const (
 type Config struct {
 	// the version checker dg will also gather pods and so has the same options
 	// as the dynamic datagatherer
-	Dynamic                     k8s.ConfigDynamic
+	DynamicPod k8s.ConfigDynamic
+	// the nodes information is also gathered by the version checker datagatherer
+	DynamicNode                 k8s.ConfigDynamic
 	VersionCheckerClientOptions vcclient.Options
 	// Currently unused, but keeping to allow future config of VersionChecker
 	VersionCheckerCheckerOptions vcapi.Options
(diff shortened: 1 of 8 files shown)


pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -257,6 +277,26 @@ type containerResult struct {
...
+	if !ok {
+		return nil, fmt.Errorf("failed to parse nodes loaded from DataGatherer")
+	}
+	for _, v := range nodes.Items {
+		var node v1.Node
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(v.Object, &node)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse node from unstructured data")
+		}
+		if err = g.nodeArchitecture.Add(&node); err != nil {
```
charlieegan3, line 295: I might add a comment here explaining the impact this has on version checker. My guess is that this gives VC a means to lookup the arch for the node of a given pod to determine the correct latest version?

pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -53,7 +54,9 @@ const (
 type Config struct {
 	// the version checker dg will also gather pods and so has the same options
 	// as the dynamic datagatherer
-	Dynamic                     k8s.ConfigDynamic
+	Dynamic k8s.ConfigDynamic
```
charlieegan3, line 57: I think it makes sense to have 2, one for nodes, and one for pods. However, I'd rename the pods one to something like `DynamicPod`

pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -84,6 +87,13 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	c.Dynamic.GroupVersionResource.Group = ""
 	c.Dynamic.GroupVersionResource.Version = "v1"
 	c.Dynamic.GroupVersionResource.Resource = "pods"
+	// node dynamic dg
+	c.DynamicNode.KubeConfigPath = aux.Dynamic.KubeConfigPath
+	c.DynamicNode.ExcludeNamespaces = make([]string, 0)
```
charlieegan3, line 92: Might be a more terse way to write this?

```suggestion
	c.DynamicNode.ExcludeNamespaces = []string{}
```

charlieegan3:
/lgtm
/hold

Looks good, pending a few minor comments.

pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -257,7 +277,29 @@ type containerResult struct {
 
 // Fetch retrieves cluster information from GKE.
 func (g *DataGatherer) Fetch() (interface{}, error) {
-	rawPods, err := g.dynamicDg.Fetch()
+	// Get nodes information to update version-checker architecture structure
+	rawNodes, err := g.nodeDynamicDg.Fetch()
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch nodes, %s", err)
```
j-fuentes, line 283: I would recommend using `%v` when printing error rather than `%s`, so it prints all the content of the error, which can be useful especially if the error has been wrapped multiple times.

Not a hard requirement though.

pkg/datagatherer/versionchecker/versionchecker.go
```diff
@@ -257,7 +277,29 @@ type containerResult struct {
...
+	}
+	nodes, ok := rawNodes.(*unstructured.UnstructuredList)
+	if !ok {
+		return nil, fmt.Errorf("failed to parse nodes loaded from DataGatherer")
+	}
+	for _, v := range nodes.Items {
+		var node v1.Node
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(v.Object, &node)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse node from unstructured data")
```
j-fuentes, line 293: Print here the content of the error as it can be useful when debugging.

j-fuentes (approved):

j-fuentes:
Looking good. Consider my suggestions.


/lgtm
/hold

tfadeyi:
Is it fine if merge this PR? Or should we keep it on hold for version-checker's release?

j-fuentes:
> Is it fine if merge this PR? Or should we keep it on hold for version-checker's release?

Let's go ahead and merge this so we can beta-test this agent.

Once the changes are applied upstream in version-checker we can update the dependency.

PR #211, opened 2021-01-22
Create Istio analysis datagatherer

wwwil:
```release-note
Add an Istio data gatherer to run Istio analysis as part of Preflight checks.
```

This data gatherer uses several Kubernetes dynamic data gatherers to fetch all resource kinds used by Istio analysis. The list of kinds needed is currently built into the data gatherer. 

All gathered resources are then combined into a single YAML document which is fed into Istio's analyzer using a reader.

Using the existing Kubernetes dynamic data gatherers, rather than allowing Istio to gather resources itself, means that Namespace filtering and Secret redaction are applied. The Istio analysis code has no access to the cluster.


pkg/datagatherer/istio/istio.go
```diff
@@ -0,0 +1,140 @@
...
+	sa := local.NewSourceAnalyzer(istioSchema.MustGet(), analyzers.AllCombined(),
+		"", resource.Namespace(istioNamespace), nil, true, 30*time.Second)
+
+	return &DataGatherer{
+		sa:       sa,
+		dynamics: dynamics,
+	}, nil
+}
+
+// Fetch retrieves information from GKE.
```
charlieegan3, line 93: suggestion:

```suggestion
// Fetch retrieves Istio resources from the K8s API and runs istio analysis
```

charlieegan3:
> Using the existing Kubernetes dynamic data gatherers, rather than allowing Istio to gather resources itself, means that Namespace filtering and Secret redaction are applied.

This is good.

I think it makes sense to have the istio analysis in the agent as you have here. We don't have a way to run that on the backend and it would involved some special handling for istio resources.

I'm a little nervous about the hard-coded GVR list with istio being a relatively new project, are we worried about these changing soon? Might it be good to have way to override them in the agent's dg config?

At the moment, I see that only the results from istioctl analyse are sent? I wonder if it's worth also sending all the raw resources in order to allow us to write custom policy over these beyond what's enforced with istioctl analyse?

wwwil:
I've updated this today and think it's now ready for review! 🎉 

The tests for this data gatherer pass, though I seem to have introduced a test failure with the Kubernetes dynamic data gatherer. I think this may be related to a change with the version of the Kubernetes client which I had to make for the Istio stuff to work. I'm not sure how to handle this or what the best approach is for Go dependencies generally. If anyone could advise I'd appreciate that, or I can dig into it when I work on it again next Friday.

> I think it makes sense to have the istio analysis in the agent as you have here. We don't have a way to run that on the backend and it would involved some special handling for istio resources.

Cool! Yeah I thought a data gatherer seemed like the best place to run it as they're designed to be extensible and it means the backend can continue to just evaluate JSON with Rego.

> I'm a little nervous about the hard-coded GVR list with istio being a relatively new project, are we worried about these changing soon? Might it be good to have way to override them in the agent's dg config?

Yeah this isn't ideal. I have added the option to specify which resources are gathered in the data gatherer config so it's easier to change, and if none are specified then the built in list is used by default.

We may have to add new resources for future Istio releases, though in this case the version of `istioctl analyze` used would also need to be updated to include the new checks. This does add some maintenance burden to keep up with changes to Istio but it should be minimal, and is certainly easier than maintaining all the checks ourself in Rego.

> At the moment, I see that only the results from istioctl analyse are sent? I wonder if it's worth also sending all the raw resources in order to allow us to write custom policy over these beyond what's enforced with istioctl analyse?

Yeah I'd like to add some additional 'Jetstack opinionated' Isito checks in Rego, which would require raw resource data. We could also send the raw resources from this data gatherer but I'm not sure what format they should be in. Currently this returns:

```json
{
  "Messages": [...]
  "SkippedAnalyzers": [...]
  "ExecutedAnalyzers": [...]
}
```

So we could add them to this like:

```json
{
  "Messages": [...]
  "SkippedAnalyzers": [...]
  "ExecutedAnalyzers": [...]
  "pods": [...]
  "namespaces": [...]
  ...
}
```

Alternatively it may be easier to just use the Kubernetes dynamic data gatherer directly. This would fetch some resources multiple times, but the Rego would look more like other packages in terms of how it accessed the data. This approach would also make it clearer to users what data you were going get and only send relevant resources to the backend: The Isitio data gatherer would just give Istio analysis results. For the Istio package to perform other checks it would need to get data for relevant resources separately.

I'd like to get some input on this! I can see advantages and disadvantages to either approach...

pkg/datagatherer/istio/gvrs.go
```diff
@@ -0,0 +1,142 @@
+package isito
```
charlieegan3, line 1:
```suggestion
package istio
```

pkg/datagatherer/istio/istio.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	// If a list of resources is not set then use the data gatherer default list.
+	gvrs := c.Resources
+	if len(gvrs) == 0 {
+		gvrs = defaultGVRs
+	}
+
+	// Create a list of Kubernetes dynamic data gatherers to fetch all the required resources for Istio analysis.
+	var dynamics []datagatherer.DataGatherer
```
charlieegan3, line 78: Just a suggestion, i'd rather give this a more descriptive name.

```suggestion
	var dynamicDataGatherers []datagatherer.DataGatherer
```

pkg/datagatherer/istio/istio.go
```diff
@@ -0,0 +1,156 @@
...
+		dynamics = append(dynamics, dynamic)
+	}
+
+	// Create an Istio SourceAnalyzer.
+	sa := local.NewSourceAnalyzer(istioSchema.MustGet(), analyzers.AllCombined(),
+		"", resource.Namespace(istioNamespace), nil, true, 30*time.Second)
+
+	return &DataGatherer{
+		sa:       sa,
+		dynamics: dynamics,
```
charlieegan3, line 100:
```suggestion
		dynamicDataGatherers: dynamics,
```

pkg/datagatherer/istio/istio.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		dynamics = append(dynamics, dynamic)
+	}
+
+	// Create an Istio SourceAnalyzer.
+	sa := local.NewSourceAnalyzer(istioSchema.MustGet(), analyzers.AllCombined(),
+		"", resource.Namespace(istioNamespace), nil, true, 30*time.Second)
+
+	return &DataGatherer{
+		sa:       sa,
```
charlieegan3, line 99:
```suggestion
		sourceAnalyzer:       sa,
```

pkg/datagatherer/istio/istio.go
```diff
@@ -0,0 +1,156 @@
...
+	if len(errors) > 0 {
+		return fmt.Errorf(strings.Join(errors, ", "))
+	}
+
+	return nil
+}
+
+// DataGatherer is a DataGatherer for Istio.
+type DataGatherer struct {
+	sa       *local.SourceAnalyzer
```
charlieegan3, line 55: See comments below about more descriptive field names.

pkg/datagatherer/istio/istio.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// Convert the slice of Unstructured resources into a YAML document string.
+	allResourcesYAMLString := ""
+	for _, resource := range allResources {
+		// Marshall the unstructured resource to a YAML string.
+		resourceYAML, err := yaml.Marshal(resource.Object)
+		if err != nil {
+			return nil, err
+		}
+		allResourcesYAMLString = string(resourceYAML) + "---\n" + allResourcesYAMLString
```
charlieegan3, line 138: Might be easier to append to a slice and then strings.Join(slice, "---\n") at the end.

pkg/datagatherer/istio/istio.go
```diff
@@ -0,0 +1,156 @@
...
+	for _, gvr := range gvrs {
+		configDynamic := k8s.ConfigDynamic{
+			KubeConfigPath:       c.KubeConfigPath,
+			GroupVersionResource: gvr,
+			ExcludeNamespaces:    c.ExcludeNamespaces,
+			IncludeNamespaces:    c.IncludeNamespaces,
+		}
+		dynamic, err := configDynamic.NewDataGatherer(ctx)
+		if err != nil {
+			// TODO: Might be better to collect errors here rather than just give up straight away.
```
charlieegan3, line 88: This would be nice, but not essential.

charlieegan3:
Hey Wil, thanks for your continued work on this.

> Alternatively it may be easier to just use the Kubernetes dynamic data gatherer directly.

I think that this is the way to go (so no work needed here). Hopefully in the output from istioctl analyse there's information which an uniquely ID the resource in question? e.g. a UID, selflink or even name/namespace (selflink would be preferred since it's our primary reference for indexing other k8s resources)

pkg/datagatherer/istio/istio_test.go
```diff
@@ -0,0 +1,191 @@
...
+	"net/url"
+	"os"
+	"testing"
+
+	"gopkg.in/yaml.v2"
+)
+
+const tempFilePrefix = "preflight-test-istio-datagatherer"
+
+//// TestFetchRealCluster is not used, but has been left here for development use to allow testing against a real cluster.
```
charlieegan3, line 19: I think I'd rather this was stored as a personal snippet since we don't have it for our other datagatherers.

pkg/datagatherer/k8s/dynamic_test.go
```diff
@@ -130,98 +130,103 @@ func TestDynamicGatherer_Fetch(t *testing.T) {
...
-			expected: asUnstructuredList(
-				getObject("apps/v1", "Deployment", "foo1", "testns", false),
-				getObject("apps/v1", "Deployment", "foo2", "testns", false),
-			),
-		},
-		// Note that we can't test use of fieldSelector to exclude namespaces
-		// here as the as the fake client does not implement it.
-		// See go/pkg/mod/k8s.io/client-go@v0.17.0/dynamic/fake/simple.go:291
-		// TODO: Add a custom reactor to allow testing of fieldSelector.
+		// TODO restore these test cases by adding to the new list of gvr to list type mappings, all mapp to UnstructuredList
```
charlieegan3, line 133: @wwwil I think I've worked out what's needed to fix this if we do go this route, see below.

pkg/datagatherer/k8s/dynamic_test.go
```diff
@@ -130,98 +130,103 @@ func TestDynamicGatherer_Fetch(t *testing.T) {
...
+		//	// See go/pkg/mod/k8s.io/client-go@v0.17.0/dynamic/fake/simple.go:291
+		//	// TODO: Add a custom reactor to allow testing of fieldSelector.
 	}
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
-			cl := fake.NewSimpleDynamicClient(emptyScheme, test.objects...)
+			gvr := schema.GroupVersionResource{Group: "foobar", Version: "v1", Resource: "foos"}
+			gvrToListKind := map[schema.GroupVersionResource]string{
+				gvr: "UnstructuredList",
+			}
```
charlieegan3, line 228: We need map all gvrs to some common list type. UnstructuredList seemed as good as any, but it's not an API kind so might be better to find another one.

This is just to make sure we're appeasing the fake client for the tests, so I'm not so worried about how we do it.

pkg/datagatherer/k8s/dynamic_test.go
```diff
@@ -237,11 +242,25 @@ func TestDynamicGatherer_Fetch(t *testing.T) {
...
-				gotJSON, _ := json.MarshalIndent(res, "", "  ")
-				t.Fatalf("unexpected JSON: \ngot \n%s\nwant\n%s", string(gotJSON), expectedJSON)
+
+			if test.expected != nil {
+				list, ok := res.(*unstructured.UnstructuredList)
+				if !ok {
+					t.Errorf("expected result be an UnstructuredList but wasn't")
+				}
+
+				// TODO this is pointless, we can make the expected value an unstructured.UnstructuredList and avoid this check
```
charlieegan3, line 252: Might be worth just using the list as the expected value rather than asserting the type from an interace{} we set up in this very test.

pkg/datagatherer/istio/gvrs.go
wwwil: Oops haha 🤦

pkg/datagatherer/istio/istio.go
wwwil: I've removed this as I think it's unnecessary. The only config that differs between dynamic data gatherers is the `gvr`, which isn't checked at this stage anyway.

pkg/datagatherer/istio/istio.go
```diff
@@ -0,0 +1,156 @@
...
+	// KubeConfigPath is the path to the kubeconfig file. If empty, will assume it runs in-cluster.
+	KubeConfigPath string `yaml:"kubeconfig"`
+	// IstioNamespace is the Namespace for Istio components, this is usually `istio-system`.
+	IstioNamespace string `yaml:"istio-namespace"`
+	// ExcludeNamespaces is a list of namespaces to exclude.
+	ExcludeNamespaces []string `yaml:"exclude-namespaces"`
+	// IncludeNamespaces is a list of namespaces to include.
+	IncludeNamespaces []string `yaml:"include-namespaces"`
+	// Resources is a list of GroupVersionResources to collect for Istio analysis.
+	Resources []schema.GroupVersionResource
```
charlieegan3, line 35: Does this also need a yaml struct tag to support loading of a custom list?

charlieegan3: (if this isn't trivially easy, then I'm fine to leave that as a TODO for now)

pkg/datagatherer/istio/istio_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+const tempFilePrefix = "preflight-test-istio-datagatherer"
+
+// TestFetch runs a full test of the Istio data gatherer; running a fake Kubernetes API, using dynamic data gatherers to
+// fetch resources, running fetched resources through Istio analysis, and checking that the analysis messaged generated
+// are what was expected.
+func TestFetch(t *testing.T) {
+	// Local server to handle requests made by Kubernetes dynamic data gatherer.
```
charlieegan3, line 23: As I see this, the reason we need a localServer for this rather than using the fake client is that we have no way to inject the fake client to these 'level 2' datagathers which use the dynamic datagatherer as a dependency (version checker being the other datagatherer of the same level).

I'm happy with using a local server and feel that it's what I might have reached for when testing the dynamic dg originally if I worked on it.

However, I think that this is worth explaining in a comment as to why it's different from the tests for the dynamic dg.

charlieegan3: I made a comment in the version checker dg related to this here: #218

pkg/datagatherer/istio/istio_test.go
```diff
@@ -0,0 +1,166 @@
...
+			t.Fatalf("Unexpected URL was called: %s", r.URL.Path)
+		}
+
+		w.Write(responseContent)
+	}))
+
+	return localServer
+}
+
+var testNamespaces = `
```
charlieegan3, line 148: optional: this is short enough that it can be inside the createlocaltestserver case statement.

charlieegan3:
/lgtm  
/hold

hold to allow you to make any further changes. But this is ready to go in as I see it really. Thanks! and sorry for the delay in reviewing this.

pkg/datagatherer/istio/istio.go
wwwil: I've added a `YAML` tag for `Resources`. The `schema.GroupVersionResource` struct doesn't have YAML tags but I don't think they are required as its fields only have simple one word names.

This is also covered by the test as it unmarshals a config string that specifies resources.

pkg/datagatherer/istio/istio_test.go
wwwil: I've added an explanation for this 👍

pkg/datagatherer/istio/istio_test.go
wwwil: I'm inclined to leave it out, I think the function is easier to read without this string in and it will require less change if the testing is expanded in future.

wwwil:
@charlieegan3 I think this is good to merge now if you feel all the comments have been resolved 👍

pkg/datagatherer/istio/istio_test.go
```diff
@@ -0,0 +1,168 @@
...
+kubeconfig: %s
+resources:
+- group:    ""
+  version:  "v1"
+  resource: "namespaces"
+`
+
+// createKubeConfigWithServer creates a kubeconfig file on disk with a reference to the local server.
+func createKubeConfigWithServer(server string) (string, error) {
+	content := fmt.Sprintf(kuebConfigString, server)
```
charlieegan3, line 93:
```suggestion
	content := fmt.Sprintf(kubeConfigString, server)
```

pkg/datagatherer/istio/istio_test.go
```diff
@@ -0,0 +1,168 @@
...
+		return "", fmt.Errorf("failed to write to tmp kubeconfig file")
+	}
+	if err := tempFile.Close(); err != nil {
+		return "", fmt.Errorf("failed to close tmp kubeconfig file after writing")
+	}
+
+	return tempFile.Name(), nil
+}
+
+var kuebConfigString = `
```
charlieegan3, line 109:
```suggestion
var kubeConfigString = `
```

charlieegan3:
Thanks @wwwil! :tada:   
  
As discussed, I'm going to merge this to save it being left behind by any other work. Feel free to open follow on PRs for changes while you work on the policy side of this.

/lgtm

PR #212, opened 2021-02-02
Load the period from config file

charlieegan3:
This will allow users to update the period on which data is sent by only
applying a new config.

### DIFF
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -41,7 +41,7 @@ func init() {
 		&agent.Period,
 		"period",
 		"p",
-		time.Hour,
+		0,
 		"Time between scans (given as XhYmZs).",
 	)
 	agentCmd.PersistentFlags().StringVarP(
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -42,7 +42,7 @@ func init() {
 		"period",
 		"p",
 		0,
-		"Time between scans (given as XhYmZs).",
+		"Override time between scans in the configuration file (given as XhYmZs).",
 	)
 	agentCmd.PersistentFlags().StringVarP(
 		&agent.CredentialsPath,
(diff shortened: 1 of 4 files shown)


pkg/agent/run.go
```diff
@@ -84,6 +84,12 @@ func getConfiguration(ctx context.Context) (Config, *client.PreflightClient) {
 		}
 	}
 
+	// if period is set in the config, then use that
```
tfadeyi, line 87: Looks good:+1: But could we do the assignment to Period outside `getConfiguration`?
Does the config has priority over the cli args?

charlieegan3: I moved it to the main loop in fe707383d090dbe264fcafd4d2d541e790bd8e27. I think this is what we want. Other option is to set the default for the period to 0 and fail if it's not set, then they set it in the flag, or in config, where the flag takes priority.

Actually, I like that better.

charlieegan3: 18e1ebcb3ef8a766a57637fb70d1a637bdc5f293

tfadeyi (approved):

dgellow (approved):
LGTM

cmd/agent.go
```diff
@@ -41,7 +41,7 @@ func init() {
 		&agent.Period,
 		"period",
 		"p",
-		time.Hour,
+		0,
 		"Time between scans (given as XhYmZs).",
```
j-fuentes, line 45:
```suggestion
		"Override time between scans in configuration file (given as XhYmZs).",
```

j-fuentes (approved):

cmd/agent.go
charlieegan3: Fixed in 0b00c186f8cdc4753036be85be1993de6c586e8f

Issue #137, opened 2020-03-10
As a user I'd like the installation of the agent to be simpler


wwwil:
Currently there are instructions on how to deploy the Preflight agent to a cluster in the `README.md`. These include inline Kubernetes manifests which users can copy, edit, and deploy.

Ideally the deployment process should be even easier and more automated. This should improve the user experience, make it less likely that mistakes are introduced while copying manifests, and make casual users more likely to deploy the agent and try Preflight.

The only part of the deployment that user must edit is the agent configuration. Most of this configuration can have sensible defaults, except the user token. This means it should all be achievable with a few commands, which user can copy from the `README.md` and run. We probably can't quite have 'one click' installation but can reduce the required steps to a few commands for a minimal deployment.

**Ideas**

- The most minimal option would be for users to just **curl** a file with several Kubernetes manifests, edit the config or use `sed` to inject a token value from an environment variable, and `kubectl apply` the file.

- A tidier version of this would be to use **Kustomize** for deployment. Users could fetch a default config file and edit as required. Kustomize Secret/ConfigMap generation would then get this into the cluster, inject values using parches, and deploy the other default manifests. Users can make their own overlays to modify the rest of the deployment.
 
- Another option is a **Helm** chart. This seems overkill but is how a lot of people expect to install applications into their cluster so is worth considering.

- Manifests could be **generated by the backend**, so when a user signs up and a token is generated a set of manifests is also generated which they can just download and apply, or apply directly from an endpoint we provide.

**Acceptance Criteria:**

Acceptance Criteria is a set of test scenarios specific to this particular issue. It should capture functional and non-functional requirements.

- [ ] There is a minimal deployment process
- [ ] The deployment process has automated testing

**Risks:**
- Poor deployment automation could make the process more painful for users
  - This can be mitigated by ensuring the deployment process is tested
- Obfuscating the deployment manifests could put users off as they're not sure what they're putting in their cluster
  - This can be mitigated by ensuring that manifests are still visible in the repo and are well documented
- Overcomplicating the deployment process can make it harder for developers to maintain
  - This requires us to agree a balance of automation and maintainability. 

**Assumptions:**
- Users want easier deployment
- Users might make mistakes copying manifests
- Casual users will be put off by having to copy and edit multiple YAML files.

**Dependencies:** 
- This will introduce a dependency whatever process we use to test the deployment automation
  - For example we may introduce kind as a dependency for testing
- This may also introduce a deployment tool as a dependency
[...]

PR #213, opened 2021-02-09
A Preflight Agent Helm Chart for use by the Jetstack Secure package on Google Marketplace

wallrj:
We want to install the Preflight Agent as part of the [jetstack-secure package on Google Cloud Marketplace](https://github.com/jetstack/jsp-gcm/).
For this, we need to create a Helm chart for the preflight agent,
so I thought I may as well put the chart here.

I used the `agent_installation.yaml`  from https://platform.jetstack.io/agent_installation.yaml as described in the installation instructions in the "Machine Identity / Cert Inventory" section of that site.

The installation instructions in the README of this repo seem directed at the alternative Preflight configuration of the agent.

I have only looked at the certinventory version of the agent_installation.yaml. 
I have not looked at https://platform.jetstack.io/preflight_agent_installation.yaml

I'm not sure how to reconcile those two different configurations.
Perhaps I could rename this chart to "preflight-certinventory" or something.

Anyway, please take a look, and let me know what you think.
If you prefer not to include this chart in the repo, I can always add it to https://github.com/jetstack/jsp-gcm/ instead.

xref: https://github.com/jetstack/preflight/issues/137 https://github.com/jetstack/jsp-gcm/issues/8

```release-note
A Helm chart for easier / customizable installation of the agent
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,44 @@ taking place in the SaaS backend.
 The following instructions walk through the installation of the Preflight agent
 to gather data about cluster pods and send them to the backend for analysis.
 
+### Recommended Installation: Using Jetstack Secure
+
+The preferred way to install the Preflight agent is to sign in to https://platform.jetstack.io/
+and follow the instructions there.
+For example, to configure the agent for "Machine Identity / Cert Inventory":
+
+1. Visit https://platform.jetstack.io/
+2. Click the "Machine Identity" button, in the tool bar on the left
+3. Click "ADD CLUSTER"
+4. Follow the instructions
+5. Click "COPY COMMAND TO CLIPBOARD" to copy the credentials and configuration command to the clipboard
+
+
+### Helm Installation
+
+You can install the Preflight agent using Helm, as follows, but this is not recommended:
+
+```
+helm install --create-namespace --namespace jetstack-secure jetstack-secure deploy/charts/preflight
+```
+
+Next you will need to register the agent.
+For example, to configure the agent for "Machine Identity / Cert Inventory", please follow these steps:
+
+1. Visit https://platform.jetstack.io/
+2. Click the "Machine Identity" button, in the tool bar on the left
+3. Click "ADD CLUSTER"
+4. Follow the instructions
+5. Click "COPY COMMAND TO CLIPBOARD" to copy the credentials and configuration command to the clipboard
+6. Paste, inspect and then execute the command in your terminal
+
+
+### Manual Installation
+
+You can install the agent from scratch, by manually creating deployment manifests 
+and then using `kubectl apply` to deploy the agent to the cluster,
+but this is not recommended.
+
 **To complete the secret manifest below, you will need to have a Preflight
 token.**
 
--- a/deploy/charts/preflight/templates/NOTES.txt
+++ b/deploy/charts/preflight/templates/NOTES.txt
@@ -1,0 +1,10 @@
+Congratulations! The Jetstack Secure Agent is now installed.
+
+To register the agent, please follow these steps:
+
+1. Visit https://platform.jetstack.io/
+2. Click the "Machine Identity" button, in the tool bar on the left
+3. Click "ADD CLUSTER"
+4. Follow the instructions
+5. Click "COPY COMMAND TO CLIPBOARD" to copy the credentials and configuration command to the clipboard
+6. Paste, inspect and then execute the command in your terminal
--- a/deploy/charts/preflight/templates/agent_installation.yaml
+++ b/deploy/charts/preflight/templates/agent_installation.yaml
@@ -172,6 +172,10 @@ spec:
       {{- include "preflight.selectorLabels" . | nindent 6 }}
   template:
     metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       labels:
         {{- include "preflight.selectorLabels" . | nindent 8 }}
     spec:
--- a/deploy/charts/preflight/templates/agent_installation.yaml
+++ b/deploy/charts/preflight/templates/agent_installation.yaml
@@ -158,7 +158,7 @@ spec:
           secretName: agent-credentials
       containers:
       - name: agent
-        image: quay.io/jetstack/preflight:v0.1.27
+        image: "quay.io/jetstack/preflight:v{{ .Chart.AppVersion }}"
         args:
         - "agent"
         - "-c"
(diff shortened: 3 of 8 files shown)


wallrj:
Hey @maelvls please take a look at this too, if you have a moment, and tell me if I've done anything stupid and whether you think this is going to be usable with Google Cloud Marketplace deployer.

README.md
```diff
@@ -87,6 +87,44 @@ taking place in the SaaS backend.
...
+
+Next you will need to register the agent.
+For example, to configure the agent for "Machine Identity / Cert Inventory", please follow these steps:
+
+1. Visit https://platform.jetstack.io/
+2. Click the "Machine Identity" button, in the tool bar on the left
+3. Click "ADD CLUSTER"
+4. Follow the instructions
+5. Click "COPY COMMAND TO CLIPBOARD" to copy the credentials and configuration command to the clipboard
+6. Paste, inspect and then execute the command in your terminal
```
maelvls, line 119: Will the instructions there have the right namespace i.e. `jetstack-secure` in the above example? Or does the user have to change the namespace in the copy-pasted configuration command?

deploy/charts/preflight/templates/NOTES.txt
```diff
@@ -0,0 +1,10 @@
+Congratulations! The Jetstack Secure Agent is now installed.
+
+To register the agent, please follow these steps:
+
+1. Visit https://platform.jetstack.io/
+2. Click the "Machine Identity" button, in the tool bar on the left
+3. Click "ADD CLUSTER"
+4. Follow the instructions
+5. Click "COPY COMMAND TO CLIPBOARD" to copy the credentials and configuration command to the clipboard
+6. Paste, inspect and then execute the command in your terminal
```
maelvls, line 10: Wow, very nice! I didn't even thing about adding a `NOTES.txt` in the google-cas-issuer Helm chart... I'll have to do that too 😅

deploy/charts/preflight/templates/agent_installation.yaml
```diff
@@ -0,0 +1,214 @@
...
+  labels:
+    {{- include "preflight.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: view
+subjects:
+- kind: ServiceAccount
+  name: agent
+  namespace: {{ .Release.Namespace }}
```
maelvls, line 159: Since the Marketplace release does not allow us to create the RBAC roles, rolebindings and service accounts in the Helm chart, could we have a flag here to disable the creation of service accounts and RBAC roles and rolebindings?

e.g. like [in cert-manager](https://github.com/jetstack/cert-manager/blob/35febb171706826f27d71af466c624c25733c135/deploy/charts/cert-manager/templates/rbac.yaml#L1):

```yaml
{{- if .Values.global.rbac.create -}}

{{ end }}
```

deploy/charts/preflight/templates/agent_installation.yaml
```diff
@@ -0,0 +1,214 @@
...
+      - name: agent
+        image: "{{ .Values.image.repository }}:v{{ .Chart.AppVersion }}"
+        args:
+        - "agent"
+        - "-c"
+        - "/etc/jetstack-secure/agent/config/config.yaml"
+        - "-k"
+        - "/etc/jetstack-secure/agent/credentials/credentials.json"
+        - "-p"
+        - "0h1m0s"
```
maelvls, line 200: What is `-p`? Just wondering 😅

deploy/charts/preflight/templates/agent_installation.yaml
```diff
@@ -0,0 +1,214 @@
...
+        {{- include "preflight.selectorLabels" . | nindent 8 }}
+    spec:
+      serviceAccountName: agent
+      volumes:
+      - name: config
+        configMap:
+          name: agent-config
+      - name: credentials
+        secret:
+          secretName: agent-credentials
```
maelvls, line 189: Imagining that the user goes to the Marketplace and clicks on "Deploy", what will happen with this missing `agent-credentials` secret?

My guess is that the replicaset will stay in failure mode ("the volume cannot be mounted due to the secret not found")

Should we do nothing about that?

deploy/charts/preflight/templates/agent_installation.yaml
```diff
@@ -0,0 +1,214 @@
...
+  template:
+    metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "preflight.selectorLabels" . | nindent 8 }}
+    spec:
+      serviceAccountName: agent
```
maelvls, line 182: I think the service account name needs to be configurable so that we can swap it out in the Marketplace release. I guess we can take the example [from cert-manager](https://github.com/jetstack/cert-manager/blob/master/deploy/charts/cert-manager/templates/deployment.yaml#L53):

```
      serviceAccountName: {{ index .Values "cert-manager" "serviceAccountName" }}
```

Note the `index .Values`: this is due to the fact that Helm does not like JSON fields that contain a dash, so instead of doing

```
{{ .Values.cert-manager.serviceAccountName }}
```

we instead have to do:

```
{{ index .Values "cert-manager" "serviceAccountName" }}
```

maelvls (commented):
Looks very nice!!! I left a couple of comments 👍

wallrj:
Hey @maelvls and @charlieegan3 thanks for your feedback. 
As discussed on Slack, I'm going to move this chart to jsp-gcm repo for now until we figure out how the GCM integration will work.

deploy/charts/preflight/templates/agent_installation.yaml
wallrj: There's a `serviceAccountName` defined in the _helpers.tpl file, so I used that. It's derived from the values.yaml value or the release name.

deploy/charts/preflight/templates/agent_installation.yaml
wallrj: Yes, it hangs, waiting to mount the Secret. Let's discuss how to deal with that in a followup PR in jsp-gcm

deploy/charts/preflight/templates/agent_installation.yaml
wallrj: :shrug: I just copied it from the downloaded agent_installation file. Perhaps it's the frequency of cluster scanning?

README.md
wallrj: The instructions work with this namespace.

PR #218, opened 2021-02-19
Explain use of localServer in version checker dg

charlieegan3:
We are using this since we don't support the injecting of a fake client.
This is a difference with the dynamic dg I feel is worth highlighting.


tfadeyi (approved):

charlieegan3:
Thanks Olu.

PR #223, opened 2021-04-12
Move to Github actions and add support for arm

j-fuentes:


j-fuentes:
TODO: configure registry credentials so we can push the final image in `master` and tags.

j-fuentes:
This is the test workflow running for this PR: https://github.com/j-fuentes/preflight/actions/runs/744443032

j-fuentes:
I think this is ready now.

We just need to disable prow (jetstack/testing) for this repo once we validate this works.

I think we could just disable the CI stuff and keep `tide` and the dco check.

j-fuentes:
Actually, in 851a733 I make prow CI innocuous. We will remove it altogether after this, in another PR. Once we disable it in jetstack/testing as well.

.github/workflows/release-master.yml
```diff
@@ -0,0 +1,55 @@
...
+    # Setting up dind service container
+    services:
+      docker:
+        image: docker:20.10.5-dind
+        env:
+          DOCKER_DRIVER: overlay
+          DOCKER_HOST: tcp://localhost:2375
+    steps:
+    - name: Install Tools
+      run: apk add --update make git jq rsync curl
```
charlieegan3, line 42: I think that some of these are present already?

.github/workflows/release-tag.yml
```diff
@@ -0,0 +1,55 @@
+# if changing this name, also update promotion.yaml
+name: release-tag
+
+on:
+  push:
+    tags:
+      - v*
```
charlieegan3, line 7: Would this be better do do on the publication of a release?

.github/workflows/release-tag.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+    - name: Login to registry
+      uses: docker/login-action@v1
+      with:
+        registry: quay.io
+        username: ${{ secrets.QUAY_USER }}
+        password: ${{ secrets.QUAY_PASSWORD }}
+    - name: Build images
+      run: make push-docker-image
```
charlieegan3, line 55: Shame that all but the last step is duplicated here. I wonder if we could have another task which pulls the image for the sha of the tag. Then tags that as the release, and pushes to quay?

.github/workflows/test.yml
```diff
@@ -0,0 +1,52 @@
+# if changing this name, also update promotion.yaml
+name: test
+
+on:
+  push:
+    branches-ignore:
+      - master
```
charlieegan3, line 7: hmm, can't we use `if: github.ref == 'refs/heads/master'` -> make push canary in this file instead to save duplication?

charlieegan3 (approved):

.github/workflows/release-master.yml
j-fuentes:
```
$ docker run -ti docker:20.10.5 sh                                                   
/ # make
sh: make: not found
/ # git
sh: git: not found
/ # jq
sh: jq: not found
/ # rsync
sh: rsync: not found
/ # curl
sh: curl: not found
```

Apparently they are not.

.github/workflows/test.yml
j-fuentes: But I would like to avoid loading the secrets in a workflow that runs on every branch.

I am not sure what the implementation is, not sure if we can prevent loading the secrets with `if`.

.github/workflows/release-tag.yml
j-fuentes: I have tried this:

- pull this image which is multi-platform: 
image
```
$ docker pull josefuentes/kubectl:v1.20.0
v1.20.0: Pulling from josefuentes/kubectl
75646c2fb410: Already exists 
97150b587ac2: Pull complete 
5a321e5a7e17: Pull complete 
29f06f9cf893: Pull complete 
Digest: sha256:2daebe4eae191d4ebc2c58a600c988d9f43571d1a972cae839a806d47c31a1ad
Status: Downloaded newer image for josefuentes/kubectl:v1.20.0
docker.io/josefuentes/kubectl:v1.20.0
```

- Tag it with another tag:

```
$ docker tag josefuentes/kubectl:v1.20.0 josefuentes/kubectl:v1.20.0-copy

```


- Push the new tag:
```
$ docker push josefuentes/kubectl:v1.20.0-copy                           
The push refers to repository [docker.io/josefuentes/kubectl]
c317b8b2eb88: Layer already exists 
759b01e303b5: Layer already exists 
cf32df1a228a: Layer already exists 
346fddbbb0ff: Layer already exists 
v1.20.0-copy: digest: sha256:da9fa76c7a3761d63c46436e2a1ab44f2880518f4531dc1f5b64d99cc0a2209f size: 1160
```



As you can see, the resulting image is not multiplatform:

image



If we want to do that, we need to find some `buildx` magic. I don't see a clean simple solution at the moment.

The best I can think of implies using another Dockerfile to build a new image just with `FROM <image>:<commit>`.

But I don't think we want that, The sha of the resulting image will be different than the original one anyway.

j-fuentes: Nevermind! I have found a way!

.github/workflows/release-master.yml
charlieegan3: Ahh, my bad. They are in the github actions ubuntu image. Not the docker image...

.github/workflows/release-tag.yml
charlieegan3: Fair enough.

.github/workflows/release-tag.yml
j-fuentes: this should do the job: d836c1f

I have tested the command locally and ended up with this:

```
docker buildx imagetools create josefuentes/kubectl:v1.20.0 --tag josefuentes/kubectl:v1.20.0-copy2       
docker.io/josefuentes/kubectl:v1.20.0-copy2
```


image

.github/workflows/test.yml
charlieegan3: Is there a way you can stop the secrets being loaded in a non-master branch anyway?

.github/workflows/release-tag.yml
charlieegan3: Looks nice!

.github/workflows/test.yml
charlieegan3: Ahh I see, you can use environments...

PR #227, opened 2021-04-27
Use moby/buildkit:master when setting up buildx

davidsbond:
Suggested fix when hitting the 401 authorized error that we are when trying to push new images:

https://github.com/docker/buildx/issues/327


j-fuentes (approved):
Thanks, David!

PR #230, opened 2021-04-27
New CI workflow

davidsbond:
* Moves compilation of binaries into hack/build.sh. In this file, new architectures can be compiled by adding a new
function to the script. There are instructions in comments
* Removes the docker-in-docker build. Instead, the CI now installs go 1.13.7 and compiles the binaries directly
on the actions host
* Merges the release-master and release-tag workflows. Instead, a single workflow is triggered that will push to
the git SHA or to master. If a tag has been pushed, it should use the tag instead.

Personally, I find this a much clearer way to manage the compilation and docker build process. Instead of a large
Makefile with lots of templating, we have a single bash script for handling cross-compilation that can be added
to easily. Likewise, the docker build process also has a single bash script.


j-fuentes:
> Removes the docker_build job from the test workflow. Using go test ./... should pick up on any go build failures

I would rather run docker build on PRs to test changes in the Dockerfile itself. Maybe not in all platforms, but at least in one.

.github/workflows/release-master.yml
```diff
@@ -5,55 +5,61 @@ on:
...
       - master
+    tags:
+      - v*
 
 jobs:
   golint:
-    name: vet fmt and lint
-    runs-on: ubuntu-20.04
+    name: Lint
+    runs-on: ubuntu-latest
```
j-fuentes, line 14: I have mixed feelings about this.

In general, it is better to pin to a version, so our CI does not break when GitHub decides to make `ubuntu-latest` point to a newer version.

In this particular case, if CI breaks it is not a big deal: we will realise and come back to fix it. And since this pipeline is not a live system, it is not that time-sensitive.

There is still the risk that a change in `ubuntu-latest` breaks something but does not make our CI fail. Anyway, I think the risk is low and with `latest` we gain auto-updates, so let's give it a try.

davidsbond: I don't mind switching it back to a fixed version if we prefer that. I've been using `ubuntu-latest` in most actions and I haven't had anything break yet. We can always refix should something come up

davidsbond:
@j-fuentes 

> I would rather run docker build on PRs to test changes in the Dockerfile itself. Maybe not

Yeah I think that's fair, I'll add a job to ensure that the image builds on linux/amd64

.github/workflows/release-master.yml
```diff
@@ -5,55 +5,61 @@ on:
...
-    container:
-      image: docker:20.10.5
-      options: -t
-    # Setting up dind service container
-    services:
-      docker:
-        image: docker:20.10.5-dind
-        env:
-          DOCKER_DRIVER: overlay
-          DOCKER_HOST: tcp://localhost:2375
```
j-fuentes, line 39: Cool! I didn't realise this was not needed with the docker actions.

davidsbond: Yeah, if you aren't using a container for the entire job, the ubuntu VM comes with docker already available.

Makefile
```diff
@@ -57,69 +18,8 @@ vet:
...
-	rm -rf ./builds
-	docker build --rm -t preflight-bin -f ./builder.dockerfile \
-		--build-arg oauth_client_id=$(OAUTH_CLIENT_ID) \
-		--build-arg oauth_client_secret=$(OAUTH_CLIENT_SECRET) \
-		--build-arg oauth_auth_server_domain=$(OAUTH_AUTH_SERVER_DOMAIN) \
-		.
-	docker create --rm --name=preflight-bin-container preflight-bin
-	docker cp preflight-bin-container:/go/github.com/jetstack/preflight/builds ./builds
-	docker rm preflight-bin-container
-	docker rmi preflight-bin
```
j-fuentes, line 87: We added this just to have more reproducibility in builds: i.e. if in my local machine I have docker 1.14 but I want to build with the same version that CI uses, I can use this and avoid installing another Go version on my system.

It is not a dealbreaker, but I think it is useful.

I am ok if this goes away if we gain some speed in CI.

hack/docker.sh
```diff
@@ -0,0 +1,17 @@
...
+PLATFORMS=linux/arm/v7,linux/arm64/v8,linux/amd64
+DOCKER_IMAGE=quay.io/jetstack/preflight
+
+# Uses the git tag or commit SHA.
+VERSION="$(bash ./hack/version.sh)"
+
+docker buildx build \
+  --platform ${PLATFORMS} \
+  --tag "${DOCKER_IMAGE}:${VERSION}" \
+  --tag "${DOCKER_IMAGE}:latest" \
```
j-fuentes, line 15: Let's keep an eye on this when we merge this. I am not 100% this is going to work but it is worth a shot.

hack/docker.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Script that builds and pushes docker images for multiple architectures. If you want to include a new architecture,
+# ensure that it is compiled in the build.sh script, then add your OS/ARCH combination to the PLATFORMS variable.
+
+PLATFORMS=linux/arm/v7,linux/arm64/v8,linux/amd64
+DOCKER_IMAGE=quay.io/jetstack/preflight
+
+# Uses the git tag or commit SHA.
+VERSION="$(bash ./hack/version.sh)"
```
j-fuentes, line 10: How does this work with tags? Does it run twice one with `VERSION=<commit sha>` and one with `VERSION=<tag>`?

hack/docker.sh
davidsbond: Yeah I wasn't 100% sure it was going to fix the particular pushing problem. I was finding the build process quite hard to debug and felt like a tidy up first could make this easier to fix. It'll be great if it does too

hack/docker.sh
davidsbond: So when a tag is pushed, the `release-master` workflow will run, that `version.sh` script will then return the git tag instead of the SHA. Likewise, when `master` is pushed, the workflow will run, but since there's no tag it will return the SHA instead.

Makefile
davidsbond: Lets see how it goes without it, won't be hard to add back in if we decide we need it again. Right now our builds aren't 100% reproducible anyway since we don't vendor our dependencies. To me it didn't seem necessary for the actual CI workflow, but if it's handy locally I can add it back

hack/build.sh
```diff
@@ -0,0 +1,88 @@
...
+"-X 'github.com/jetstack/preflight/pkg/version.Commit=${COMMIT}'"\
+"-X 'github.com/jetstack/preflight/pkg/version.BuildDate=${DATE}'"\
+"-X 'github.com/jetstack/preflight/pkg/version.GoVersion=${GOVERSION}'"\
+"-X 'github.com/jetstack/preflight/pkg/client.ClientID=${OAUTH_CLIENT_ID}'"\
+"-X 'github.com/jetstack/preflight/pkg/client.ClientSecret=${OAUTH_CLIENT_SECRET}'"\
+"-X 'github.com/jetstack/preflight/pkg/client.AuthServerDomain=${OAUTH_AUTH_SERVER_DOMAIN}'" \
+  -o "${OUTPUT}"
+}
+
+compile_all
```
j-fuentes, line 88: Personally I prefer having this and `docker.sh` in the Makefile itself:

```
OAUTH_CLIENT_ID?=k3TrDbfLhCgnpAbOiiT2kIE1AbovKzjo
OAUTH_CLIENT_SECRET?=f39w_3KT9Vp0VhzcPzvh-uVbudzqCFmHER3Huj0dvHgJwVrjxsoOQPIw_1SDiCfa
OAUTH_AUTH_SERVER_DOMAIN?=auth.jetstack.io

define LDFLAGS
-X "github.com/jetstack/preflight/pkg/version.PreflightVersion=$(VERSION)" \
-X "github.com/jetstack/preflight/pkg/version.Platform=$(GOOS)/$(GOARCH)" \
-X "github.com/jetstack/preflight/pkg/version.Commit=$(COMMIT)" \
-X "github.com/jetstack/preflight/pkg/version.BuildDate=$(DATE)" \
-X "github.com/jetstack/preflight/pkg/version.GoVersion=$(GOVERSION)" \
-X "github.com/jetstack/preflight/pkg/client.ClientID=$(OAUTH_CLIENT_ID)" \
-X "github.com/jetstack/preflight/pkg/client.ClientSecret=$(OAUTH_CLIENT_SECRET)" \
-X "github.com/jetstack/preflight/pkg/client.AuthServerDomain=$(OAUTH_AUTH_SERVER_DOMAIN)"
endef

GO_BUILD:=CGO_ENABLED=0 go build -ldflags '$(LDFLAGS)'
GO_INSTALL:=CGO_ENABLED=0 go install -ldflags '$(LDFLAGS

.PHONY: ./builds/$(GOOS)/$(GOARCH)/$(BIN_NAME)
./builds/$(GOOS)/$(GOARCH)/$(BIN_NAME):
	GOOS=$(GOOS) GOARCH=$(GOARCH) $(GO_BUILD) -o ./builds/$(GOOS)/$(GOARCH)/$(BIN_NAME) .
.PHONY: ./builds/$(GOOS)/$(GOARCH)/v$(GOARM)/$(BIN_NAME)
./builds/$(GOOS)/$(GOARCH)/v$(GOARM)/$(BIN_NAME):
	GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) $(GO_BUILD) -o ./builds/$(GOOS)/$(GOARCH)/v$(GOARM)/$(BIN_NAME) .

build-all-platforms:
	$(MAKE) GOOS=linux   GOARCH=amd64       ./builds/linux/amd64/$(BIN_NAME)
	$(MAKE) GOOS=linux   GOARCH=arm64       ./builds/linux/arm64/$(BIN_NAME)
	$(MAKE) GOOS=linux   GOARCH=arm GOARM=7 ./builds/linux/arm/v7/$(BIN_NAME)
	$(MAKE) GOOS=darwin  GOARCH=amd64       ./builds/darwin/amd64/$(BIN_NAME)
	$(MAKE) GOOS=windows GOARCH=amd64       ./builds/windows/amd64/$(BIN_NAME)

# Docker image
PLATFORMS?=linux/arm/v7,linux/arm64/v8,linux/amd64
BUILDX_EXTRA_ARGS?=

push_buildx_args=--push $(BUILDX_EXTRA_ARGS)
push-canary_buildx_args=--tag $(DOCKER_IMAGE):canary --push $(BUILDX_EXTRA_ARGS)
build_buildx_args=$(BUILDX_EXTRA_ARGS)

.PHONY: _docker-%
_docker-%: build-all-platforms
	docker buildx build --platform $(PLATFORMS) \
	--tag $(DOCKER_IMAGE_TAG) \
	$($*_buildx_args) \
	.

build-docker-image: _docker-build
push-docker-image: _docker-push

```

I find it quicker to read and understand and easier to use:

- With `make`, if I run `make build-docker-image`, it knows it needs to build the binaries as well (as the dependency is encoded in the Makefile). You could mitigate that in your approach by adding a dep in the Makefile as well.

- Ideally, we want to reduce the differences between what CI does and what we can do locally, so relying on something like `make` to orchestrate the different steps (binary build + docker image) is better than having to execute several commands in certain order.

- Then, you someone is testing this repo locally, the common actions are going to be: either build a binary for their platform to try their change or build a docker image to try changes in things like Dockerfile (also in one platform first, so you can iterate faster). With our previous approach with make they can do `make build` and build just a binary or `make build-docker-image PLATFORM=linux/amd64`  and build just one docker image (and not push it).

- BONUS: with make you have input args autocompletion -> `make build-docker-image PLATF<tab>`.

j-fuentes: I find the Makefile better for those reasons, and it is not hard to read to me, but I am biased because I use Makefiles a lot.

If you and others think using scripts is simples and easier to maintain and understand, I have no problems using scripts instead.

@tfadeyi @charlieegan3 WDYT?

davidsbond: Fair, I think part of the problem I am having with it is I am not that familiar with how Makefiles work outside of just creating targets and making them call other targets beforehand. I can't look at the Makefile as is and instantly understand how the docker builds work.

Especially with sections like this:

```
.PHONY: ./builds/$(GOOS)/$(GOARCH)/$(BIN_NAME)
./builds/$(GOOS)/$(GOARCH)/$(BIN_NAME):
	GOOS=$(GOOS) GOARCH=$(GOARCH) $(GO_BUILD) -o ./builds/$(GOOS)/$(GOARCH)/$(BIN_NAME) .
.PHONY: ./builds/$(GOOS)/$(GOARCH)/v$(GOARM)/$(BIN_NAME)
./builds/$(GOOS)/$(GOARCH)/v$(GOARM)/$(BIN_NAME):
	GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) $(GO_BUILD) -o ./builds/$(GOOS)/$(GOARCH)/v$(GOARM)/$(BIN_NAME) .
```

To me this is just incredibly cryptic, and doesn't feel necessary.

j-fuentes: Let's inspect that section in particular block by block:

```
.PHONY: ./builds/$(GOOS)/$(GOARCH)/$(BIN_NAME)
```
This just means that a file like `./builds/linux/arm64/preflight` is going to be recreated everytime, even if it already exists. I would be up for removing this one in particular but I thought it might avoid confusions.

```
./builds/$(GOOS)/$(GOARCH)/$(BIN_NAME):
    GOOS=$(GOOS) GOARCH=$(GOARCH) $(GO_BUILD) -o ./builds/$(GOOS)/$(GOARCH)/$(BIN_NAME) .
```

This one defines how to build `./builds/linux/arm64/preflight` -> `GOOS=linux GOARCH=arm64 go build ... -o ./builds/linux/arm64/preflight`


```
.PHONY: ./builds/$(GOOS)/$(GOARCH)/v$(GOARM)/$(BIN_NAME) ./builds/$(GOOS)/$(GOARCH)/v$(GOARM)/$(BIN_NAME): GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) $(GO_BUILD) -o ./builds/$(GOOS)/$(GOARCH)/v$(GOARM)/$(BIN_NAME) .
```

This last block is the same as the previous ones but matching the case of builds with `GOARM`. There must be ways of merging the two blocks but maybe that goes too much to the black magic side of things.

j-fuentes: I should have added more documentation in the Makefile in the first place.

hack/docker.sh
j-fuentes: Right! Makes sense.

hack/build.sh
davidsbond: Thanks for the explanation 😄. My issue is that just to understand what these 5/6 lines do requires explanation is part of the problem. For anyone unfamiliar to make I think it will just come off as intimidating and hard to parse. 

By delegating to scripts, we can use concepts we're familiar with already, like functions and control flow. Perhaps I just need to level up my `Makefile` abilities, but from a perspective of developer-friendliness I'm just not sure this is the best way.

tfadeyi: I think the script is pretty clear, but I do like being able to build only for amd locally, could we add a way of just building the binary for amd as a default when calling the script?

hack/version.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
charlieegan3, line 1: If I see `#!/bin/sh` I read that as POSIX, so no bash needed here?

.github/workflows/test.yml
```diff
@@ -11,44 +11,48 @@ on:
...
+
+      # Install golang 1.13
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.13.7
+
+      # Compile multi-arch binaries
+      - name: Compile binaries
+        run: make build
```
charlieegan3, line 54: I wonder how important this is. Is it just confusing that we publish windows and mac images. Just because we can doesn't mean we should.

Yeah, yeah, but your scientists were so preoccupied with whether or not  they could that they didn't stop to… | Jurassic park quotes, Park quotes,  Best movie quotes

I wonder if we can skip this step and instead copy the binary from the built image, docker create, docker cp etc.

charlieegan3: Do we need to do this in test anyway?

hack/docker.sh
charlieegan3: I wonder if we can extract this and apply just this change to the existing script to test if it works?

.github/workflows/test.yml
davidsbond: So this workflow doesn't actually push any images, it just ensures that the docker image builds. The reason for compiling the binaries here is that all the docker image does now is copy the built binary into itself. So we need to build the binaries on the host VM first, so that `docker build` will pick them up.

It doesn't necessarily need to exist, however it might be nice to catch things that break the docker build in PRs rather than once merged to master.

hack/build.sh
charlieegan3: I confess that I also find the script easier to read. However, I'd like to just check what we need.

I actually build local binaries with go build manually. I don't use the makefile. Generally I use go run locally. If I needed a binary, I could work out how to make one the same from a makefile or script.

I'd like to find a system that allows us to use the same binaries in the image as we push to the release. Or have the release binaries build separately from pushing the image in a concurrent action.

I'd also like to see this split out from the actual fix: https://github.com/jetstack/preflight/pull/230/files#r621265094

.github/workflows/test.yml
charlieegan3: > is that all the docker image does now is copy the built binary into itself

How does that work for multiple architectures. I can't see where that happens.

davidsbond:
Thanks all for the comments, I think there's some places I can improve on this change based on your feedback.

1. `make build` should just spit out a binary for your current system
2. Putting the cross-compile all the binaries into its own target, so `make multi-arch` or something? Then the CI will use that target and we use `make build` locally

@charlieegan3 :

> I actually build local binaries with go build manually.

Do you do that with the agent repository just to confirm you can compile? Or do you also do all the ldflags etc locally too? Are you building to build or building to run later?

@tfadeyi, for just building amd by default, you could just use `go build`, or do you also want all the ldflags as well? In which case, I think 1 might be the change that works for you?

charlieegan3:
> Or do you also do all the ldflags etc locally too

Yeah, I have done.

> Are you building to build or building to run later?

Usually to replicate something. so yeah, I run them on my machine sometimes (once every few months I'd say)

Your suggestions sound good to me re 1 + 2

.github/workflows/test.yml
davidsbond: So for each architecture, the image is built:

```Dockerfile
# TARGETPLATFORM comes from the buildx context and it will be something like `linux/arm64/v8` or `linux/amd64`.
# Ref: https://docs.docker.com/buildx/working-with-buildx/
ARG TARGETPLATFORM
```

The `TARGETPLATFORM` is automatically set by the `buildx` tool and we use it to build the path to the correct binary, which is copied in.

```Dockerfile
COPY ./builds/${TARGETPLATFORM}/preflight /bin/preflight
```

When binaries are compiled, they're placed into a directory tree like `OS/ARCH/preflight`

davidsbond:
Closing for now, hasn't been looked at in ages and don't really need these changes

PR #233, opened 2021-05-04
Modify CI to push multiple tags

davidsbond:
The retagging commands don't work with quay.io, so instead we can use the multiple --tag parameters allowed
in the `docker buildx build` command. This makes the `release-tag` workflow redundant, as all it did was
perform that retagging.

Now, when a commit is merged to master, an image should be created for `latest`, `canary` and the commit SHA. When
a tag is pushed, an image should be created for `latest`, `canary` and the tag.


charlieegan3 (approved):

Makefile
```diff
@@ -97,17 +97,15 @@ build_buildx_args=$(BUILDX_EXTRA_ARGS)
 .PHONY: _docker-%
 _docker-%: build-all-platforms
 	docker buildx build --platform $(PLATFORMS) \
-	--tag $(DOCKER_IMAGE_TAG) \
+	--tag $(DOCKER_IMAGE):$(VERSION) \
```
j-fuentes, line 100: I am not sure this is going to work. I think I tried that in the past and also errors with 401, but it is worth another shot.

j-fuentes (approved):

PR #248, opened 2021-07-30
Add the compact flag

### LINKED ISSUE #245: Add `--compact` print flag to the echo server
I'd like to have the option to turn on and off the echo server output colors, by passing the `--compact` flag to the echo command.

if --compact is passed then the output should be without colours and formatting.

if it isn't passed to the command then the output should be with colours.

Weeblin:
Use compact flag to switch echo output to compact format.
fixes #245 
Screenshot from 2021-07-30 06-10-06

### DIFF
--- a/pkg/echo/echo.go
+++ b/pkg/echo/echo.go
@@ -14,6 +14,8 @@ import (
 
 var EchoListen, AllowedToken string
 
+var Compact bool
+
 func Echo(cmd *cobra.Command, args []string) {
 	http.HandleFunc("/", echoHandler)
 	fmt.Println("Listening to requests at ", EchoListen)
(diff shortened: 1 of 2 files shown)


pkg/echo/echo.go
```diff
@@ -45,20 +47,28 @@ func echoHandler(w http.ResponseWriter, r *http.Request) {
...
-	color.Green("-- %s %s -> created %d\n", r.Method, r.URL.Path, http.StatusCreated)
-	fmt.Printf("received %d readings:\n", len(payload.DataReadings))
+	if Compact {
+		fmt.Printf("-- %s %s -> created %d\n", r.Method, r.URL.Path, http.StatusCreated)
+		fmt.Printf("received %d readings:\n", len(payload.DataReadings))
+		for _, r := range payload.DataReadings {
+			fmt.Printf("%+v\n", r)
+		}
+	} else {
+		color.Green("-- %s %s -> created %d\n", r.Method, r.URL.Path, http.StatusCreated)
```
tfadeyi, line 57: I think it's good :+1:  we could also have a function specifically for print in colours and one to printed the compacted output, what do you think?

tfadeyi (approved):

Issue #250, opened 2021-08-02
Generate RBAC based on the current config


tfadeyi:
Generate RBAC based on the current config, and in such a way that the code can be shared with the backend.

j-fuentes:
More context on this one.

At

charlieegan3:
I've been working a little with Wenlin on this today and my plan was, if there was more time, to build this incrementally with TDD in stages, e.g.

1. GVRs in, RBAC string templated out
2. GVRs in, RBAC structs out
3. GVRs and Namespaces in, scoped rbac generated based on some list of known namespaces in the case of excluded namespaces
4. means of getting GVRs from loaded dynamic data gatherers.
5. adding a command to spit this all out.
6.

charlieegan3:
You are correct, I was imagining that it'd be possible to supply a list of namespaces that comes from kubectl get ns if the agent were running locally using the installers credentials.

I'm thinking that the first use of this is going to be a command in the agent, and if the agent, running in that command had access to the installers creds, it could get a list of all namespaces.

We might not be able to do this when using it from the platform side, however I think it's a feature worth adding here in the agent.

PR #252, opened 2021-08-04
Rbac generate

Weeblin:
Made a string of RBAC from GVRs.
#250

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -44,9 +44,9 @@ type Endpoint struct {
 }
 
 type DataGatherer struct {
-	Kind     string
-	Name     string
-	DataPath string
+	Kind     string `yaml:"kind"`
+	Name     string `yaml:"name"`
+	DataPath string `yaml:"data_path"`
 	Config   datagatherer.Config
 }
 
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -30,7 +30,7 @@ type Config struct {
 	OrganizationID string `yaml:"organization_id"`
 	// ClusterID is the cluster that the agent is scanning.
 	ClusterID     string         `yaml:"cluster_id"`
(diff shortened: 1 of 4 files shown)


tfadeyi:
Lgtm :+1:  could you add a description to the PR

tfadeyi (approved):

pkg/agent/config.go
```diff
@@ -30,7 +30,7 @@ type Config struct {
 	OrganizationID string `yaml:"organization_id"`
 	// ClusterID is the cluster that the agent is scanning.
 	ClusterID     string         `yaml:"cluster_id"`
-	DataGatherers []dataGatherer `yaml:"data-gatherers"`
+	DataGatherers []DataGatherer `yaml:"data-gatherers"`
```
jakexks, line 33: data-gatherers seems inconsistent with cluster_id and organization_id. Would _ be better?
```suggestion
	DataGatherers []DataGatherer `yaml:"data_gatherers"`
```

tfadeyi: :+1: We'll see if we can change it in another PR. Changing that now would require a couple of other changes in the platform side also.

pkg/agent/config.go
```diff
@@ -43,7 +43,7 @@ type Endpoint struct {
 	Path     string `yaml:"path"`
 }
 
-type dataGatherer struct {
+type DataGatherer struct {
 	Kind     string
 	Name     string
 	DataPath string
```
jakexks, line 49: If this field is now exported I recommend adding `yaml:""` tags
```suggestion
type DataGatherer struct {
	Kind     string `yaml:"kind"`
	Name     string `yaml:"name"`
	DataPath string `yaml:"data_path"`
```

jakexks (approved):

PR #258, opened 2021-08-24
250 command

Weeblin:
A PR for solving RBAC string formatting issue

### DIFF
--- a/agent.yaml
+++ b/agent.yaml
@@ -11,3 +11,9 @@ data-gatherers:
     name: "dummy-fail"
     config:
       always-fail: true
+  - kind: "k8s-dynamic"
+    name: "k8s/pods"
+    config:
+      resource-type:
+         resource: pods
+         version: v1
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/jetstack/preflight/pkg/agent"
+	"github.com/jetstack/preflight/pkg/datagatherer"
 	"github.com/spf13/cobra"
 )
 
@@ -25,6 +26,14 @@ var agentInfoCmd = &cobra.Command{
 		fmt.Println()
 		printOAuth2Config()
 	},
+}
+
+var agentRBACCmd = &cobra.Command{
+	Use:   "rbac",
+	Short: "print RBAC",
+	Long:  `Print RBAC string by reading GVRs`,
+	Run:    out := generateFullManifest(agent.DataGatherer)
+			fmt.Sprintf(out)
 }
 
 func init() {
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -2,10 +2,13 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
 	"time"
 
 	"github.com/jetstack/preflight/pkg/agent"
-	"github.com/jetstack/preflight/pkg/datagatherer"
+	"github.com/jetstack/preflight/pkg/permissions"
 	"github.com/spf13/cobra"
 )
 
@@ -32,13 +35,33 @@ var agentRBACCmd = &cobra.Command{
 	Use:   "rbac",
 	Short: "print RBAC",
 	Long:  `Print RBAC string by reading GVRs`,
-	Run:    out := generateFullManifest(agent.DataGatherer)
-			fmt.Sprintf(out)
+	Run: func(cmd *cobra.Command, args []string) {
+
+		file, err := os.Open(agent.ConfigFilePath)
+		if err != nil {
+			log.Fatalf("Failed to load config file for agent from: %s", agent.ConfigFilePath)
+		}
+		defer file.Close()
+
+		b, err := ioutil.ReadAll(file)
+		if err != nil {
+			log.Fatalf("Failed to read config file: %s", err)
+		}
+
+		config, err := agent.ParseConfig(b)
+		if err != nil {
+			log.Fatalf("Failed to parse config file: %s", err)
+		}
+
+		out := permissions.GenerateFullManifest(config.DataGatherers)
+		fmt.Print(out)
+	},
 }
 
 func init() {
 	rootCmd.AddCommand(agentCmd)
 	agentCmd.AddCommand(agentInfoCmd)
+	agentCmd.AddCommand(agentRBACCmd)
 	agentCmd.PersistentFlags().StringVarP(
 		&agent.ConfigFilePath,
 		"agent-config-file",
--- a/pkg/permissions/generate.go
+++ b/pkg/permissions/generate.go
@@ -1,6 +1,7 @@
 package permissions
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/jetstack/preflight/pkg/agent"
@@ -115,5 +116,14 @@ func GenerateAgentRBACManifests(dataGatherers []agent.DataGatherer) AgentRBACMan
 }
 
 func generateFullManifest(dataGatherers []agent.DataGatherer) string {
-
+	agentRBACManifestsStruct := GenerateAgentRBACManifests(dataGatherers)
+	agentRBACString, err := json.Marshal(agentRBACManifestsStruct)
+	if err != nil {
+		fmt.Print(err)
+	}
+	var out string
+	for _, s := range agentRBACString {
+		out += string(s)
+	}
+	return out
 }
--- a/pkg/permissions/generate.go
+++ b/pkg/permissions/generate.go
@@ -26,9 +26,9 @@ func AgentCLR(clusterRole []rbac.ClusterRole) string {
 	for _, clr := range clusterRole {
 		var aG, rC, vB string
 		for _, rule := range clr.Rules {
-			aG = strings.Join(rule.APIGroups, " -\n")
-			rC = strings.Join(rule.Resources, " -\n")
-			vB = strings.Join(rule.Verbs, " -\n")
+			aG = strings.Join(rule.APIGroups, "\", \"")
+			rC = strings.Join(rule.Resources, "\", \"")
+			vB = strings.Join(rule.Verbs, "\", \"")
 		}
 
 		got = append(got, fmt.Sprintf(
@@ -37,12 +37,9 @@ kind: ClusterRole
 metadata:
 	name: jetstack-secure-agent-pods-reader
 rules:
-- apiGroups: 
-	-%s
-	resources: 
-	-%s
-	verbs: 
-	-%s`, aG, rC, vB))
+- apiGroups: ["%s"]
+	resources: ["%s"]
+	verbs: ["%s"]`, aG, rC, vB))
 	}
 	out := strings.Join(got, "\n")
 	return out
@@ -76,7 +73,7 @@ func AgentRB(RB []rbac.RoleBinding) string {
 	for _, rb := range RB {
 		got = append(got, fmt.Sprintf(
 			`apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
+kind: RoleBinding
 metadata:
   name: %s
   namespaces:%s
--- a/pkg/permissions/generate.go
+++ b/pkg/permissions/generate.go
@@ -1,8 +1,8 @@
 package permissions
 
 import (
-	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/jetstack/preflight/pkg/agent"
 	"github.com/jetstack/preflight/pkg/datagatherer/k8s"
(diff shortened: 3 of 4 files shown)


pkg/permissions/generate.go
```diff
@@ -19,6 +20,80 @@ type AgentRBACManifests struct {
 	RoleBindings []rbac.RoleBinding
 }
 
+//*******************************
+func AgentCLR(clusterRole []rbac.ClusterRole) string {
```
tfadeyi, line 24:
```suggestion
func agentCLR(clusterRole []rbac.ClusterRole) string {
```

pkg/permissions/generate.go
```diff
@@ -19,6 +20,80 @@ type AgentRBACManifests struct {
...
+	var got []string
+	for _, clr := range clusterRole {
+		var aG, rC, vB string
```
tfadeyi, line 27: Can we use more expressive variable names here?

pkg/permissions/generate.go
tfadeyi: could we name the function something more clear, expressing what the function does? an example could be `createAgentClusterRoleManifests` or `makeClusterRoleManifests`

pkg/permissions/generate.go
```diff
@@ -113,3 +188,17 @@ func GenerateAgentRBACManifests(dataGatherers []agent.DataGatherer) AgentRBACMan
...
+func GenerateFullManifest(dataGatherers []agent.DataGatherer) string {
+	agentRBACManifestsStruct := GenerateAgentRBACManifests(dataGatherers)
+	agentCLR := AgentCLR(agentRBACManifestsStruct.ClusterRoles)
+	agentCLRB := AgentCLRB(agentRBACManifestsStruct.ClusterRoleBindings)
+	agentRB := AgentRB(agentRBACManifestsStruct.RoleBindings)
+
+	out := fmt.Sprintf(
+		`%s
+%s
+%s`, agentCLR, agentCLRB, agentRB)
```
tfadeyi, line 201: in this way that you have here of concatenating strings, you are not considering when either `agentCLR,agentCLRB` or `agentRB` are empty strings.

example if `agentCLRB` is an empty string:
```
        apiVersion: rbac.authorization.k8s.io/v1
        kind: ClusterRole
        metadata:
          name: jetstack-secure-agent-pods-reader
        rules:
        - apiGroups: [""]
          resources: ["pods"]
          verbs: ["get", "list", "watch"]
    --> Here you're going to have a new line for the missing  `agentCLRB`    
        ---
        apiVersion: rbac.authorization.k8s.io/v1
        kind: RoleBinding
        metadata:
          name: jetstack-secure-agent-pods-reader
          namespace: foobar
        roleRef:
          kind: ClusterRole
          name: jetstack-secure-agent-pods-reader
          apiGroup: rbac.authorization.k8s.io
        subjects:
        - kind: ServiceAccount
          name: agent
          namespace: jetstack-secure
``

Comment (tfadeyi) [reply pkg/permissions/generate.go:201]: a solution could check if a variable is an empty string before appending it to the output

File: pkg/permissions/generate.go [hunk 4]
```diff
@@ -19,6 +20,80 @@ type AgentRBACManifests struct {
...
+- kind: %s
+  name: %s
+  namespace: %s`, clrb.ObjectMeta.Name, clrb.RoleRef.Name, clrb.RoleRef.APIGroup, clrb.Subjects[0].Kind, clrb.Subjects[0].Name, clrb.Subjects[0].Namespace))
+	}
+	out := strings.Join(got, "\n")
+	return out
+}
+
+//*******************************
+func AgentRB(RB []rbac.RoleBinding) string {
```
tfadeyi, line 72:
```suggestion
func agentRB(RB []rbac.RoleBinding) string {
```

tfadeyi: could we name the function something more clear, expressing what the function does?

pkg/permissions/generate.go
```diff
@@ -19,6 +20,80 @@ type AgentRBACManifests struct {
...
+- apiGroups: ["%s"]
+	resources: ["%s"]
+	verbs: ["%s"]`, aG, rC, vB))
+	}
+	out := strings.Join(got, "\n")
+	return out
+}
+
+//*******************************
+func AgentCLRB(CLRB []rbac.ClusterRoleBinding) string {
```
tfadeyi, line 49:
```suggestion
func agentCLRB(clrbs []rbac.ClusterRoleBinding) string {
```

tfadeyi: could we name the function something more clear, expressing what the function does?

pkg/permissions/generate.go
```diff
@@ -19,6 +20,80 @@ type AgentRBACManifests struct {
...
+			aG = strings.Join(rule.APIGroups, "\", \"")
+			rC = strings.Join(rule.Resources, "\", \"")
+			vB = strings.Join(rule.Verbs, "\", \"")
+		}
+
+		got = append(got, fmt.Sprintf(
+			`apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+	name: jetstack-secure-agent-pods-reader
```
tfadeyi, line 38:
```suggestion
kind: ClusterRole
metadata:
   name: jetstack-secure-agent-pods-reader
```

tfadeyi, line 38: There are too many whitespaces before `name: jetstack-secure-agent-pods-reader`

tfadeyi: The name shouldn't be `jetstack-secure-agent-pods-reader` but should be the name present in `clr` like `clr.ObjectMeta.Name`

pkg/permissions/generate.go
tfadeyi: For variables of slices, arrays and collections usually the name of the variable should be plural name

pkg/permissions/generate.go
Weeblin: I did a tab here, which seems to be 4 spaces, and the expected and other parts are all using tab as well, so I thought it will be easier to line up.

agent.yaml
```diff
@@ -11,3 +11,9 @@ data-gatherers:
     name: "dummy-fail"
     config:
       always-fail: true
+  - kind: "k8s-dynamic"
```
tfadeyi, line 14: We can remove this data gatherer

pkg/permissions/generate_test.go
```diff
@@ -11,6 +11,107 @@ import (
...
+  name: jetstack-secure-agent-pods-reader
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: agent
+  namespace: jetstack-secure`,
+		},
+	}
+
+	// for _, input := range testCases {
```
tfadeyi, line 96: we delete thes comments

cmd/agent.go
```diff
@@ -27,9 +31,37 @@ var agentInfoCmd = &cobra.Command{
 	},
 }
 
+var agentRBACCmd = &cobra.Command{
+	Use:   "rbac",
+	Short: "print RBAC",
+	Long:  `Print RBAC string by reading GVRs`,
```
tfadeyi, line 37:
```suggestion
	Long:  `Print the agent's minimal RBAC depending on the agent's configuration`,
```

pkg/permissions/generate.go
```diff
@@ -19,6 +20,80 @@ type AgentRBACManifests struct {
 	RoleBindings []rbac.RoleBinding
 }
 
+//*******************************
+func AgentCLR(clusterRoles []rbac.ClusterRole) string {
```
tfadeyi, line 24:
```suggestion
func agentCLR(clusterRoles []rbac.ClusterRole) string {
```

cmd/agent.go
```diff
@@ -27,9 +31,37 @@ var agentInfoCmd = &cobra.Command{
...
+		file, err := os.Open(agent.ConfigFilePath)
+		if err != nil {
+			log.Fatalf("Failed to load config file for agent from: %s", agent.ConfigFilePath)
+		}
+		defer file.Close()
+
+		b, err := ioutil.ReadAll(file)
+		if err != nil {
+			log.Fatalf("Failed to read config file: %s", err)
+		}
```
davidsbond, line 49:
```suggestion
		b, err := ioutil.ReadFile(agent.ConfigFilePath)
		if err != nil {
			log.Fatalf("Failed to load config file for agent from: %s", agent.ConfigFilePath)
		}
```

This could be simplified by using the [ioutil.ReadFile](https://pkg.go.dev/io/ioutil#ReadFile) function. Just skips having to open the file and defer its closure.

cmd/agent.go
```diff
@@ -27,9 +31,37 @@ var agentInfoCmd = &cobra.Command{
...
+
+		config, err := agent.ParseConfig(b)
+		if err != nil {
+			log.Fatalf("Failed to parse config file: %s", err)
```
davidsbond, line 53: If you do the suggestion above then this is fine. However, worth pointing out that `log.Fatalf` uses `os.Exit` under the hood. Which causes the program to immediately terminate without handling your `defer` functions. So your `file.Close` would never actually be called.

pkg/permissions/generate.go
```diff
@@ -19,6 +20,80 @@ type AgentRBACManifests struct {
...
+  name: %s
+  namespace: %s`, clrb.ObjectMeta.Name, clrb.RoleRef.Name, clrb.RoleRef.APIGroup, clrb.Subjects[0].Kind, clrb.Subjects[0].Name, clrb.Subjects[0].Namespace))
+	}
+	out := strings.Join(got, "\n")
+	return out
+}
+
+//*******************************
+func AgentRB(RoleBindings []rbac.RoleBinding) string {
+	var got []string
```
davidsbond, line 73: `got` isn't a very descriptive variable name, perhaps there's a better name to use?

pkg/permissions/generate.go
```diff
@@ -19,6 +20,80 @@ type AgentRBACManifests struct {
...
+- kind: %s
+  name: %s
+  namespace: %s`, clrb.ObjectMeta.Name, clrb.RoleRef.Name, clrb.RoleRef.APIGroup, clrb.Subjects[0].Kind, clrb.Subjects[0].Name, clrb.Subjects[0].Namespace))
+	}
+	out := strings.Join(got, "\n")
+	return out
+}
+
+//*******************************
+func AgentRB(RoleBindings []rbac.RoleBinding) string {
```
davidsbond, line 72:
```suggestion
func AgentRB(roleBindings []rbac.RoleBinding) string {
```

Function parameters should be camel-cased

pkg/permissions/generate_test.go
```diff
@@ -11,6 +11,107 @@ import (
...
+				},
+			},
+			expectedRBACManifests: `apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+	name: jetstack-secure-agent-pods-reader
+rules:
+- apiGroups: [""]
+	resources: ["pods"]
+	verbs: ["get", "list", "watch"]
```
davidsbond, line 41: This won't be valid YAML. When you use `-` you're creating a list. Since we have a list of objects, these are the fields of those objects. So it needs to look like this:

```suggestion
- apiGroups: [""]
  resources: ["pods"]
  verbs: ["get", "list", "watch"]
```

pkg/permissions/generate.go
```diff
@@ -19,6 +20,80 @@ type AgentRBACManifests struct {
...
+rules:
+- apiGroups: ["%s"]
+	resources: ["%s"]
+	verbs: ["%s"]`, apiGroups, resources, verbs))
+	}
+	out := strings.Join(got, "\n")
+	return out
+}
+
+//*******************************
```
davidsbond, line 48: We don't need these comments full of asterisks.

pkg/permissions/generate.go
```diff
@@ -19,6 +20,80 @@ type AgentRBACManifests struct {
 	RoleBindings []rbac.RoleBinding
 }
 
+//*******************************
```
davidsbond, line 23: Same here, don't need comments full of asterisks

davidsbond (commented):
Overall, this looks alright except it needs some minor changes that I've suggested. 

I'm not 100% sure about the approach we're taking for building the YAML. Using the multiline strings has shown to have issues with indentation and whitespacing. I think it will work fine for now, but we may want to refactor it to use  [yaml.Marshal](https://pkg.go.dev/gopkg.in/yaml.v2#Marshal) to produce the textual form of the resources. We could use the [strings.Builder](https://pkg.go.dev/strings#Builder) type and just append all the marshalled data to it.

For example:

```go
clusterRoles := []rbac.ClusterRole{
// elements...
}

builder := strings.NewBuilder()
for _, cb := range clusterRoles {
  data, err := yaml.Marshal(cb)
  if err != nil {
   // handle error
  }

  builder.Write(data)
  builder.WriteString("\n---\n")
}

return building.String()
```

PR #260, opened 2021-09-02
250 config

Weeblin:
Save input/default config yaml file into data gatherer struct
relate to #250


pkg/configs/default.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
+	"github.com/jetstack/preflight/pkg/agent"
+	"gopkg.in/yaml.v2"
+)
+
+func GetDefaultDataGatherers() ([]agent.DataGatherer, error) {
```
davidsbond, line 12: Please add some documentation for your functions, in the format of `// GetDefaultDataGatherers ...`

pkg/configs/default.go
```diff
@@ -0,0 +1,38 @@
...
+	return getDataGatherers(defaultYAML)
+}
+
+func getDataGatherers(yamlFile []byte) ([]agent.DataGatherer, error) {
+	// this will unmarshal the data gatherer
+
+	// defaultYAML, err := ioutil.ReadFile(yamlFile)
+	// if err != nil {
+	// 	fmt.Print("fail to read the file")
+	// }
```
davidsbond, line 33: Please don't leave commented out code in the files, just remove it

pkg/configs/default.go
```diff
@@ -0,0 +1,38 @@
...
+	"io/ioutil"
+	"path/filepath"
+
+	"github.com/jetstack/preflight/pkg/agent"
+	"gopkg.in/yaml.v2"
+)
+
+func GetDefaultDataGatherers() ([]agent.DataGatherer, error) {
+	// This will read the default.yaml and call the getDataGatherers()
+	filename, err := filepath.Abs("./default.yaml")
```
davidsbond, line 14: This works because the file is local to your code in the repository. This won't be the case when the binary is built. You'll want to embed the file into the binary, which you can do with the `embed` package. 

To do this, you'll need to add an import at the top of the file `_ "embed"`. Then create a package-level variable like this:

```go
//go:embed default.yaml
var dataGatherersYAML string
```

This tells the go compiler that there is a static file you want to embed into the binary, and that its content should be put into the `dataGatherersYAML` string.

You can read more about how embed works [here](https://pkg.go.dev/embed)

j-fuentes:
Abandoned

PR #264, opened 2022-01-11
pkg/agent: fix dropped errors

alrs:
This fixes three dropped `err` variables in `pkg/agent`.

Signed-off-by: Lars Lehtonen <lars.lehtonen@gmail.com>


tfadeyi (approved):
Thank you very much for spotting those! LGTM

PR #265, opened 2022-01-28
Open shift route filter

mol-george:
This PR is to exclude sensitive fields from OpenShift Route resources.

It is likely further changes are required before the PR can be approved & merged.

Two of the main issues I can think of are:
1. I was not able to establish with absolute certainty what fields the **Route** resource has and which of these we want to keep. If the [defined fields](https://github.com/jetstack/jetstack-secure/blob/2605a7512be45e3b3007c48686e9866ce9953052/pkg/datagatherer/k8s/fieldfilter.go#L28-L46) are not correct, further changes are required. [This](https://docs.openshift.com/container-platform/4.8/networking/routes/route-configuration.html#nw-ingress-creating-a-route-via-an-ingress_route-configuration) (see the _YAML Definition of an autogenerated route_) so far is the most relevant OpenShift documentation I could find that lists the fields.
2. I am not sure if the [logic I used to select fields for secrets and routes](https://github.com/jetstack/jetstack-secure/blob/2605a7512be45e3b3007c48686e9866ce9953052/pkg/datagatherer/k8s/dynamic.go#L279-L297) e.g. why/when should I be breaking out of the inner loop
3. I am not entirely sure of Routes api group [seems to be `route.openshift.io`](https://github.com/openshift/router/blob/5ecca61eb85d72f2f38564daa33410c4655e429e/deploy/route_crd.yaml#L8) but I am not sure if my assumption is correct.

I assume there are other things that need to be addressed so this is at this stage still a work in progress.

### DIFF
--- a/pkg/datagatherer/k8s/fieldfilter.go
+++ b/pkg/datagatherer/k8s/fieldfilter.go
@@ -25,6 +25,26 @@ var SecretSelectedFields = []string{
 	"/data/ca.crt",
 }
 
+// RouteSelectedFields is the list of fields sent from OpenShift Route objects to the
+// backend
+var RouteSelectedFields = []string{
+	"kind",
+	"apiVersion",
+	"metadata.annotations",
+	"metadata.name",
+	"metadata.namespace",
+	"metadata.ownerReferences",
+	"metadata.selfLink",
+	"metadata.uid",
+	"spec.host",
+	"spec.to.kind",
+	"spec.to.name",
+	"spec.tls.termination",
+	"spec.tls.certificate",
+	"spec.tls.caCertificate",
+	"spec.tls.destinationCACertificate",
+}
+
 // RedactFields are removed from all objects
 var RedactFields = []string{
 	"metadata.managedFields",
--- a/pkg/datagatherer/k8s/fieldfilter.go
+++ b/pkg/datagatherer/k8s/fieldfilter.go
@@ -38,11 +38,16 @@ var RouteSelectedFields = []string{
 	"metadata.uid",
 	"spec.host",
 	"spec.to.kind",
+	"spec.to.port",
 	"spec.to.name",
+	"spec.to.weight",
 	"spec.tls.termination",
 	"spec.tls.certificate",
 	"spec.tls.caCertificate",
 	"spec.tls.destinationCACertificate",
+	"spec.tls.insecureEdgeTerminationPolicy",
+	"spec.wildcardPolicy",
+	"status",
 }
 
 // RedactFields are removed from all objects
--- a/pkg/datagatherer/k8s/dynamic.go
+++ b/pkg/datagatherer/k8s/dynamic.go
@@ -282,7 +282,7 @@ func redactList(list []*api.GatheredResource) error {
 				Select(SecretSelectedFields, resource)
 
 				// route object
-			} else if gvk.Kind == "Route" && gvk.Group == "apiextensions.k8s.io" {
+			} else if gvk.Kind == "Route" && gvk.Group == "route.openshift.io" {
 				Select(RouteSelectedFields, resource)
 
 				// break once a Secret or Route object has been processed as, no
--- a/pkg/datagatherer/k8s/dynamic.go
+++ b/pkg/datagatherer/k8s/dynamic.go
@@ -281,15 +281,11 @@ func redactList(list []*api.GatheredResource) error {
 			if gvk.Kind == "Secret" && (gvk.Group == "core" || gvk.Group == "") {
 				Select(SecretSelectedFields, resource)
 
-				// break when the object has been processed as a secret, no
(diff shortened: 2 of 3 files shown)


pkg/datagatherer/k8s/fieldfilter.go
```diff
@@ -25,6 +25,26 @@ var SecretSelectedFields = []string{
...
+	"metadata.ownerReferences",
+	"metadata.selfLink",
+	"metadata.uid",
+	"spec.host",
+	"spec.to.kind",
+	"spec.to.name",
+	"spec.tls.termination",
+	"spec.tls.certificate",
+	"spec.tls.caCertificate",
+	"spec.tls.destinationCACertificate",
```
j-fuentes, line 45: From the [docs](https://docs.openshift.com/container-platform/3.11/rest_api/route_openshift_io/route-route-openshift-io-v1.html), I only see potential secret info in `spec.tls`.

```suggestion
	"kind",
	"apiVersion",
	"metadata.annotations",
	"metadata.name",
	"metadata.namespace",
	"metadata.ownerReferences",
	"metadata.selfLink",
	"metadata.uid",
	"spec.host",
	"spec.to.kind",
	"spec.to.port",
	"spec.to.name",
	"spec.to.weight",
	"spec.tls.termination",
	"spec.tls.certificate",
	"spec.tls.caCertificate",
	"spec.tls.destinationCACertificate",
	"spec.tls.insecureEdgeTerminationPolicy",
	"spec.wildcardPolicy",
	"status",
```

mol-george: Thank you @j-fuentes that looks good!

mol-george:
Regarding my earlier comment:
>3.  I am not sure if the [logic I used to select fields for secrets and routes](https://github.com/jetstack/jetstack-secure/blob/2605a7512be45e3b3007c48686e9866ce9953052/pkg/datagatherer/k8s/dynamic.go#L279-L297) e.g. why/when should I be breaking out of the inner loop

@j-fuentes I did not realised that I must have linked to a specific commit. The latest version of the same code is [this](https://github.com/jetstack/jetstack-secure/blob/6b91e848b8373331cdf30153e276e63fd5d11488/pkg/datagatherer/k8s/dynamic.go#L279-L293).

j-fuentes:
> Regarding my earlier comment:
> 
> > **2**.  I am not sure if the [logic I used to select fields for secrets and routes](https://github.com/jetstack/jetstack-secure/blob/2605a7512be45e3b3007c48686e9866ce9953052/pkg/datagatherer/k8s/dynamic.go#L279-L297) e.g. why/when should I be breaking out of the inner loop
> 
> @j-fuentes I did not realised that I must have linked to a specific commit. The latest version of the same code is [this](https://github.com/jetstack/jetstack-secure/blob/6b91e848b8373331cdf30153e276e63fd5d11488/pkg/datagatherer/k8s/dynamic.go#L279-L293).

So originally we had:

```golang
func redactList(list []*api.GatheredResource) error {
	for i := range list {
		item := list[i].Resource.(*unstructured.Unstructured)
		// Determine the kind of items in case this is a generic 'mixed' list.
		gvks, _, err := scheme.Scheme.ObjectKinds(item)
		if err != nil {
			return errors.WithStack(err)
		}

		resource := item

		for _, gvk := range gvks {
			// If this item is a Secret then we need to redact it.
			if gvk.Kind == "Secret" && (gvk.Group == "core" || gvk.Group == "") {
				Select(SecretSelectedFields, resource)

				// break when the object has been processed as a secret, no
				// other kinds have redact modifications
				break
			}
		}
		// remove managedFields from all resources
		Redact(RedactFields, resource)
	}
	return nil
}
```

There is this `gvks, _, err := scheme.Scheme.ObjectKinds(item)` part that was confusing me. That [all possible group,version,kind of the go object](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#Scheme.ObjectKinds). So syntactically it can be more than one. I don't think that is the case of the resources we are dealing with but let's stick to what the k8s library telling us.

In that case, I would remove the `break` completely, and the reason is: so we filter out all the possible fields that match our filters.

mol-george:
>

mol-george:
> In that case, I would remove the `break` completely, and the reason is: so we filter out all the possible fields that match our filters.
Agreed. Did so.

Unrelated, for whatever reason the latest 2 commits did not show up initially in the PR but they did after I closed and reopened the PR - probably some eventual consistency issue.

```sh
jetstack-secure on 🌱 open-shift-route-filter via 🐳 default via 🐹 v1.17.6
➜ git lg | head -3
dbb8300 minor: removed extraneous new line    <----
26ab2bf removed redactList() for loop's break    <----
407468c Update pkg/datagatherer/k8s/fieldfilter.go
```

j-fuentes (approved):
Thank you, George!

PR #267, opened 2022-02-21
Add signing, provenance and SBOM

ribbybibby:
Signs the agent container image with cosign's keyless signing and Github OIDC. Also signs and attaches a SLSA provenance attestation and a CycloneDX SBOM.

The provenance provided takes the agent up to SLSA 1.

The signatures and other attachments are stored in a separate repository to the actual container images. There are two main reasons for this:
1. Although quay.io supports cosign signatures as of the latest release, it doesn't support uploading attestations or SBOMs.
2. Pulling SBOMs, attestations and signatures for verification can skew image pull metrics for the repository. The number of pulls strikes me as something we might want to track reliably for the agent.

### DIFF
--- a/.github/workflows/release-master.yml
+++ b/.github/workflows/release-master.yml
@@ -44,7 +44,7 @@ jobs:
       id-token: write
     steps:
     - name: Install Tools
-      run: apk add --update make git jq rsync curl bash
+      run: apk add --update make git jq rsync curl bash coreutils
     - name: Install cosign
       uses: sigstore/cosign-installer@51f8e5c6fce54e46006ae97d73b2b6315f518752
       with:
--- a/.github/workflows/release-master.yml
+++ b/.github/workflows/release-master.yml
@@ -38,9 +38,19 @@ jobs:
         env:
           DOCKER_DRIVER: overlay
           DOCKER_HOST: tcp://localhost:2375
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
     steps:
     - name: Install Tools
-      run: apk add --update make git jq rsync curl
+      run: apk add --update make git jq rsync curl bash
+    - name: Install cosign
+      uses: sigstore/cosign-installer@51f8e5c6fce54e46006ae97d73b2b6315f518752
+      with:
+        cosign-release: v1.5.2
+    - name: Install Syft
+      uses: anchore/sbom-action/download-syft@3626d7d7b13e87ee6c6f9ded3940dea05a3967bc
     - uses: actions/checkout@v2.4.0
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v1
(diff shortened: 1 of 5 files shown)


.github/workflows/release-master.yml
```diff
@@ -38,21 +38,59 @@ jobs:
...
           DOCKER_DRIVER: overlay
           DOCKER_HOST: tcp://localhost:2375
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
     steps:
     - name: Install Tools
-      run: apk add --update make git jq rsync curl
+      run: apk add --update make git jq rsync curl bash coreutils
```
ChaosInTheCRD, line 47: just so understand, what's the need for adding `bash` and `coreutils` here?

Makefile
```diff
@@ -106,6 +106,23 @@ _docker-%: build-all-platforms
...
+export COSIGN_REPOSITORY?=ghcr.io/jetstack/jetstack-secure/cosign
+export COSIGN_EXPERIMENTAL=1
+
+.PHONY: sign-docker-image
+sign-docker-image:
+	@cosign sign $(DOCKER_IMAGE):$(VERSION)
+
+.PHONY: sbom-docker-image
+sbom-docker-image:
+	@syft $(DOCKER_IMAGE):$(VERSION) -o cyclonedx > bom.xml
```
ChaosInTheCRD, line 118: I guess there's no need to create two separate SBOMs (one for source and one for base image) because `syft` is managing to pickup all the Go dependencies within the compiled image?

.github/workflows/release-master.yml
ribbybibby: It's installing `bash` because the `cosign-installer` action requires it. 

The version of `base64` in the container image doesn't support the `-w` flag which the `slsa-provenance-action` requires, hence installing `coreutils` to get a more up-to-date version of `base64`.

I'll make this clear in a comment.

Makefile
ribbybibby: Yep, you'd just end up with duplication.

docs/guides/cosign/guide.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Software Bill of Materials (SBOM)
+
+To verify and download the SBOM:
+
+1. Ensure `cosign` is installed
+2. Configure the signature repository and enable experimental features:
+
+```
+export COSIGN_REPOSITORY=ghcr.io/jetstack/jetstack-secure/cosign
```
ChaosInTheCRD, line 70: ah so will this tell cosign to look for the metadata associated to this image in ghcr?

ribbybibby: Yep, it will look for signatures/attestations/SBOMs in that repository rather than the target repository.

Makefile
ChaosInTheCRD: As long as we are sure that this will pick up all dependencies?

ribbybibby:
```
$ pwd
/home/ribbybibby/go/src/github.com/jetstack/jetstack-secure

$ syft dir:. | grep -v go-module
 ✔ Indexed .
 ✔ Cataloged packages              [128 packages]
NAME                                                       VERSION                               TYPE
```
^ the only dependencies that it picks up from the source are go-modules which are also extracted when scanning the container image

j-fuentes (approved):

Makefile
ChaosInTheCRD: was just sanity checking - awesome 😄

dbarranco (commented):
late review, sorry. LGTM.
Thanks lots!

PR #286, opened 2022-03-02
Add pprof profiling endpoints to the agent

tfadeyi:
This adds a pprof server running on 6060 to aid in investigating memory
usage of the agents. This can be eneabled or disable through a flag
`profiling`

Signed-off-by: oluwole.fadeyi <oluwole.fadeyi@jetstack.io>

### DIFF
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -117,4 +117,11 @@ func init() {
 		os.Getenv("API_TOKEN"),
 		"Token used for authentication when API tokens are in use on the backend",
 	)
+	agentCmd.PersistentFlags().BoolVarP(
+		&agent.Profiling,
+		"enable-pprof",
+		"",
+		false,
+		"Enables the pprof profiling server on the agent (port: 6060).",
+	)
 }
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -4,9 +4,12 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/http"
+	_ "net/http/pprof"
 	"net/url"
 	"os"
 	"strings"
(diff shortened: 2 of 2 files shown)


pkg/agent/run.go
```diff
@@ -64,6 +69,14 @@ func Run(cmd *cobra.Command, args []string) {
 	defer cancel()
 	config, preflightClient := getConfiguration()
 
+	if Profiling {
+		log.Printf("pprof profiling was enabled.\nRunning profiling on port :6060")
+		go func() {
+			err := http.ListenAndServe(":6060", nil)
+			log.Fatalf("failed to run pprof profiler: %s", err)
```
davidsbond, line 76: This will exit the application regardless of the error that `ListenAndServe` returns, we shouldn't call `Fatal` for a `nil` error or for `http.ErrServerStopped` (I think that's what the error is called)

cmd/agent.go
```diff
@@ -117,4 +117,11 @@ func init() {
 		os.Getenv("API_TOKEN"),
 		"Token used for authentication when API tokens are in use on the backend",
 	)
+	agentCmd.PersistentFlags().BoolVarP(
+		&agent.Profiling,
+		"profiling",
```
davidsbond, line 122:
```suggestion
		"enable-pprof",
```

I think this would be a better flag to describe what it does.

pkg/agent/run.go
```diff
@@ -64,6 +70,16 @@ func Run(cmd *cobra.Command, args []string) {
 	defer cancel()
 	config, preflightClient := getConfiguration()
 
+	if Profiling {
+		log.Printf("pprof profiling was enabled.\nRunning profiling on port :6060")
+		go func() {
+			err := http.ListenAndServe(":6060", nil)
+			if err != nil && errors.Is(err, http.ErrServerClosed) {
```
davidsbond, line 77:
```suggestion
			if err != nil && !errors.Is(err, http.ErrServerClosed) {
```

davidsbond (approved):

PR #308, opened 2022-04-01
Refactor k8d/dynamic datagatherer

tfadeyi:
The k8s/dynamic is now going to use a sharedinformer for k8s native
v1 resources, instead of using a dynamicinformer. This reduceses the
current memory usage by about 25MB.

Before the refactoring, the memory usage was around ~117MB https://github.com/jetstack/jetstack-secure/pull/306

**After refactoring** the average memory usage is ~92MB
image

**After refactoring profiling**
out-new-refactoring-openshift-VI-all

Signed-off-by: Oluwole Fadeyi <oluwole.fadeyi@jetstack.io>

### DIFF
--- a/pkg/datagatherer/k8s/cache.go
+++ b/pkg/datagatherer/k8s/cache.go
@@ -6,7 +6,7 @@ import (
 
 	"github.com/jetstack/preflight/api"
 	"github.com/pmylund/go-cache"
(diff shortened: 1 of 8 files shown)


tfadeyi:
The PR is ready for a review.

But currently the versionchecker datagatherer unit test seems to be fail currently, going to fix it.

davidsbond (approved):

PR #329, opened 2022-05-11
Remove replace directives and get modules upgraded

davidsbond:
This commit removes the replace directives within the go.mod file and upgrades
dependencies that can be upgraded. It also upgrades the module to go 1.18 and
removes deprecated tools like golint.

The rationale here is that we've left this module for so long things have gotten
quite out of date and automatic upgrades now usually fail due to dependencies
pinned by the replace directive. We also depend on this module internally
so the issues propagate there too.


tfadeyi:
I see, hopefully this PR should allow us to update the other dependencies, like client-go and k8s api.

I think removing the istio data gatherer package should be fine, their library as moved packages around quite a bit making a bit difficult to maintain on our side. It would be nice to have a way to save some of Wil's work on this though.

tfadeyi (approved):

PR #445, opened 2023-06-07
TLS Protect Cloud Authentication Addition

d0ktadr3y:
This PR adds the ability to add authentication for TLS Protect Cloud to the agent for uploading cluster information. A new flag is introduced to pass a new type of credentials file.

### DIFF
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -77,11 +77,11 @@ func init() {
 		"Location of the credentials file. For OAuth2 based authentication.",
 	)
 	agentCmd.PersistentFlags().StringVarP(
-		&agent.TLSProtectCloudCredentialsPath,
-		"tls-protect-cloud-credentials-file",
-		"t",
+		&agent.VenafiSvcAccountCredentialsPath,
+		"venafi-svc-account-credentials-file",
 		"",
-		"Location of the TLS Protect Cloud credentials file. For TLS Protect Cloud based authentication.",
+		"",
+		"Location of the Venafi Cloud service account credentials file. For Venafi Cloud service account based authentication.",
 	)
 	agentCmd.PersistentFlags().BoolVarP(
 		&agent.OneShot,
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -76,12 +76,12 @@ func init() {
 		"",
 		"Location of the credentials file. For OAuth2 based authentication.",
 	)
-	agentCmd.PersistentFlags().StringVarP(
-		&agent.VenafiSvcAccountCredentialsPath,
-		"venafi-svc-account-credentials-file",
+	agentCmd.PersistentFlags().BoolVarP(
+		&agent.VenafiCloudMode,
+		"venafi-cloud",
 		"",
-		"",
-		"Location of the Venafi Cloud service account credentials file. For Venafi Cloud service account based authentication.",
+		false,
+		"Runs agent with parsing config and credentials file in Venafi Cloud format if true.",
 	)
 	agentCmd.PersistentFlags().BoolVarP(
 		&agent.OneShot,
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -76,6 +76,13 @@ func init() {
 		"",
 		"Location of the credentials file. For OAuth2 based authentication.",
 	)
+	agentCmd.PersistentFlags().StringVarP(
+		&agent.TLSProtectCloudCredentialsPath,
+		"tls-protect-cloud-credentials-file",
+		"t",
+		"",
+		"Location of the TLS Protect Cloud credentials file. For TLS Protect Cloud based authentication.",
+	)
 	agentCmd.PersistentFlags().BoolVarP(
 		&agent.OneShot,
 		"one-shot",
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -254,21 +254,7 @@ func getConfiguration() (Config, client.Client) {
 	var preflightClient client.Client
 	switch {
 	case credentials != nil:
-		switch creds := credentials.(type) {
-		case *client.VenafiSvcAccountCredentials:
-			log.Println("Venafi Cloud mode was specified, using Venafi Service Account authentication.")
-			// check if config has Venafi Cloud data
-			if config.VenafiCloud == nil {
-				log.Fatalf("Failed to find config for venafi-cloud: required for Venafi Cloud auth")
-			}
-			preflightClient, err = client.NewVenafiCloudClient(agentMetadata, creds, baseURL, config.VenafiCloud.UploadID, config.VenafiCloud.UploadPath)
-
-		case *client.OAuthCredentials:
-			log.Println("A credentials file was specified, using oauth authentication.")
-			preflightClient, err = client.NewOAuthClient(agentMetadata, creds, baseURL)
-		default:
-			err = errors.New("credentials file is in unknown format")
-		}
+		preflightClient, err = createCredentialClient(credentials, config, agentMetadata, baseURL)
 	case APIToken != "":
 		log.Println("An API token was specified, using API token authentication.")
 		preflightClient, err = client.NewAPITokenClient(agentMetadata, APIToken, baseURL)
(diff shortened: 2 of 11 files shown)


cmd/agent.go
```diff
@@ -76,6 +76,13 @@ func init() {
 		"",
 		"Location of the credentials file. For OAuth2 based authentication.",
 	)
+	agentCmd.PersistentFlags().StringVarP(
+		&agent.TLSProtectCloudCredentialsPath,
+		"tls-protect-cloud-credentials-file",
```
j-fuentes, line 81:
```suggestion
		"venafi-svc-account-credentials-file",
```

cmd/agent.go
```diff
@@ -76,6 +76,13 @@ func init() {
...
+		"t",
```
j-fuentes, line 82:
```suggestion
		"",
```

j-fuentes: I cannot think of a good short name

cmd/agent.go
```diff
@@ -76,6 +76,13 @@ func init() {
...
+		"",
+		"Location of the TLS Protect Cloud credentials file. For TLS Protect Cloud based authentication.",
```
j-fuentes, line 84:
```suggestion
		"Location of the Venafi Cloud service account credentials file. For Venafi Cloud service account based authentication.",
```

j-fuentes (commented):
I think it all looks great. The only suggested changes are to make the configuration clearer.

j-fuentes (approved):

tfadeyi:
Just a couple of notes from going through the PR:
* Add boolean flag if agent needs to use VaaS auth, i.e: --venafiCloud . This will be update the value of venafiCloudAuth 
* Is there a need to specify a different flag for the VenafiSvcAccountCredentialsPath , we could just use the --credentials flag and just have a check on the --venafiCloud to see how to parse the --credentials either use client.ParseVenafiSvcAccountCredentials(b) or client.ParseCredentials(b).
* Do we need the endpoint path to be configurable? Not sure if venafi-cloud.upload_path is really needed since the endpoint won't change much.
   *  In the case the endpoint changes, in the backend we are going to change in the endpoint version, from v1/uploads -> v2/uploads. We could have a field in the agent to specify the api version, `venafi-cloud.api-version`
   
These are non blocking.

d0ktadr3y:
Thanks for your comments @tfadeyi, here are some responses to those.
> * Add boolean flag if agent needs to use VaaS auth, i.e: --venafiCloud . This will be update the value of venafiCloudAuth
> * we could just use the --credentials flag and just have a check on the --venafiCloud to see how to parse the --credentials either use client.ParseVenafiSvcAccountCredentials(b) or client.ParseCredentials(b)

I like the combination of both of these, it makes more sense and cleans up the usability for the user. Thanks for the suggestion, I will incorporate this!

> * Do we need the endpoint path to be configurable? Not sure if venafi-cloud.upload_path is really needed since the endpoint won't change much.

This was mostly done to allow users to utilize a proxy if it is set up in their environment. Not sure if we should default this value either (api version or path change) so probably will leave it right now.

pkg/agent/run.go
```diff
@@ -247,8 +254,21 @@ func getConfiguration() (Config, client.Client) {
 	var preflightClient client.Client
 	switch {
 	case credentials != nil:
-		log.Println("A credentials file was specified, using oauth authentication.")
-		preflightClient, err = client.NewOAuthClient(agentMetadata, credentials, baseURL)
+		switch creds := credentials.(type) {
```
tfadeyi, line 257: Suggestion: can we have the nested switch in a separate function?

pkg/client/client.go
```diff
@@ -15,6 +15,11 @@ type (
 		PostDataReadings(orgID, clusterID string, readings []*api.DataReading) error
 		Post(path string, body io.Reader) (*http.Response, error)
 	}
+
```
tfadeyi, line 18: Nitpick: could you add a comment for the new interface?

pkg/client/client_oauth.go
```diff
@@ -214,28 +214,28 @@ func (c *OAuthClient) renewAccessToken() error {
 	return nil
 }
 
-// ParseCredentials reads credentials into a struct used. Performs validations.
-func ParseCredentials(data []byte) (*Credentials, error) {
-	var credentials Credentials
+// ParseOAuthCredentials reads credentials into an OAuthCredentials struct. Performs validations.
+func ParseOAuthCredentials(data []byte) (Credentials, error) {
```
tfadeyi, line 218: Nitpick: I'd expect a `ParseOAuthCredentials` to return a `OAuthCredentials` concrete struct directly rather than the interface. currently a caller calling ParseOAuthCredentials would have to do a typecheck for `OAuthCredentials` on the return item even if the method is called `ParseOAuthCredentials`.

pkg/client/client_venafi_cloud.go
```diff
@@ -0,0 +1,377 @@
...
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/golang-jwt/jwt/v4"
+	"github.com/google/uuid"
+	"github.com/hashicorp/go-multierror"
+	"github.com/jetstack/preflight/api"
+	"gopkg.in/yaml.v2"
```
tfadeyi, line 29: Nitpick: We should try and use the latest version of the package, `gopkg.in/yaml.v3`

pkg/client/client_venafi_cloud.go
```diff
@@ -0,0 +1,377 @@
...
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+	"io"
+	"io/ioutil"
```
tfadeyi, line 14: Nitpick: We should try and not use the io/ioutil, in new code since it's deprecated

pkg/client/client_venafi_cloud.go
```diff
@@ -0,0 +1,377 @@
...
+		uploadID      string
+		uploadPath    string
+		privateKey    crypto.PrivateKey
+		jwtSigningAlg jwt.SigningMethod
+		lock          sync.RWMutex
+	}
+
+	VenafiSvcAccountCredentials struct {
+		// ClientID is the service account client ID
+		ClientID string `yaml:"client_id,omitempty"`
```
tfadeyi, line 56: Question: was there a reason for using yaml instead of json for the credentials?

tfadeyi (approved):
lgtm, some minor comments

tfadeyi:
Nitpick: we might want to change the uploadID to uploaderID

pkg/client/client_venafi_cloud.go
j-fuentes: I think because that is the format of the credentials from Venafi cloud, but let's wait for Andrey to confirm

j-fuentes (approved):
Approve, module comments from Olu

pkg/client/client_venafi_cloud.go
d0ktadr3y: Yes, it was copied from another way to read in those credentials. Will change to json to match OAuth.

pkg/client/client_venafi_cloud.go
d0ktadr3y: Good to just use `io.ReadAll`?

pkg/client/client_venafi_cloud.go
d0ktadr3y: With the move to json, this is no longer required here.

pkg/client/client_oauth.go
d0ktadr3y: You're right, it defeats the purpose. I've updated it to return the correct struct.

pkg/agent/run.go
d0ktadr3y: Yes, moved.

d0ktadr3y:
@j-fuentes @tfadeyi, please take another look and let me know if this works. If anything, we can merge it and create another PR to make changes if we need to.

tfadeyi:
lgtm :+1:

pkg/client/client_venafi_cloud.go
```diff
@@ -0,0 +1,375 @@
...
+	if err != nil {
+		return err
+	}
+
+	c.lock.Lock()
+	c.accessToken = &venafiCloudAccessToken{
+		accessToken:    accessToken.AccessToken,
+		expirationTime: now.Add(time.Duration(accessToken.ExpiresIn) * time.Second),
+	}
+	c.lock.Unlock()
```
maelvls, line 256: Hey, passing by: why is `c.accessToken` only locked during the write operation but not locked during read operations?

As it is, this mutex does protect against two simultaneous updates (since two HTTP requests might trigger `updateAccessToken` simultaneously, but another HTTP request might be reading `c.accessToken` while it is being written, defeating the purpose of the mutex. Note that we wouldn't be able to catch this problem with `go tests -race` because the client doesn't have unit tests yet.

pkg/client/client_venafi_cloud.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+
+	values := url.Values{}
+	values.Set("grant_type", requiredGrantType)
+	values.Set("assertion", jwtToken)
+
+	tokenURL := fullURL(c.baseURL, accessTokenEndpoint)
+	if err != nil {
+		return err
+	}
```
maelvls, line 233: **Passing by:** I guess this `err != nil` needed 😅

PR #458, opened 2023-07-17
Refactor the venafi-cloud configuration

tfadeyi:
These changes solidify the way users configure the agent to communicate to venafi-cloud. It allows the agent to set the `--venafi-cloud` without having to define 

```yaml
venafi-cloud:
  uploader_id: ""
  upload_path: ""
```
 in the `config.yaml`.
 
 The yaml configuration can still be used for a non default agent setup. It will override the default `uploader_id` and `upload_path`.

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -127,7 +127,7 @@ func (c *Config) Dump() (string, error) {
 	return string(d), nil
 }
 
-func (c *Config) validate(isVenafi bool) error {
+func (c *Config) validate(isVenafiCloudMode bool) error {
 	var result *multierror.Error
 
 	// configured for Venafi Cloud
@@ -139,7 +139,7 @@ func (c *Config) validate(isVenafi bool) error {
 		if _, err := url.Parse(c.VenafiCloud.UploadPath); err != nil {
 			result = multierror.Append(result, fmt.Errorf("upload_path is not a valid URL"))
 		}
-	} else if !isVenafi {
+	} else if !isVenafiCloudMode {
 		if c.OrganizationID == "" {
 			result = multierror.Append(result, fmt.Errorf("organization_id is required"))
 		}
@@ -167,7 +167,7 @@ func (c *Config) validate(isVenafi bool) error {
 }
 
 // ParseConfig reads config into a struct used to configure running agents
-func ParseConfig(data []byte, isVenafi bool) (Config, error) {
+func ParseConfig(data []byte, isVenafiCloudMode bool) (Config, error) {
 	var config Config
 
 	err := yaml.Unmarshal(data, &config)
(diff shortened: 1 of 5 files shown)


j-fuentes (approved):
lgtm, but let's wait to see if @d0ktadr3y has comments

pkg/agent/config.go
```diff
@@ -126,22 +127,19 @@ func (c *Config) Dump() (string, error) {
 	return string(d), nil
 }
 
-func (c *Config) validate() error {
+func (c *Config) validate(isVenafi bool) error {
```
aashishlohani, line 130: @tfadeyi for clarity should we change this parameter name to 'isVenafiCloudMode'?

pkg/agent/run.go
```diff
@@ -274,11 +279,15 @@ func createCredentialClient(credentials client.Credentials, config Config, agent
 	switch creds := credentials.(type) {
 	case *client.VenafiSvcAccountCredentials:
 		log.Println("Venafi Cloud mode was specified, using Venafi Service Account authentication.")
-		// check if config has Venafi Cloud data
-		if config.VenafiCloud == nil {
-			log.Fatalf("Failed to find config for venafi-cloud: required for Venafi Cloud mode")
+		// check if config has Venafi Cloud data, use config data if it's present
+		uploaderID := creds.ClientID
+		uploadPath := ""
```
d0ktadr3y, line 284: Should we set the uploadPath to the default here?

d0ktadr3y (approved):
Good work, thanks for catching those things for me @tfadeyi. Just a few questions and naming things but LGTM.

pkg/agent/run.go
tfadeyi: Yeah, I was think about it, but I thought maybe the default value of the uploadPath should be dictated by the `client` package

d0ktadr3y: @tfadeyi okay, I'm fine with that!

pkg/agent/config.go
tfadeyi: I thought that but it doesn't sound well with name of the function `validate`. I can change it

d0ktadr3y: Thanks @tfadeyi, we can always change it again. I think when we may have to make changes again at some point when we fully transition.

pkg/client/client_venafi_cloud.go
```diff
@@ -70,9 +70,11 @@ type (
 )
 
 const (
-	vaasProdURL         = "https://api.venafi.cloud"
-	accessTokenEndpoint = "/v1/oauth/token/serviceaccount"
-	requiredGrantType   = "urn:ietf:params:oauth:grant-type:jwt-bearer"
+	// URL for the venafi-cloud backend services
+	VenafiCloudProdURL               = "https://api.venafi.cloud"
+	defaultVenafiCloudUploadEndpoint = "v1/tlspk/uploads"
```
maelvls, line 75: Why is this different from the value in [the Venafi Kubernetes Agent Helm chart](https://github.com/jetstack/jetstack-secure/blob/473277b69510d75734a1a714cf09d47583ba53e2/deploy/charts/venafi-kubernetes-agent/templates/configmap.yaml#L17), which looks like:

```yaml
```

Does it mean that the default value doesn't work as of today?

PR #459, opened 2023-07-18
Allow server to be configured with `--venafi-cloud`

tfadeyi:
This allows the server URL to be specified in the agent.yaml even when `--venafi-cloud` flag is present.

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -177,7 +177,7 @@ func ParseConfig(data []byte, isVenafiCloudMode bool) (Config, error) {
 
 	if config.Server == "" && config.Endpoint.Host == "" && config.Endpoint.Path == "" {
 		config.Server = "https://preflight.jetstack.io"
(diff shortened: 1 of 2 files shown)


pkg/agent/config.go
```diff
@@ -177,7 +177,7 @@ func ParseConfig(data []byte, isVenafiCloudMode bool) (Config, error) {
 
 	if config.Server == "" && config.Endpoint.Host == "" && config.Endpoint.Path == "" {
 		config.Server = "https://preflight.jetstack.io"
```
nielshub, line 179: this should be the devstack url? defined in the agent yaml config?

nielshub (approved):

pkg/agent/config.go
tfadeyi: This URL is for the standalone backend. If you want to define a different URL for the devstack, you just need to add this at the top of your configuration.

```yaml
server: <devstack url>
```

PR #466, opened 2023-09-25
Chart compatible with TLSPK VCP auth

hawksight:
Creates a new chart to connect to the TLS Protect for Kuberentes Module which is part of the Venafi Control Plane (Or TLS Protect Cloud). Due to a new authentication method, it is simpler and easier to maintain a separate chart for this.

I have not completed everything docs wise here. I need more details about where we will put the chart, and how to authenticate. For now I have assumed the same as existing TLSPK but replaced "TLSPK" with "VENAFI". This may need a follow up one I know more.

PR #467, opened 2023-09-26
Allow to use ClientID and PrivateKeyPath as flags instead of credentials.json file

j-fuentes:
This will help simplify #466.

This changes make it possible to use `--client-id` to specify the client ID (and optionally `--private-key-path`) when using the Venafi Cloud authentication, allowing the user to not need to provide a `credentials.json` file.

This change should be backwards compatible.

### DIFF
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -88,7 +88,7 @@ func init() {
 		"client-id",
 		"",
 		"",
-		"Venafi Cloud Service Account client ID. If you use this flag you don't need to use --venafi-cloud as it will assume you are authentiating against Venafi Cloud. Using this removed the need to use a credentials file with Venafi Cloud mode.",
+		"Venafi Cloud Service Account client ID. If you use this flag you don't need to use --venafi-cloud as it will assume you are authenticating against Venafi Cloud. Using this removes the need to use a credentials file with Venafi Cloud mode.",
 	)
 	agentCmd.PersistentFlags().StringVarP(
 		&agent.PrivateKeyPath,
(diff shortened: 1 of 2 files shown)


hawksight (approved):
This looks like it adds the two options that means `credentials.json` would not be needed.
Can't speak too much for the code but looks good to me.

So with this addition I can use these flags where needed. Will the code base still want me to provide `credentials.json` though? Or will that be optional if the values are configured with these flags?

cmd/agent.go
```diff
@@ -81,7 +81,21 @@ func init() {
...
 		false,
-		"Runs agent with parsing config and credentials file in Venafi Cloud format if true.",
+		"Runs agent with parsing config (and credentials file if provided) in Venafi Cloud format if true.",
+	)
+	agentCmd.PersistentFlags().StringVarP(
+		&agent.ClientID,
+		"client-id",
+		"",
+		"",
+		"Venafi Cloud Service Account client ID. If you use this flag you don't need to use --venafi-cloud as it will assume you are authentiating against Venafi Cloud. Using this removed the need to use a credentials file with Venafi Cloud mode.",
```
tfadeyi, line 91:
```suggestion
		"Venafi Cloud Service Account client ID. If you use this flag you don't need to use --venafi-cloud as it will assume you are authenticating against Venafi Cloud. Using this removes the need to use a credentials file with Venafi Cloud mode.",
```

tfadeyi:
LGTM :+1:

@hawksight 

> Will the code base still want me to provide credentials.json though? Or will that be optional if the values are configured with these flags?

The `credentials.json` should be optional if the clientID and private key flags are passed

tfadeyi (approved):

PR #470, opened 2023-11-08
chore: Update agent version to latest

hawksight:
Using the latest `jetstack-agent` chart it showed as out of date by the platform.
image

This fixes that.


hawksight:
Need to update the chart version, don't merge just yet.

SpectralHiss (approved):

PR #502, opened 2024-01-24
chore: Switch default image to Venafi public registry

hawksight:
Now that Venafi has it's public registry with the `venafi-agent` image published, this naming makes more sense to align with this chart.

I'm leaving this in draft because now we have a version disparity between the chart version & the appVersion. I think Venafi product is keen to keep this in line, but this repository has no guidelines or process to ensure that.

Will discuss before making this ready to review.
For now we could bump some deps and release v0.1.44 of the container image and release this chart change in sync.
But I would rather we set a process for this.


achuchev:
@hawksight, I believe it's essential to maintain consistency across the image versions and Helm charts.

hawksight:
Thanks @achuchev for picking up the above. 

I'm happy to go either way. But keeping them in sync might require some release process setup. I suggest we take that offline to setup.

This PR was just example where versions could become skewed.

hawksight:
Closing this as it is outdated [and resolved](https://github.com/jetstack/jetstack-secure/blob/master/deploy/charts/venafi-kubernetes-agent/values.yaml#L55). 

cert-manager team own the code now and the version aligned with their best practices.

PR #508, opened 2024-02-05
feat(client): Allow agent to send cluster description

tfadeyi:
Allows the user to configure a cluster description to be used during the cluster registration in the Venafi control plane.

**Changes:**

* Add `cluster_description` to the agent configuration.
* Update the client interface and concrete structs to support `PostDataReadingsWithOptions`.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -10,9 +10,19 @@ import (
 )
 
 type (
+	// Options is the struct describing additional information pertinent to an agent that isn't a data reading
+	// These fields will then be uploaded together with data readings.
+	Options struct {
+		OrgID              string
+		ClusterID          string
+		ClusterName        string
+		ClusterDescription string
+	}
+
 	// The Client interface describes types that perform requests against the Jetstack Secure backend.
 	Client interface {
 		PostDataReadings(orgID, clusterID string, readings []*api.DataReading) error
+		PostDataReadingsWithOptions(readings []*api.DataReading, options Options) error
 		Post(path string, body io.Reader) (*http.Response, error)
 	}
 
--- a/pkg/client/client_venafi_cloud.go
+++ b/pkg/client/client_venafi_cloud.go
@@ -7,6 +7,7 @@ import (
 	"crypto/ed25519"
 	"crypto/rsa"
 	"crypto/x509"
+	"encoding/base64"
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
@@ -25,6 +26,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/hashicorp/go-multierror"
 	"github.com/jetstack/preflight/api"
+	"github.com/microcosm-cc/bluemonday"
 )
 
 type (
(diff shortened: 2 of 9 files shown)


dbarranco (approved):

pkg/client/client_venafi_cloud.go
```diff
@@ -152,6 +154,57 @@ func (c *VenafiSvcAccountCredentials) IsClientSet() bool {
...
+	if err != nil {
+		return err
+	}
+
+	// validate options and send them as URL params
+	query := venafiCloudUploadURL.Query()
+	stripHTML := bluemonday.StrictPolicy()
+	if opts.ClusterName != "" {
+		query.Add("name", stripHTML.Sanitize(opts.ClusterName))
+	}
```
maelvls, line 184: (Passing by) Just for future reference, I was wondering this: why do we pass the query param `name` set to the config's `cluster_id`? The `cluster_id` is already present in the request body of `/v1/tlspk/upload/clusterdata/:uploaderId`, more specifically under `agent_metadata`. For example:

```http
POST https://api.venafi.cloud/v1/tlspk/upload/clusterdata/no HTTP/2.0

authorization: Bearer ...
content-length: 536
accept-encoding: gzip
{
  "agent_metadata": {
    "version": "development",
    "cluster_id": "mvalais Kind Cluster"
  },
  "data_gather_time": "2024-07-26T13:03:39.831129Z",
  "data_readings": [
    {
      "cluster_id": "mvalais Kind Cluster",
      "data-gatherer": "k8s-discovery",
      "timestamp": "2024-07-26T15:03:09+02:00",
      "data": {
        "server_version": {
          "major": "1",
          "minor": "29",
          "gitVersion": "v1.29.2",
          "gitCommit": "4b8e819355d791d96b7e9d9efe4cbafae2311c88",
          "gitTreeState": "clean",
          "buildDate": "2024-02-14T22:25:42Z",
          "goVersion": "go1.21.7",
          "compiler": "gc",
          "platform": "linux/arm64"
        }
      },
      "schema_version": "v2.0.0"
    }
  ]
}
```

When I don't pass the query param `name=mvalais+Kind+Cluster`, I get the following error:

```json
[{"message":"could not find a Kubernetes cluster linked to the client in the Venafi Control Plane"}
```

pkg/client/client.go
```diff
@@ -10,9 +10,19 @@ import (
 )
 
 type (
+	// Options is the struct describing additional information pertinent to an agent that isn't a data reading
+	// These fields will then be uploaded together with data readings.
+	Options struct {
+		OrgID              string
+		ClusterID          string
+		ClusterName        string
```
maelvls, line 18: What's the difference between ClusterName and ClusterID? It seems like ClusterName is set to ClusterID in `run.go`:

```go
err := preflightClient.PostDataReadingsWithOptions(readings, client.Options{
	ClusterName:        config.ClusterID,
	ClusterDescription: config.ClusterDescription,
})
```

and ClusterID isn't set?

maelvls: Looking up earlier in the file run.go:

```go
// orgID and clusterID are not required for Venafi Cloud auth
```

Sounds like we could comment the `Options` struct like this:

```go
	Options struct {
		OrgID              string // Only used with platform.jetstack.io
		ClusterID          string // Only used with platform.jetstack.io
		ClusterName        string // Only used with Venafi Cloud
		ClusterDescription string // Only used with Venafi Cloud
	}
```

pkg/client/client_venafi_cloud.go
tfadeyi: The payload body could be quite large, several MB depending on the cluster where the agent is installed, in the backend entrypoint we wanted to be able to get this information without having to parse the entire upload before the actual processing pipeline kicks in. (Note: we could stream the upload JSON but we thought query params would work well instead).

The error can be improved, but essentially you have 2 options to onboard a cluster:
* venctl installed which creates a service account and cluster_connection (relation of cluster and service account) in VCP.
* Helm installation which requires the backend entrypoint to create a cluster_connection in VCP for your cluster and service account.

The error is saying that the cluster and service account are not linked in VCP.

In the venafi docs we tell users following the helm installation that they are **required** to set the cluster name, which will also be present in the UI. We could introduce an `auto-onboard` field to the config but since the cluster name field is required for the helm installation we thought we could use it rather than introducing another field.

If the cluster name is set in the config it will be used in the query param. The user wouldn't interact with the query params.

pkg/client/client_venafi_cloud.go
```diff
@@ -152,6 +154,57 @@ func (c *VenafiSvcAccountCredentials) IsClientSet() bool {
...
+
+	venafiCloudUploadURL, err := url.Parse(filepath.Join(c.uploadPath, c.uploaderID))
+	if err != nil {
+		return err
+	}
+
+	// validate options and send them as URL params
+	query := venafiCloudUploadURL.Query()
+	stripHTML := bluemonday.StrictPolicy()
```
maelvls, line 181: (Passing by) Why is `stripHTML.Sanitize` needed? `query.Encode` already URL-encodes values.

pkg/client/client_venafi_cloud.go
maelvls, line 184: (Passing by) Why was `stripHTML.Sanitize` needed here? Isn't overkill in this context?  `query.Encode` already URL-encodes values, and this value isn't meant to be interpreted by any HTML parsers

maelvls: Thanks, that a super useful insight!

pkg/client/client_venafi_cloud.go
```diff
@@ -152,6 +154,57 @@ func (c *VenafiSvcAccountCredentials) IsClientSet() bool {
...
+	if opts.ClusterName != "" {
+		query.Add("name", stripHTML.Sanitize(opts.ClusterName))
+	}
+	if opts.ClusterDescription != "" {
+		query.Add("description", base64.RawURLEncoding.EncodeToString([]byte(stripHTML.Sanitize(opts.ClusterDescription))))
+	}
```
maelvls, line 187: Why base 64 encode this? Query parameters are already URL-encoded by query.Encode, which should be enough to pass that data to the server.

Same question as above regarding `stripHTML.Sanitize` too

PR #510, opened 2024-02-05
feat(onboarding): Add onboarding values to the helm chart

tfadeyi:
Adds cluster onboarding values to the agent helm chart.

Change:
* Add onboarding object to the values.yaml
* Add onboarding.clusterName which updates the agent configMap template 
* Add onboarding.clusterDescription which updates the agent configMap template
* Update Autogen docs

**Note**: Wait until v0.1.44 agent is released before merging

### DIFF
--- a/deploy/charts/venafi-kubernetes-agent/values.yaml
+++ b/deploy/charts/venafi-kubernetes-agent/values.yaml
@@ -91,6 +91,10 @@ config:
   clientId: ""
   # -- Send data back to the platform every minute unless changed
   period: "0h1m0s"
+  # -- Name for the cluster resource if it needs to be created in Venafi Control Plane
+  clusterName: ""
+  # -- Description for the cluster resource if it needs to be created in Venafi Control Plane
+  clusterDescription: ""
 
   # -- Specify ConfigMap details to load config from an existing resource.
   # This should be blank by default unless you have you own config.
--- a/deploy/charts/venafi-kubernetes-agent/Chart.yaml
+++ b/deploy/charts/venafi-kubernetes-agent/Chart.yaml
@@ -3,5 +3,5 @@ name: venafi-kubernetes-agent
 description: |-
   The Venafi Kubernetes Agent connects your Kubernetes or Openshift cluster to the Venafi Control Plane.
 type: application
-version: 0.1.43
-appVersion: "v0.1.43"
+version: 0.1.44
+appVersion: "v0.1.44"
--- a/deploy/charts/venafi-kubernetes-agent/README.md
+++ b/deploy/charts/venafi-kubernetes-agent/README.md
@@ -2,7 +2,7 @@
 
 The Venafi Kubernetes Agent connects your Kubernetes or Openshift cluster to the Venafi Control Plane.
 
-![Version: 0.1.43](https://img.shields.io/badge/Version-0.1.43-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.1.43](https://img.shields.io/badge/AppVersion-v0.1.43-informational?style=flat-square)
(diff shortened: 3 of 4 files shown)


hawksight:
@tfadeyi I had a couple of q's for context:
- What is the significance of "onboarding"? 
- Could these keys just be under `config.clusertName` & `config.clusterDescription`?
- Are the two fields `clusterName` & `clusterDescription` optional?

dbarranco (approved):

achuchev:
@hawksight All good points. Let me know if you want to understand more about the proposed change.

@tfadeyi I agree with Peter that the new options have to be under `config.`

deploy/charts/venafi-kubernetes-agent/values.yaml
```diff
@@ -91,6 +91,10 @@ config:
   clientId: ""
   # -- Send data back to the platform every minute unless changed
   period: "0h1m0s"
+  # -- Venafi Control Plane cluster name used for onboarding information.
```
achuchev, line 94:
```suggestion
  # -- Name for the cluster resource if it needs to be created in Venafi Control Plane.
```

deploy/charts/venafi-kubernetes-agent/values.yaml
```diff
@@ -91,6 +91,10 @@ config:
...
+  clusterName: ""
+  # -- Venafi Control Plane cluster description used for onboarding information.
```
achuchev, line 96:
```suggestion
  # -- Description for the cluster resource if it needs to be created in Venafi Control Plane.
```

hawksight:
@tfadeyi thanks for the explanation. What would happen if somewhere were to change the `clusterName` and `clusterDescription` fields? They would just change on the VCP?

I was thinking that if they are a "set once and never change" type field, then it makes sense to consolidate them under their own heading `onboarding`. But if they are just optional and can be changed on an ongoing basis, then I think it makes more sense to have them sit alongside the other configurable variables:

```shell
helm show values oci://$VENAFI_REGISTRY_PRIVATE_US/charts/venafi-kubernetes-agent | yq .
```

```yaml
config:
  clientId: "XXXXXXXXXXXXXXX"
  clusterName: "black-pearl"
  clusterDescription: "The fastest cluster on the seven seas"
```

It also feels a little more natural and consistent with the [jetstack-agent config](https://github.com/jetstack/jetstack-secure/blob/master/deploy/charts/jetstack-agent/values.yaml#L80-L87).

Also thinking about the user install experience, it just makes the command a tiny bit shorter:

```shell
# This
helm diff upgrade --install venafi-kubernetes-agent oci://$VENAFI_REGISTRY_PRIVATE_US/charts/venafi-kubernetes-agent --namespace $VENAFI_NAMESPACE --create-namespace --set config.clientId="$VENAFI_CLIENT_ID" --set config.clusterName="abcdefg" --set config.clusterDescription="Alphabet" --version "0.1.43" --registry-config "${VENAFI_CREDS_LOCAL_PATH}/${VENAFI_CRED_NAME}/config.json"
# As opposed to this:
helm diff upgrade --install venafi-kubernetes-agent oci://$VENAFI_REGISTRY_PRIVATE_US/charts/venafi-kubernetes-agent --namespace $VENAFI_NAMESPACE --create-namespace --set config.clientId="$VENAFI_CLIENT_ID" --set config.onboarding.clusterName="abcdefg" --set config.onboarding.clusterDescription="Alphabet" --version "0.1.43" --registry-config "${VENAFI_CREDS_LOCAL_PATH}/${VENAFI_CRED_NAME}/config.json"
```

### Other note

Perhaps we should remove the chart bump from here and [here](https://github.com/jetstack/jetstack-secure/pull/502) and do that explicitly as it's own commit before creating a tag / release on this repo?

achuchev:
@hawksight: The cluster name and description will be preserved in VCP even if the agent reports new values. This ensures consistency and avoids unintended updates.

tfadeyi:
> Perhaps we should remove the chart bump from https://github.com/jetstack/jetstack-secure/pull/502 and here and do that explicitly as it's own commit before creating a tag / release on this repo?

I've already made a new release, didn't see the message in time. Looking at the other issue, I think we can come up with a more defined process for it.

PR #525, opened 2024-03-22
fix/vc-31703-agent-memory-startup-spikes

mol-george:
This attempts to fix the issue in [VC-31703](https://venafi.atlassian.net/browse/VC-31703).

The initial data gatherer sync operation is currently executed async which might cause in an unreasonable amount of memory being used by the agent pod which results in the pod being OOM Killed.

Executing this operation one data gatherer at the time might mitigate this. Extending the time took for the initial sync is not a concern at this point.

[VC-31703]: https://venafi.atlassian.net/browse/VC-31703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -18,14 +18,15 @@ import (
 
 	"github.com/cenkalti/backoff"
 	"github.com/hashicorp/go-multierror"
-	"github.com/jetstack/preflight/api"
-	"github.com/jetstack/preflight/pkg/client"
-	"github.com/jetstack/preflight/pkg/datagatherer"
-	"github.com/jetstack/preflight/pkg/version"
 	json "github.com/json-iterator/go"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/cobra"
+
+	"github.com/jetstack/preflight/api"
+	"github.com/jetstack/preflight/pkg/client"
+	"github.com/jetstack/preflight/pkg/datagatherer"
+	"github.com/jetstack/preflight/pkg/version"
 )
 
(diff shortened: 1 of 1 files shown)


tfadeyi (approved):

maelvls:
Passing by:

- The memory spike certainly comes from the initial "list" that client-go caches have to do. Before this PR, the initial "list" of the 26 resource types was being done all at once. After this PR, the initial "list" is being performed one after the other.
- The reason is most certainly because the initial "list" needs to return the entire list of objects [reflector.go#L396](https://github.com/kubernetes/client-go/blob/04f67d5d4f902fde74dab51fe19698e80f2ddeb3/tools/cache/reflector.go#L396) before they can be fed to the cache. The server-side pagination ("chunking") has always been turned off on the APi server due to kube-apiserver performance reasons:
  > Request with pagination, due to performance reasons, translates to a full (pagination is ignored) LIST from the watch cache.
- Using the standard list-and-then-watch mechanism, there is no way to process the items as they come from the API server (i.e., put them into the cache) due to the pagination being turned off.
- One solution would be to move to WatchList: [3157-watch-list](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/3157-watch-list/README.md#appendix).
- WatchList was introduced in Alpha in 1.27, moved to Beta in 1.31, and maybe will become GA in Kubernetes 1.33 or 1.34
- WatchList status: https://github.com/kubernetes/enhancements/issues/3157
- Kudos to @wallrj for the extensive research:
  - https://github.com/cert-manager/cert-manager/pull/7161
  - https://github.com/cert-manager/cert-manager/pull/7175

PR #537, opened 2024-05-14
fix(log): Some improvements to the logs

james-w:
The current logs are somewhat confusing, and missing some very useful information. This is an attempt
to clean it up in a few ways:

  *  Log the number of items from each dg. Each datagatherer can return 
     the number of items it collected so that the logs tell us a bit more about
     what was found in the cluster, and can help find where any items have
     been missed.

  * Log a more informative error message when giving up on uploading
    readings to the server. This would be the last message before
    the pod exits, so if the pod ends up in CrashLoopBackoff it's
    important to highlight this as the reason.

  * Remove the "Running Agent" log message, it's not clear what it means

  * Put [] around the body as if the body is empty the log message
    can read like the following message is the body.

```
2024/05/10 16:33:43 retrying in 25.555756126s after error: received response with status code 404. Body:
W0510 16:33:43.832278   10875 reflector.go:535] pkg/mod/k8s.io/client-go@v0.28.3/tools/cache/reflector.go:229: failed to list route.openshift.io/v1, Resource=routes: the server could not find the requested resource
```

  * Remove "using parent stop channel" from a log message as it's not
    clear what it means. It used to be conditional, and now it isn't,
    so it's not providing any information, and it's confusing as to
    what it might be trying to tell you if you aren't familiar with that
    code.

 There are couple of code cleanups included as well:

  *  Fix a crash bug when REST config can't be loaded.
     `discoveryClient` is uninitialized, so the pointer can't be deferenced,
      so return an empty struct if there's an error.

  *  Remove the informer context in the dynamic dg. The informer
     context is created, but not passed to anything, so this code doesn't do anything.

The changes are split into a few logical commits if we want to break it up and review them independently,
as I know these might not be the best approaches for all of these things.

### DIFF
--- a/pkg/agent/dummy_data_gatherer.go
+++ b/pkg/agent/dummy_data_gatherer.go
@@ -44,7 +44,7 @@ func (g *dummyDataGatherer) Delete() error {
 	return nil
 }
 
-func (c *dummyDataGatherer) Fetch() (interface{}, error) {
+func (c *dummyDataGatherer) Fetch() (interface{}, int, error) {
 	var err error
 	if c.attemptNumber < c.FailedAttempts {
 		err = fmt.Errorf("First %d attempts will fail", c.FailedAttempts)
@@ -53,5 +53,5 @@ func (c *dummyDataGatherer) Fetch() (interface{}, error) {
 		err = fmt.Errorf("This data gatherer will always fail")
 	}
 	c.attemptNumber++
-	return nil, err
+	return nil, -1, err
 }
--- a/pkg/client/client_venafi_cloud.go
+++ b/pkg/client/client_venafi_cloud.go
@@ -199,7 +199,7 @@ func (c *VenafiCloudClient) PostDataReadingsWithOptions(readings []*api.DataRead
 		if err == nil {
 			errorContent = string(body)
 		}
-		return fmt.Errorf("received response with status code %d. Body: %s", code, errorContent)
+		return fmt.Errorf("received response with status code %d. Body: [%s]", code, errorContent)
 	}
 
 	return nil
@@ -235,7 +235,7 @@ func (c *VenafiCloudClient) PostDataReadings(_ string, _ string, readings []*api
 		if err == nil {
 			errorContent = string(body)
 		}
-		return fmt.Errorf("received response with status code %d. Body: %s", code, errorContent)
+		return fmt.Errorf("received response with status code %d. Body: [%s]", code, errorContent)
 	}
 
 	return nil
@@ -327,7 +327,7 @@ func (c *VenafiCloudClient) sendHTTPRequest(request *http.Request, responseObjec
 
 	if response.StatusCode != http.StatusOK && response.StatusCode != http.StatusCreated {
 		body, _ := io.ReadAll(response.Body)
-		return fmt.Errorf("failed to execute http request to VaaS. Request %s, status code: %d, body: %s", request.URL, response.StatusCode, body)
+		return fmt.Errorf("failed to execute http request to Venafi Control Plane. Request %s, status code: %d, body: [%s]", request.URL, response.StatusCode, body)
 	}
 
 	body, err := io.ReadAll(response.Body)
--- a/pkg/datagatherer/k8s/dynamic.go
+++ b/pkg/datagatherer/k8s/dynamic.go
@@ -302,7 +302,7 @@ func (g *DataGathererDynamic) Run(stopCh <-chan struct{}) error {
 // before collecting the resources.
 func (g *DataGathererDynamic) WaitForCacheSync(stopCh <-chan struct{}) error {
 	if !k8scache.WaitForCacheSync(stopCh, g.informer.HasSynced) {
-		return fmt.Errorf("timed out waiting for caches to sync, using parent stop channel")
+		return fmt.Errorf("timed out waiting for Kubernetes caches to sync")
 	}
 
 	return nil
--- a/pkg/datagatherer/k8s/dynamic.go
+++ b/pkg/datagatherer/k8s/dynamic.go
@@ -307,9 +307,9 @@ func (g *DataGathererDynamic) Delete() error {
 
 // Fetch will fetch the requested data from the apiserver, or return an error
 // if fetching the data fails.
-func (g *DataGathererDynamic) Fetch() (interface{}, error) {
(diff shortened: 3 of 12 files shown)


maelvls:
Hey, I've taken a look at your PR. I've went ahead with reviewing this as part of the [handover effort](https://venafi.atlassian.net/wiki/spaces/TPFK/pages/2409103391/Venafi+Agent+Project+Handover) started this morning. I'll let Olu do a final review as I don't know much about this code base.


> Log a more informative error message **when giving up on uploading readings** to the server. This would be the last message before the pod exits, so if the pod ends up in CrashLoopBackoff it's important to highlight this as the reason.

I wasn't able to reproduce the agent giving up, or I haven't waited long enough, I guess the backoff threshold is high:

```console
$ export HTTPS_PROXY=foo
$ go run . agent -c config.yaml --client-id XXXXX -k /tmp/key --venafi-cloud -p 5s 2>&1 | grep api.venafi.cloud
2024/05/14 14:54:23 Posting data to: https://api.venafi.cloud/
2024/05/14 14:54:23 retrying in 17.654784782s after error: post to server failed: Post "https://api.venafi.cloud/v1/oauth/token/serviceaccount": proxyconnect tcp: dial tcp: lookup foo: no such host
2024/05/14 14:54:41 Posting data to: https://api.venafi.cloud/
2024/05/14 14:54:41 retrying in 56.057860693s after error: post to server failed: Post "https://api.venafi.cloud/v1/oauth/token/serviceaccount": proxyconnect tcp: dial tcp: lookup foo: no such host
2024/05/14 14:55:37 Posting data to: https://api.venafi.cloud/
2024/05/14 14:55:37 retrying in 42.620269206s after error: post to server failed: Post "https://api.venafi.cloud/v1/oauth/token/serviceaccount": proxyconnect tcp: dial tcp: lookup foo: no such host
2024/05/14 14:56:19 Posting data to: https://api.venafi.cloud/
2024/05/14 14:56:19 retrying in 51.204385623s after error: post to server failed: Post "https://api.venafi.cloud/v1/oauth/token/serviceaccount": proxyconnect tcp: dial tcp: lookup foo: no such host
2024/05/14 14:57:11 Posting data to: https://api.venafi.cloud/
2024/05/14 14:57:11 retrying in 2m16.713596561s after error: post to server failed: Post "https://api.venafi.cloud/v1/oauth/token/serviceaccount": proxyconnect tcp: dial tcp: lookup foo: no such host
```

(I've hidden all the useless messages about missing resources)

About the REST config bug, well spotted. I wasn't able to reproduce the panic (if that's how you spotted this)

Anyways, I think all the changes you made are sensible.

pkg/client/client_venafi_cloud.go
```diff
@@ -327,7 +327,7 @@ func (c *VenafiCloudClient) sendHTTPRequest(request *http.Request, responseObjec
 
 	if response.StatusCode != http.StatusOK && response.StatusCode != http.StatusCreated {
 		body, _ := io.ReadAll(response.Body)
-		return fmt.Errorf("failed to execute http request to VaaS. Request %s, status code: %d, body: %s", request.URL, response.StatusCode, body)
+		return fmt.Errorf("failed to execute http request to VaaS. Request %s, status code: %d, body: [%s]", request.URL, response.StatusCode, body)
```
dbarranco, line 330:
```suggestion
		return fmt.Errorf("failed to execute http request to Venafi Control Plane. Request %s, status code: %d, body: [%s]", request.URL, response.StatusCode, body)
```

dbarranco: I think nowadays we're trying to use more the VCP / Venafi Control Plane naming instead of VaaS

pkg/datagatherer/k8s/dynamic.go
```diff
@@ -302,7 +292,7 @@ func (g *DataGathererDynamic) Run(stopCh <-chan struct{}) error {
 // before collecting the resources.
 func (g *DataGathererDynamic) WaitForCacheSync(stopCh <-chan struct{}) error {
 	if !k8scache.WaitForCacheSync(stopCh, g.informer.HasSynced) {
-		return fmt.Errorf("timed out waiting for caches to sync, using parent stop channel")
+		return fmt.Errorf("timed out waiting for caches to sync")
```
dbarranco, line 295:
```suggestion
		return fmt.Errorf("timed out waiting for Kubernetes caches to sync")
```

dbarranco: Maybe it'd be better to indicate which kind of cache? Just so people don't get confused with other caches

dbarranco (approved):
Looks good to me, just added a couple of minor suggestions.

We should think about structuring these logs and converting them into events soon 🤔

james-w:
> Hey, I've taken a look at your PR. I've went ahead with reviewing this as part of the [handover effort](https://venafi.atlassian.net/wiki/spaces/TPFK/pages/2409103391/Venafi+Agent+Project+Handover) started this morning. I'll let Olu do a final review as I don't know much about this code base.

Thanks!

> > Log a more informative error message **when giving up on uploading readings** to the server. This would be the last message before the pod exits, so if the pod ends up in CrashLoopBackoff it's important to highlight this as the reason.
> 
> I wasn't able to reproduce the agent giving up, or I haven't waited long enough, I guess the backoff threshold is high:

I believe the default timeout is 15m before it gives up.

> About the REST config bug, well spotted. I wasn't able to reproduce the panic (if that's how you spotted this)

It was. Running locally without a kubeconfig file should reproduce.

> Anyways, I think all the changes you made are sensible.

Thanks.

pkg/agent/dummy_data_gatherer.go
```diff
@@ -53,5 +53,5 @@ func (c *dummyDataGatherer) Fetch() (interface{}, error) {
 		err = fmt.Errorf("This data gatherer will always fail")
 	}
 	c.attemptNumber++
-	return nil, err
+	return nil, 0, err
```
tfadeyi, line 56: Nitpick: could we make this `-1` ?

tfadeyi (approved):
LGTM :+1:

james-w:
Thanks all, I pushed an update with the suggested changes.

tfadeyi:
Thank you :+1: LGTM, I'll merge the changes to master

PR #539, opened 2024-05-29
[VC-32659] Remove the default CPU limit

wallrj:
In https://docs.venafi.cloud/vaas/k8s-components/c-k8s-components-best-practice/#cpu we say:
> Read [Stop Using CPU Limits on Kubernetes](https://home.robusta.dev/blog/stop-using-cpu-limits), to learn why the best practice is to set CPU requests, but not limits.

So I've removed the default CPU limit from the chart values for venafi-kubernetes-agent.

I have not changed the `jetstack-agent` charts, because I think it is now considered legacy.

I also removed the note about using vertical pod autoscaler, because it seems irrelevant. We don't mention VPA in the documentation or in the other charts AFAIK.


hawksight (approved):
Having read through I think I agree.
Ultimately the choice is down to the platform team and their policy around what to set and not set.

The bonus here is that we are less likely to have a new install in a large cluster killed for resource starvation (of CPU) and require debugging.

maelvls:
@tfadeyi Hey, I'm not sure whether this needs to be reviewed by you, WDYT?

maelvls (approved):
Thanks for making this change, it's a good idea to not set CPU limits by default... especially because none of the "pros" for using CPU limits apply to the Venafi Agent: avoiding reliance on unpredictable spare cpu cycles due to low cpu requests; getting more predictable QoS; being able to charge back on multi-tenant environment.

tfadeyi:
LGTM

I don't think it requires my approval, I think @wallrj should have permission to merge

PR #540, opened 2024-06-20
feat: enable volumes and volumeMounts to be passed to the jetstack-agent deployment

hawksight:
I was working with a customer and needed to pass a configmap mount to the deployment.
This is not currently available in the chart.
This PR adds this capability in a generic way, similar to how it is done in the [cert-manager chart](https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/values.yaml#L317-L321).

I have specifically not created a new release of the chart. I intend to do a simple followup PR to release 0.3.2.

### DIFF
--- a/deploy/charts/jetstack-agent/README.md
+++ b/deploy/charts/jetstack-agent/README.md
@@ -133,12 +133,13 @@ kubectl logs -n jetstack-secure $(kubectl get pod -n jetstack-secure -l app.kube
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | affinity | object | `{}` |  |
+| authentication | object | `{"createSecret":false,"secretKey":"credentials.json","secretName":"agent-credentials","secretValue":"","type":"file"}` | Authentication section for the agent |
 | authentication.createSecret | bool | `false` | Reccomend that you do not use this and instead creat the credential secret outside of helm |
 | authentication.secretKey | string | `"credentials.json"` | Key name in secret |
 | authentication.secretName | string | `"agent-credentials"` | Name of the secret containing agent credentials.json |
 | authentication.secretValue | string | `""` | Base64 encoded value from Jetstack Secure Dashboard - only required when createSecret is true |
 | authentication.type | string | `"file"` | Type can be "file"/"token" determining how the agent should authenticate the to the backend |
-| command | list | `[]` |  |
+| command | list | `[]` | Override the jetstack-agent entrypoint with command list |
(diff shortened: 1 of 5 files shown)


deploy/charts/jetstack-agent/README.md
```diff
@@ -151,7 +152,7 @@ kubectl logs -n jetstack-secure $(kubectl get pod -n jetstack-secure -l app.kube
 | config.override.enabled | bool | `false` | Override disabled by default |
 | config.period | string | `"0h1m0s"` | Send data back to the platform every minute unless changed |
 | config.server | string | `"https://platform.jetstack.io"` | Overrides the server if using a proxy between agent and Jetstack Secure |
-| extraArgs | list | `[]` |  |
+| extraArgs | list | `[]` | Add additional arguments to the  |
```
aidy, line 155: Assume you were eaten by a grue

deploy/charts/jetstack-agent/tests/deployment_test.yaml
```diff
@@ -90,3 +90,27 @@ tests:
       - contains:
           path: spec.template.spec.containers[0].command
           content: notpreflight
+
+  # Check the volumes and volumeMounts works correctly
+  - it: Volumes and VolumenMounts added correctly
```
aidy, line 95:
```suggestion
  - it: Volumes and VolumeMounts added correctly
```

deploy/charts/jetstack-agent/README.md
hawksight: Yep, brain was working faster than my fingers and forgot to finish that.

hawksight:
@aidy thank you, have corrected those now. Can you cast one more eye over the changes please?

aidy (approved):
lgtm, although we should probably also bring these changes into the venafi agent chart.

hawksight:
> should probably also bring these changes into the venafi agent chart

Agreed, but I don't have a CTA for that, and it doesn't help the customer problem I have. So I'll drop that as an issue for now. #541

deploy/charts/jetstack-agent/tests/values/custom-volumes.yaml
```diff
@@ -0,0 +1,11 @@
+volumes:
+  - name: cabundle
+    configmap:
+      name: cabundle
+      optional: true
```
maelvls, line 5: Why is this optional? Isn't this prone to silent unexpected situations such as you forget to create the CA bundle configmap and the pod somehow isn't able to connect?

PR #542, opened 2024-06-20
chore: Upgrade chart version to 0.4.0 for jetstack-agent

hawksight:
Bump minor version to release the chart change to be able to add Volumes & VolumeMounts from #540.

### DIFF
--- a/deploy/charts/jetstack-agent/Chart.yaml
+++ b/deploy/charts/jetstack-agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: jetstack-agent
 description: TLS Protect for Kubernetes Agent
 type: application
-version: 0.3.1
+version: 0.4.0
 appVersion: "v0.1.43"
 home: https://github.com/jetstack/jetstack-secure
 maintainers:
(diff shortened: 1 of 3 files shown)


deploy/charts/jetstack-agent/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: jetstack-agent
 description: TLS Protect for Kubernetes Agent
 type: application
-version: 0.3.1
+version: 0.3.2
```
aidy, line 5: This is bumping the hotfix/patch version, not the minor version.

imo, this should be a minor version (0.4.0), but if you've already cut a 0.3.2 release, or if you really want it to be 0.3.2, I'm not feeling particularly pedantic today.

hawksight: I'm easy. I have not cut a release (create from master manually after merging).
Minor encompases new features correct?
Patch is just fixing existing features?

Happy to change to keep this right, just checking it is right.

aidy (approved):

PR #551, opened 2024-07-16
go.mod: update the Go dependencies so that we can use venafi-connection

maelvls:
The venafi-connection-lib library uses more recent versions of the dependencies in the Venafi Kubernetes Agent. When I ran

    export GOPRIVATE=github.com/jetstack
    go get github.com/jetstack/venafi-connection-lib@latest

most of the dependencies in this project got updated. This commit captures these changes.


maelvls:
Closing in favor of #552.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/jetstack/preflight
 
-go 1.21
+go 1.22.0
+
+toolchain go1.22.3
```
maelvls, line 5: This is wrong. This indicates that I was using a `go.work`.

PR #553, opened 2024-07-31
fix: Set master build pipeline to no use experimental cosign with env var

### LINKED ISSUE #545: CI: "Generate provenance" fails
The following step in the workflow always seems to fail for quite a while: https://github.com/jetstack/jetstack-secure/blob/master/.github/workflows/release-master.yml#L95-L100

We don't see it because it only runs on master branch. Given no-one has noticed or reported I take it is not a huge issue.
But perhaps we should check why it is failing and look to fix?

hawksight:
Fixes #545

### DIFF
--- a/.github/workflows/release-master.yml
+++ b/.github/workflows/release-master.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - build-fix-provenance
     tags:
       - v*
 jobs:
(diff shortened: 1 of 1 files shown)


maelvls:
I propose to close this: we have re-enabled cosign in #556 thanks to Makefile Modules.

Feel free to re-open if there is still work to be done.

PR #555, opened 2024-08-12
Makefile Modules: Change nothing except for Makefile moved to `make/02_mod.mk`

maelvls:
Makefile Modules are too tempting, I want to tap into them, especially the tools module.

As a first "baby" step, I propose to move `Makefile` to `make/02_mod.mk`: nothing else changes. The GitHub Actions are the same, the old targets in `Makefile` work the same.

My idea is to migrate more and more to native makefile-modules targets without having to migrate the whole thing at once.


maelvls:
Ooof that's a lot of work to fix `make build-docker-image`. The makefile itself runs within the container, it's just super complex for not many reasons (buildx, qemu, etc).

Closing for now.

inteon:
@maelvls the PR looks good from a Makefile modules perspective. Let's keep it as reference for when we have a bit more time to figure out these buildx, qemu etc. tools.

PR #563, opened 2024-08-23
VC-35630: Add unit tests to the code that loads config/flags

maelvls:
Stacked on top of https://github.com/jetstack/jetstack-secure/pull/564.

Ref: [VC-35630](https://venafi.atlassian.net/browse/VC-35630)

I found that almost none of the "config/flags" code is tested, which made me wary of changing the code without adding more tests. And since I'll have to change this code again for [VC-35630](https://venafi.atlassian.net/browse/VC-35630), I'd prefer writing some unit tests first.

This PR aims to change no actual logic. The changes you see are the refactoring I had to do to be able to unit test how configuration is loaded.

[VC-35630]: https://venafi.atlassian.net/browse/VC-35630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[VC-35630]: https://venafi.atlassian.net/browse/VC-35630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -3,8 +3,6 @@ package cmd
 import (
 	"fmt"
 	"io/ioutil"
-	"os"
-	"time"
 
 	"github.com/jetstack/preflight/pkg/agent"
 	"github.com/jetstack/preflight/pkg/logs"
@@ -37,16 +35,16 @@ var agentRBACCmd = &cobra.Command{
 	Long:  `Print RBAC string by reading GVRs`,
 	Run: func(cmd *cobra.Command, args []string) {
 
-		b, err := ioutil.ReadFile(agent.ConfigFilePath)
+		b, err := ioutil.ReadFile(agent.Flags.ConfigFilePath)
 		if err != nil {
 			logs.Log.Fatalf("Failed to read config file: %s", err)
 		}
-		config, err := agent.ParseConfig(b, false)
+		cfg, err := agent.ParseConfig(b, false)
 		if err != nil {
 			logs.Log.Fatalf("Failed to parse config file: %s", err)
 		}
 
-		out := permissions.GenerateFullManifest(config.DataGatherers)
+		out := permissions.GenerateFullManifest(cfg.DataGatherers)
 		fmt.Print(out)
 	},
 }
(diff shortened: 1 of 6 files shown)


maelvls:
I've just realized that this PR is impossible to review due to the lots of files I moved and then changed. I'm not sure how to make the review simpler 😕

cmd/agent.go
```diff
@@ -55,119 +54,5 @@ func init() {
...
-		"Enables the pprof profiling server on the agent (port: 6060).",
-	)
-	agentCmd.PersistentFlags().BoolVarP(
-		&agent.Prometheus,
-		"enable-metrics",
-		"",
-		false,
-		"Enables Prometheus metrics server on the agent (port: 8081).",
-	)
+	config.InitAgentCmdFlags(agentCmd, &agent.AgentCmdFlags)
```
maelvls, line 57: **Self-review:** I've moved the definition of the flags for the `agent` command to the `config` package so that the struct and the flags they relate to are defined in the same file. I find it easier to keep things consistent this way.

pkg/agent/config_test.go
maelvls, line 1: **Self-review:** I chose to move `config.go` and `config_test.go` to a separate `config/` folder because the `agent` package is super busy. I realize that it's not a great reason and moving things around makes this PR super hard to read... I'm OK reverting this if you ask me

maelvls: I've moved `config.go` and `config_test.go` back to the `agent` package. It's now much simpler to review! I'll click "Resolve" now.

cmd/agent.go
```diff
@@ -55,119 +53,5 @@ func init() {
...
-		"Enables the pprof profiling server on the agent (port: 6060).",
-	)
-	agentCmd.PersistentFlags().BoolVarP(
-		&agent.Prometheus,
-		"enable-metrics",
-		"",
-		false,
-		"Enables Prometheus metrics server on the agent (port: 8081).",
-	)
+	agent.InitAgentCmdFlags(agentCmd, &agent.Flags)
```
maelvls, line 56: **Self-review:** I've moved the definition of the flags for the `agent` command to the `config` package so that the struct and the flags they relate to are defined in the same file. I find it easier to keep things consistent this way.

pkg/agent/config.go
```diff
@@ -57,6 +68,354 @@ type VenafiCloudConfig struct {
 	UploadPath string `yaml:"upload_path,omitempty"`
 }
 
+type AgentCmdFlags struct {
```
maelvls, line 71: **Self-review:** these were defined as package-level variables, so I moved them to be a structs so that I can test `getConfiguration`.

pkg/agent/run.go
```diff
@@ -26,70 +24,11 @@ import (
...
-// May be left empty when running in Kubernetes. In this case, the namespace is
-// read from the file /var/run/secrets/kubernetes.io/serviceaccount/namespace.
-var InstallNS string
-
-// Profiling flag enabled pprof endpoints to run on the agent
-var Profiling bool
-
-// Prometheus flag enabled Prometheus metrics endpoint to run on the agent
-var Prometheus bool
+var Flags AgentCmdFlags
```
maelvls, line 31: **Self-review:** This variable is public so that the `cmd/agent.go` file can import the global var `Flags` to register the command line flags.

pkg/agent/config.go
```diff
@@ -57,6 +68,354 @@ type VenafiCloudConfig struct {
...
+		"",
+		false,
+		"Enables Prometheus metrics server on the agent (port: 8081).",
+	)
+}
+
+// getConfiguration combines the input configuration with the flags passed to
+// the agent and returns the final configuration as well as the Venafi client to
+// be used to upload data.
+func getConfiguration(log *log.Logger, cfg Config, flags AgentCmdFlags) (Config, client.Client, error) {
```
maelvls, line 263: **Self-review:** I moved `getConfiguration` over from `run.go` as-is without any modification. This is because I think this belongs to `config.go`, not `run.go`. But let me know if it makes things too difficult to review.

pkg/agent/config_test.go
```diff
@@ -1,15 +1,202 @@
...
+	"github.com/jetstack/preflight/pkg/client"
 	"github.com/kylelemons/godebug/diff"
-	"gopkg.in/d4l3k/messagediff.v1"
+	"github.com/stretchr/testify/assert"
 )
 
+func TestGetConfiguration(t *testing.T) {
+	t.Run("minimal successful configuration", func(t *testing.T) {
+		got, cl, err := getConfiguration(discardLogs(t),
+			Config{Server: "http://localhost:8080", Period: 1 * time.Hour},
```
maelvls, line 22: **Self-review:** This URL isn't actually used, I guess I should have picked a clearer URL like `https://fake`.

pkg/agent/config_test.go
```diff
@@ -1,15 +1,202 @@
...
+	"bytes"
 	"fmt"
+	"io"
+	"log"
+	"os"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/d4l3k/messagediff"
```
maelvls, line 13: **Self-review:** I'd be in favor of removing this dependency. Even if it's just a testing dep, it isn't needed: we can use testify's Equal which already gives us a good diff.

maelvls: I've removed this dependency in 21e52292fc0601cbc9a7af18ee500b0384d385b2.

pkg/agent/config_test.go
maelvls: Done. I've renamed it to `http://api.venafi.eu`.

inteon:
`go run . agent --help` gives me the same result as before this change.

inteon (approved):
LGTM

maelvls:
Something I forgot to mention: this PR is stacked on top of #564:
- Before merging this one, I need to make sure that the other PR that I'm stacking this on (the "base" PR) is merged.
- Once the base PR is merged, I can change the base of this one to `master` and click "Merge".

pkg/agent/run.go
```diff
@@ -99,17 +38,34 @@ var Prometheus bool
...
+		logs.Log.Fatalf("Failed to load config file for agent from: %s", Flags.ConfigFilePath)
+	}
+	defer file.Close()
+
+	b, err := ioutil.ReadAll(file)
+	if err != nil {
+		logs.Log.Fatalf("Failed to read config file: %s", err)
+	}
+
+	cfg, err := ParseConfig(b, Flags.StrictMode)
```
maelvls, line 58: **Post-merge comment:** Oops, it looks like I made a mistake here... It should be `Flags.VenafiCloudMode`, not `Flags.StrictMode`!

PR #566, opened 2024-08-30
VC-35630: User can now omit the `server` field in config when using the Venafi Connection mode

maelvls:
Source: [VC-35630](https://venafi.atlassian.net/browse/VC-35630)

### Problem

EU customers need to change the URL of the VCP API to point to https://api.venafi.eu/. To do that, and imagining that they are using the VenafiConnection authentication, they may try to use the `spec.vcp.url` field on their VenafiConnection resource and find that this field doesn’t do anything because the Helm chart's `config.server` is set to https://api.venafi.cloud/ by default.

Another possible scenario is that EU customers may end up with a VenafiConnection configured with the `spec.vcp.url` field set to `https://api.venafi.eu`/. This VenafiConnection will have been already working well with venafi-enhanced-issuer and approver-policy-enterprise. Once they try to switch the Agent to the VenafiConnection auth method, they will see that it doesn’t work because the Agent picks up the default value in the Agent’s helm chart, i.e.,

```
config:
  server: https://api.venafi.cloud/.
```

### Solution

In this PR, I propose to ignore the `server` field specified in the config file in Venafi Connection mode. By default, the Helm chart sets `server` to `https://api.venafi.cloud`, which means the user may get confused as to what this field does. 

To avoid confusion, the logs now show the following message when `server` isn't empty:

> Ignoring the server field specified in the config file. In Venafi Connection mode, this field is not needed. Use the VenafiConnection's spec.vcp.url field instead.

This way, the user will know that what they need to do in case they see the `server` field and think "is that what I need to change?".

I've also clarified in the Helm chart that `server` is only used for Jetstack Secure OAuth, Jetstack Secure API Token, and Venafi Cloud Keypair authentication.

[VC-35630]: https://venafi.atlassian.net/browse/VC-35630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/client/client_venconn_test.go
```diff
@@ -269,186 +259,3 @@ func run(test testcase) func(t *testing.T) {
 		assert.Equal(t, test.expectReadyCondMsg, got.Status.Conditions[0].Message)
 	}
 }
-
-func fakeVenafiCloud(t *testing.T) (*httptest.Server, *x509.Certificate) {
```
maelvls, line 273: I moved this to the `testutil.go` package.

pkg/client/client_venafi_cloud.go
```diff
@@ -288,9 +288,6 @@ func (c *VenafiCloudClient) updateAccessToken() error {
 	values.Set("assertion", jwtToken)
 
 	tokenURL := fullURL(c.baseURL, accessTokenEndpoint)
-	if err != nil {
-		return err
-	}
```
maelvls, line 293: I've removed this because the linter was complaining.

pkg/testutil/testutil.go
```diff
@@ -0,0 +1,329 @@
...
+func FakeVenafiCloud(t *testing.T) (_ *httptest.Server, _ *x509.Certificate, setAssert func(AssertRequest)) {
+	t.Helper()
+
+	var assertFn AssertRequest
+	assertFnMu := sync.Mutex{}
+	setAssert = func(setAssert AssertRequest) {
+		assertFnMu.Lock()
+		defer assertFnMu.Unlock()
+		assertFn = setAssert
+	}
```
maelvls, line 200: I needed to make sure that the request's host was correct, that's why I wrote this clunky code to set an assertion... I'd like to have a clearer way of doing that, or even better: not have to assert the host on the request...

wallrj (approved):
Thanks @maelvls 

The changes look good to me and the tests pass on my laptop.

Additionally, I rebased https://github.com/jetstack/jetstack-secure/pull/562 on this branch and re-ran the e2e/test.sh script and observed the new log message and the agent successfully posted the data to the EU API.

```
$ ./hack/e2e/test.sh
...
+ grep -q 'Data sent successfully'
2024/09/03 08:22:13 Preflight agent version: development ()
2024/09/03 08:22:13 ignoring the server field specified in the config file. In Venafi Connection mode, this field is not needed. Use the VenafiConnection's spec.vcp.url field instead.
2024/09/03 08:22:13 Venafi Connection mode was specified, using Venafi Connection authentication.
2024/09/03 08:22:13 ignoring venafi-cloud.upload_path. In Venafi Connection mode, this field is not needed.
2024/09/03 08:22:13 ignoring venafi-cloud.uploader_id. In Venafi Connection mode, this field is not needed.
2024/09/03 08:22:13 Prometheus was enabled.
Running prometheus server on port :8081
2024/09/03 08:22:58 Posting data to: https://api.venafi.eu/
2024/09/03 08:22:59 Data sent successfully.
```

PR #569, opened 2024-09-16
Replace github.com/juju/errors with github.com/pkg/errors

inteon:
`github.com/pkg/errors` is used in the rest of the project and does not have a LGPL license


inteon:
replaced with https://github.com/jetstack/jetstack-secure/pull/570 which uses a branch on this repo.

PR #570, opened 2024-09-16
Replace github.com/juju/errors with github.com/pkg/errors

inteon:
`github.com/pkg/errors` is used in the rest of the project and does not have a LGPL license


SgtCoDFish (approved):
/lgtm
/approve

Thanks, this is great!

PR #571, opened 2024-09-16
Cleanup Kubernetes datagather

inteon:


pkg/agent/run.go
```diff
@@ -101,7 +100,6 @@ func Run(cmd *cobra.Command, args []string) {
 	}
 
 	dataGatherers := map[string]datagatherer.DataGatherer{}
-	var wg sync.WaitGroup
```
inteon, line 104: This waitgroup was not used.

pkg/agent/run.go
```diff
@@ -123,41 +121,31 @@ func Run(cmd *cobra.Command, args []string) {
...
-		// seconds to perform an initial sync. It may fail, and that's fine
-		// too, it will backoff and retry of its own accord. Initial boot
-		// will only be delayed by a max of 5 seconds.
-		bootCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
-		defer cancel()
-
-		// wait for the informer to complete an initial sync, we do this to
-		// attempt to have an initial set of data for the first upload of
-		// the run.
-		if err := newDg.WaitForCacheSync(bootCtx.Done()); err != nil {
```
inteon, line 136: Instead of waiting for each dataGatherer before starting to the next, we should first start all dataGatherers and loop again to wait for them to finish.

pkg/datagatherer/k8s/dynamic.go
```diff
@@ -143,20 +143,21 @@ var kubernetesNativeResources = map[schema.GroupVersionResource]sharedInformerFu
 
 // NewDataGatherer constructs a new instance of the generic K8s data-gatherer for the provided
 func (c *ConfigDynamic) NewDataGatherer(ctx context.Context) (datagatherer.DataGatherer, error) {
-	cl, err := NewDynamicClient(c.KubeConfigPath)
```
inteon, line 146: We only have to create a dynamic client when the GVK is not a native resource.
So this function should be called as part of the `else` body.

pkg/datagatherer/k8s/dynamic.go
```diff
@@ -200,34 +199,22 @@ func (c *ConfigDynamic) newDataGathererWithClient(ctx context.Context, cl dynami
...
+				options.FieldSelector = fieldSelector.String()
 			},
-		})
-		newDataGatherer.informer = informer
-		return newDataGatherer, nil
+		)
+		newDataGatherer.informer = factory.ForResource(c.GroupVersionResource).Informer()
 	}
 
-	factory := dynamicinformer.NewFilteredDynamicSharedInformerFactory(
```
inteon, line 221: We are creating these factories to create just 1 informer.
I still create the factory ( maybe this can be simplified in the future ), but I now let it go out-of-scope and continue with the created informer instead.

pkg/datagatherer/k8s/dynamic.go
```diff
@@ -432,16 +408,6 @@ func redactList(list []*api.GatheredResource) error {
 	return nil
 }
 
-// namespaceResourceInterface will 'namespace' a NamespaceableResourceInterface
-// if the 'namespace' parameter is non-empty, otherwise it will return the
-// given ResourceInterface as-is.
-func namespaceResourceInterface(iface dynamic.NamespaceableResourceInterface, namespace string) dynamic.ResourceInterface {
```
inteon, line 438: This function was unused.

pkg/agent/run.go
```diff
@@ -145,19 +141,24 @@ func Run(cmd *cobra.Command, args []string) {
 		dataGatherers[dgConfig.Name] = newDg
 	}
 
-	// wait for initial sync period to complete. if unsuccessful, then crash
```
inteon, line 148: This logic was not doing anything, since the WaitGroup was not being used.

pkg/datagatherer/k8s/dynamic.go
SgtCoDFish: Makes sense! Is it fair to say that there's not really a reason for newDataGathererWithClient to be a separate function? It's fishy that we call it with one of cl or clientset with the other set to nil.

Not a requirement in this PR, of course!

SgtCoDFish (approved):
/lgtm
/approve

The only reason I'm even a little concerned about this is that I don't really know how the agent is tested and the extent to which it's tested. On the assumption that it is well tested I'm happy to approve!

pkg/agent/run.go
```diff
@@ -101,7 +101,15 @@ func Run(cmd *cobra.Command, args []string) {
...
-	var wg sync.WaitGroup
+	group, gctx := errgroup.WithContext(ctx)
+
+	defer func() {
+		// TODO: replace Fatalf log calls with Errorf and return the error
+		cancel()
+		if err := group.Wait(); err != nil {
+			logs.Log.Fatalf("failed to wait for controller-runtime component to stop: %v", err)
+		}
+	}()
```
maelvls, line 112: What is this doing? It seems like we now cancel immediately, and then block on `group.Wait`?

inteon: Indeed, we cancel the context to start the shutdown and wait for the shutdown to complete using group.Wait

inteon: An error group will always cancel its context when one of the goroutines returns an error.

For successful execution, we only reach this defer call when the following for loop ends because OneShot is true: https://github.com/jetstack/jetstack-secure/blob/7f756c3448bc6d541935d89fc3ccd1e7150d18f2/pkg/agent/run.go#L167-L181
If OneShot is false, the loop will keep looping.

maelvls: In my opinion, this information should be written down in a comment in code, since this isn't obvious.

maelvls: Sorry about that, I was reading `go func` instead of `defer func`... It now makes sense.

PR #575, opened 2024-09-19
VC-36043: Better error messages, warnings, and flag descriptions, no more unexpected "unauthenticated" mode, and it is now possible to use the `--period` flag without also giving the `period` field

maelvls:
Ref: [VC-36043](https://venafi.atlassian.net/browse/VC-36043 "AGENT: Prevent breaking our customer's Agent deployments when changing the Agent's command-line flags code or YAML configuration code")

I propose a few non-breaking changes to the configuration file and command-line options that will make the experience of using Venafi Kubernetes Agent better with better error messages and less confusing behaviors (such as the removal of the "unauthenticated" mode).

Thanks to the config refactoring, I was also able to write further unit tests. **That said, I might have broken something that was working before since it was pretty much impossible to test things before.** I'd like to do some tests (run the Helm chart and see if everything works as intended) to increase our confidence.

## Changes

- The unauthenticated mode no longer exists. The unauthenticated mode caused a lot of confusion as it was randomly triggered when missing some flags. Now, if you don't provide the required flags, the command will fail with a helpful error showing you what you missed.
- You can no longer use `--private-key-path` along with `--credentials-path`. Previously, `--private-key-path` would be ignored if `--credentials-path` was provided. Now, the two options are mutually exclusive and a helpful message is shown when trying to use both.
  → **potential breaking change?**
- The field `uploader_id` in the configuration file is deprecated. Setting this field will no longer do anything. A warning is now shown when using this field. The reason this field was deprecated is that it was never used by the Venafi Cloud API. Behind the scenes, the `uploader_id` is arbitrarily set to `no` so that the API doesn't complain.
- The --private-key-path flag now defaults to the empty string. Previously, it defaulted to `/etc/venafi/agent/key/privatekey.pem`. This location wasn't used by any of our deployment options, and was confusing to users trying to use `--client-id`. A helpful message is now shown when trying to run `--client-id` without `--private-key-path`.
- The `--period` flag was required even though the `period` field in the config was set. You can now set the period using the `period` field in the config file without having to also pass `--period` or `-p`. The `--period` flag now takes precedence over the config file, and a warning is shown if you set both.
- In Venafi Cloud Key Pair Service Account mode, the `server` field is no longer required. It now defaults to `https://api.venafi.cloud`.
- Improvements to `--help`:

```diff
--- mine/help-before.txt	2024-09-20 14:44:12
+++ mine/help-after.txt	2024-09-20 14:43:58
@@ -11,22 +11,22 @@
 
 Flags:
   -c, --agent-config-file agent.yaml                      Config file location, default is agent.yaml in the current working directory. (default "./agent.yaml")
-      --api-token string                                  Token used for authentication when API tokens are in use on the backend
+      --api-token string                                  Turns on the Jetstack Secure API Token mode. Defaults to the value of the env var API_TOKEN.
       --backoff-max-time duration                         Max time for retrying failed data gatherers (given as XhYmZs). (default 10m0s)
-      --client-id string                                  Venafi Cloud Service Account client ID. If you use this flag you don't need to use --venafi-cloud as it will assume you are authenticating against Venafi Cloud. Using this removes the need to use a credentials file with Venafi Cloud mode.
-  -k, --credentials-file string                           Location of the credentials file. For OAuth2 based authentication.
+      --client-id string                                  Turns on the Venafi Cloud Key Pair Service Account mode. If you use this flag you don't need to use --venafi-cloud as it will assume you are authenticating with Venafi Cloud. Using this removes the need to use a credentials file.
+  -k, --credentials-file string                           Location of the credentials file. For the Jetstack Secure OAuth and Venafi Cloud Key Pair Service Account modes.
       --enable-metrics                                    Enables Prometheus metrics server on the agent (port: 8081).
       --enable-pprof                                      Enables the pprof profiling server on the agent (port: 6060).
   -h, --help                                              help for agent
-      --input-path string                                 Input file path, if used, it will read data from a local file instead of gathering data from clusters
-      --install-namespace string                          Namespace in which the agent is running. Only needed when running the agent outside of Kubernetes. Used for testing purposes.
-      --one-shot                                          Runs agent a single time if true, or continously if false
-      --output-path string                                Output file path, if used, it will write data to a local file instead of uploading to the preflight server
+      --input-path string                                 For testing purposes. In conjunction with --one-shot, it allows you to push manually crafted data readings (in JSON format) to the Venafi Cloud API without the need to connect to a Kubernetes cluster.
+      --install-namespace string                          For testing purposes. Namespace in which the agent is running. Only needed with the Venafi Cloud VenafiConnection modewhen running the agent outside of Kubernetes.
+      --one-shot                                          For testing purposes. The agent will run once and exit. It is often used in conjunction with --output-path and/or --input-path.
+      --output-path string                                For testing purposes. In conjunction with --one-shot, it allows you to write the data readings to a file instead of uploading to the server.
   -p, --period duration                                   Override time between scans in the configuration file (given as XhYmZs).
-      --private-key-path string                           Venafi Cloud Service Account private key path. (default "/etc/venafi/agent/key/privatekey.pem")
+      --private-key-path string                           To be used in conjunction with --client-id. The path to the private key file for the service account.
       --strict                                            Runs agent in strict mode. No retry attempts will be made for a missing data gatherer's data.
-      --venafi-cloud                                      Runs agent with parsing config (and credentials file if provided) in Venafi Cloud format if true.
-      --venafi-connection string                          Name of the VenafiConnection to be used. Using this flag will enable the VenafiConnection mode.
+      --venafi-cloud                                      Turns on the Jetstack Secure OAuth mode. The flag --credentials-file must also be passed.
+      --venafi-connection string                          Turns on the Venafi Cloud VenafiConnection mode. This flag configures the name of the VenafiConnection to be used.
       --venafi-connection-namespace allowReferencesFrom   Namespace of the VenafiConnection to be used. It is only useful when the VenafiConnection isn't in the same namespace as the agent. The field allowReferencesFrom must be present on the cross-namespace VenafiConnection for the agent to use it.
 
 Use "preflight agent [command] --help" for more information about a command.

```

## What I haven't changed

- It's still mandatory to provide `venafi-cloud.upload_path` even though it is pretty much always useless. I think we should at least default the value to what the Venafi Cloud backend expects and no force users to provide it in the config file.
- We should also deprecate `input-path` and `output-path`. I don't know why these exist. We can find easier ways of testing the agent without needing these two flags. Or at least, mention in the flag description that these are meant for testing purposes.

[VC-36043]: https://venafi.atlassian.net/browse/VC-36043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### Test 1: Venafi Cloud Key Pair Service Account mode

I've used the tenant https://ven-tlspk.venafi.cloud/. To access the API key, use the user `system.admin@tlspk.qa.venafi.io` and the password is visible in the page [Production Accounts](https://venafi.atlassian.net/wiki/spaces/CT/pages/2115404149) (private to Venafi). Then go to the settings and find the API key.

```sh
export APIKEY=...
```

```sh
venctl iam service-account agent create --name "mael temp" \
  --vcp-region US \
  --output json \
  --api-key $APIKEY
#--owning-team $(curl -sS https://api.venafi.cloud/v1/teams -H "tppl-api-key: $APIKEY" | jq '.teams[0].id') \
```

venctl version: 1.13.0

```sh
KO_DOCKER_REPO=ttl.sh/maelvls/venafi-agent ko build --bare --tags 0.0.0-dev
helm package deploy/charts/venafi-kubernetes-agent --version 0.0.0-dev --app-version 0.0.0-dev
helm push venafi-kubernetes-agent-0.0.0-dev.tgz oci://ttl.sh/maelvls/charts
export VENAFI_KUBERNETES_AGENT_CLIENT_ID=$(jq -r .client_id /tmp/agent-credentials.json)
cat <<EOF > /tmp/venafi-kubernetes-agent.yaml
config:
  clusterName: mael-tmp
EOF
kubectl create ns venafi --dry-run=client -o yaml | kubectl apply -f -
kubectl create secret generic -n venafi agent-credentials --from-literal=privatekey.pem="$(jq -r .private_key /tmp/agent-credentials.json)" \
  --dry-run=client -o yaml | kubectl apply -f -
venctl components kubernetes apply \
    --venafi-kubernetes-agent \
    --venafi-kubernetes-agent-values-files /tmp/venafi-kubernetes-agent.yaml \
    --venafi-kubernetes-agent-version 0.0.0-dev \
    --venafi-kubernetes-agent-custom-image-registry "ttl.sh/maelvls" \
    --venafi-kubernetes-agent-custom-chart-repository "oci://ttl.sh/maelvls/charts"
```

> [!NOTE]
>
> If you need to run it locally to debug things (e.g., in order to run
> mitmproxy):
>
> ```sh
> kubectl scale -n venafi deploy venafi-kubernetes-agent venafi-kubernetes-agent --replicas 0
> go run . agent -c <(kubectl get cm -n venafi agent-config -oyaml | yq '.data."config.yaml"') \
>   --client-id $(jq -r .client_id /tmp/agent-credentials.json) \
>   --private-key-path <(jq -r .private_key /tmp/agent-credentials.json) -p 5m
> ```

## Test 2: VenafiConnection

I've used the tenant https://ven-tlspk.venafi.cloud/. To access the API key, use the user `system.admin@tlspk.qa.venafi.io` and the password is visible in the page [Production Accounts](https://venafi.atlassian.net/wiki/spaces/CT/pages/2115404149) (private to Venafi). Then go to the settings and find the API key.

```sh
export VEN_API_KEY=...
export \
 VEN_OWNING_TEAM=$(curl --fail-with-body -sS https://api.venafi.cloud/v1/teams -H "tppl-api-key: $VEN_API_KEY" | jq '.teams[0].id' -r) \
 OCI_BASE=ttl.sh/maelvls \
 VEN_API_KEY_PULL=$(lpass show -p glow-in-the-dark.venafi.cloud) \
 VEN_API_HOST=api.venafi.cloud \
 VEN_ZONE='tlspk-bench\Default' \
 CLOUDSDK_CORE_PROJECT=jetstack-mael-valais \
 CLOUDSDK_COMPUTE_ZONE=europe-west1-b \
 CLUSTER_NAME=test-secretless
```

```sh
./hack/e2e/test.sh
```
and got:
```
2024/09/23 15:35:23 Posting data to:
W0923 15:35:23.962141       1 reflector.go:561] k8s.io/client-go@v0.31.1/tools/cache/reflector.go:243: failed to list *v1alpha1.VenafiConnection: venaficonnections.jetstack.io is forbidden: User "system:serviceaccount:venafi:venafi-connection" cannot list resource "venaficonnections" in API group "jetstack.io" at the cluster scope
E0923 15:35:23.962229       1 reflector.go:158] "Unhandled Error" err="k8s.io/client-go@v0.31.1/tools/cache/reflector.go:243: Failed to watch *v1alpha1.VenafiConnection: failed to list *v1alpha1.VenafiConnection: venaficonnections.jetstack.io is forbidden: User \"system:serviceaccount:venafi:venafi-connection\" cannot list resource \"venaficonnections\" in API group \"jetstack.io\" at the cluster scope" logger="UnhandledError"
```

### 404 `{"message":"Not Found"}`

```log
2023/10/12 15:26:37 Posting data to: https://api.venafi.cloud/
2023/10/12 15:26:37 retrying in 4m25.621466698s after error: post to server failed: received response with status code 404. Body: {"message":"Not Found"}
```

I fixed it by re-running `venctl iam` above and re-applying the secret.

### 400 `{"error":"invalid_grant","error_description":"token_signature_verification_error"}`

```log
2024/09/20 16:32:53 Posting data to: https://api.venafi.cloud/
2024/09/20 16:32:53 retrying in 1m32.91697267s after error: post to server failed: failed to execute http request to Venafi Control Plane. Request https://api.venafi.cloud/v1/oauth/token/serviceaccount, status code: 400, body: [{"error":"invalid_grant","error_description":"token_signature_verification_error"}]
```

It was because I forgot to update `VENAFI_KUBERNETES_AGENT_CLIENT_ID` and then
re-run `venctl components...`.

### 400 `{"error":"invalid_grant","error_description":"Not found"}`

```log
2024/09/23 08:19:49 Posting data to: https://api.venafi.cloud/
2024/09/23 08:19:50 retrying in 53.802932939s after error: post to server failed: failed to execute httprequest to Venafi Control Plane. Request https://api.venafi.cloud/v1/oauth/token/serviceaccount, status code: 400, body: [{"error":"invalid_grant","error_description":"Not found"}]
```

### Missing VenafiConnection CRDs/RBAC

While running "Test 2", I saw the following errors:

```text
2024/09/23 15:35:23 Posting data to:
W0923 15:35:23.962141       1 reflector.go:561] k8s.io/client-go@v0.31.1/tools/cache/reflector.go:243: failed to list *v1alpha1.VenafiConnection: venaficonnections.jetstack.io is forbidden: User "system:serviceaccount:venafi:venafi-connection" cannot list resource "venaficonnections" in API group "jetstack.io" at the cluster scope
E0923 15:35:23.962229       1 reflector.go:158] "Unhandled Error" err="k8s.io/client-go@v0.31.1/tools/cache/reflector.go:243: Failed to watch *v1alpha1.VenafiConnection: failed to list *v1alpha1.VenafiConnection: venaficonnections.jetstack.io is forbidden: User \"system:serviceaccount:venafi:venafi-connection\" cannot list resource \"venaficonnections\" in API group \"jetstack.io\" at the cluster scope" logger="UnhandledError"
```

I found that it is because I had partially run `./hack/e2e/test.sh`, the venafi-connection helm chart had been uninstalled somehow.

### Bug 1: `--private-key-path` no longer defaults to `/etc/venafi/agent/key/privatekey.pem` and needs to be explicitly set in the Helm chart

I found this bug in the "Test 1".

While testing, I found that the Agent was erroring with an error "file not
found: ''". This was because the `--private-key-path` flag was not set in the
Helm chart. I had to set it to `/etc/venafi/agent/key/privatekey.pem` in the
`deployment.yaml` file.

→ Fixed in 9de2d8ee16b5095f982ea8c1a85fb3ade5a8e086

### Bug 2: Missing path argument `uploader_id` leading to 404 `{"message":"Not Found"}`

I found this bug in the "Test 1".

This was sent by the Agent:

```http
POST https://api.venafi.cloud/v1/tlspk/upload/clusterdata?name=mael-tmp HTTP/2.0
authorization: Bearer ytC4j1WqxeMDXTTkTHOigA==
accept: application/json
content-length: 69665
accept-encoding: gzip

{"agent_metadata":{"version":"development","cluster_id":"mael-tmp"},...
```

Response was:

```text
HTTP/2.0 404
date: Mon, 23 Sep 2024 08:38:02 GMT
content-length: 24
server: envoy

{"message":"Not Found"}
```

It looks like the path argument `uploader_id` is missing

```text
/v1/tlspk/upload/clusterdata?name=mael-tmp      <-- INCORRECT
/v1/tlspk/upload/clusterdata/no?name=mael-tmp   <-- EXPECTED
```

(`no` is an arbitrary value, since this value is ignored by the Venafi Cloud
API)

I added a unit test for this bug in https://github.com/jetstack/jetstack-secure/pull/575#issuecomment-2368460385 and fixed the bug in ee395c0c038d9b81ba24bb5c2ea43e82da6a6840.

### Bug 3: `2024/09/23 15:35:23 Posting data to:` (server string is empty)

I found this bug in the "Test 2" (VenafiConnection mode). I noticed the following log line:

```
venafi-kubernetes-agent-75bb765698-j9gwl:2024/09/23 15:35:23 Posting data to:
```

I makes sense because this log line prints the `server` field, which is probably empty because the `server` field isn't used in venaficonnection mode; instead, the venaficonnection's `url` field is used.

I created the ticket https://venafi.atlassian.net/browse/VC-36379 to fix this as the fix isn't as simple (moving to each client's `PostData` func? Hide this log line when using VenafiConnection?)


pkg/client/client_venafi_cloud.go
```diff
@@ -149,9 +151,17 @@ func (c *VenafiSvcAccountCredentials) Validate() error {
...
+// only returned when `ok` is false.
+func (c *VenafiSvcAccountCredentials) IsClientSet() (ok bool, why string) {
+	if c.ClientID == "" {
+		return false, "ClientID is empty"
+	}
+	if c.PrivateKeyFile == "" {
+		return false, "PrivateKeyFile is empty"
+	}
+
+	return true, ""
```
inteon, line 164:
```suggestion
func (c *VenafiSvcAccountCredentials) IsClientSet() error {
	if c.ClientID == "" {
		return fmt.Errorf("ClientID is empty")
	}
	if c.PrivateKeyFile == "" {
		return fmt.Errorf("PrivateKeyFile is empty")
	}

	return nil
```
small nit

pkg/client/client_venconn_test.go
```diff
@@ -200,48 +194,7 @@ func run(test testcase) func(t *testing.T) {
...
-			  namespace: venafi
-			roleRef:
-			  apiGroup: rbac.authorization.k8s.io
-			  kind: Role
-			  name: venafi-connection-secret-reader
-			subjects:
-			- kind: ServiceAccount
-			  name: venafi-connection
-			  namespace: venafi`))...)
+		given = append(given, testutil.Parse(testutil.VenConnRBAC)...)
```
inteon, line 197: Could you add a TODO to the VenConnRBAC definition? I think once we get the Makefile modules setup, we should generate this based on the Helm chart.

pkg/client/client_unauthenticated.go
inteon, line 1: Are we sure this is not used anymore.

maelvls:
Notes after a quick call with Tim:

- Regarding the removal of the unauthenticated, let's ask the TLSPK Cloud Services team whether this would impact users.
- Regarding `uploader_id`, let's add a link to the Jira ticket where the path parameter will be removed from the backend.

inteon:
Some extra comments:
- We should add some in-code comments to track the removal of the uploaderID value: https://github.com/jetstack/jetstack-secure/blob/e3caf6f69e166a72e888178a33b2a27e37f6c4d3/pkg/client/client_venconn.go#L166 and https://github.com/jetstack/jetstack-secure/blob/e3caf6f69e166a72e888178a33b2a27e37f6c4d3/pkg/client/client_venafi_cloud.go#L174
+ add a `// Deprecated:` comment on the field

- Since everyone is using the same value for `UploadPath`, we might want to Deprecate that field too and make it non-functional.

pkg/client/client_unauthenticated.go
maelvls: I've asked TLSPK Cloud Services team [on Slack](https://venafi.slack.com/archives/C0738V9QK61/p1726824352735729). I'll share any update here.

pkg/agent/config.go
```diff
@@ -258,151 +268,430 @@ func InitAgentCmdFlags(c *cobra.Command, cfg *AgentCmdFlags) {
...
+			log.Printf("Using the %s auth mode since --venafi-connection was specified.", mode)
+		case flags.APIToken != "":
+			mode = JetstackSecureAPIToken
+			log.Printf("Using the %s auth mode since --api-token was specified.", mode)
+		case !flags.VenafiCloudMode && flags.CredentialsPath != "":
+			mode = JetstackSecureOAuth
+			log.Printf("Using the %s auth mode since --credentials-file was specified without --venafi-cloud.", mode)
+		default:
+			return CombinedConfig{}, nil, fmt.Errorf("no auth mode specified. You can use one of four auth modes:\n" +
+				" - Use --venafi-cloud with --credentials-file or --client-id with --private-key-path to use the " + string(VenafiCloudKeypair) + " mode.\n" +
```
inteon, line 352:
```suggestion
				" - Use (--venafi-cloud with --credentials-file) or (--client-id with --private-key-path) to use the " + string(VenafiCloudKeypair) + " mode.\n" +
```
Somehow it took me a long time to understand this line without the parentheses.

pkg/agent/config.go
```diff
@@ -258,151 +268,433 @@ func InitAgentCmdFlags(c *cobra.Command, cfg *AgentCmdFlags) {
...
+
+	// Validation of `uploader_id`.
+	//
+	// We found that `venafi-cloud.uploader_id` doesn't do anything in the
+	// backend. Since the backend requires it for historical reasons (but cannot
+	// be empty), we just ignore whatever the user has set in the config file,
+	// and set it to an arbitrary value in the client since it doesn't matter.
+	//
+	// TODO(mael): Remove the arbitrary `/no` path parameter from the Agent once
+	// https://venafi.atlassian.net/browse/VC-35385 is done.
```
maelvls, line 438: @inteon I added this comment to remind us that the `uploader_id` path parameter will soon be removed from the API (source: https://github.com/jetstack/jetstack-secure/pull/575#issuecomment-2363268458)

pkg/agent/config.go
maelvls: Good point, adding parentheses will make it much easier to read.

tfadeyi:
>We should also deprecate input-path and output-path. I don't know why these exist. We can find easier ways of testing the agent without needing these two flags. Or at least, mention in the flag description that these are meant for testing purposes.

I don't think we should be deprecating them as they are currently used on the cloud side for manually generating the data sent by the agent locally without having to connect an actual agent to the backend or install one in a cluster.
If we don't want users to see them, we can hide flags.

pkg/client/client_venconn_test.go
maelvls: Done. I've opened a ticket to track this TODO: https://venafi.atlassian.net/browse/VC-36331

pkg/client/client_unauthenticated.go
inteon: Looks like no customers are using this (@tfadeyi confirmed this).

pkg/agent/config.go
```diff
@@ -258,151 +268,433 @@ func InitAgentCmdFlags(c *cobra.Command, cfg *AgentCmdFlags) {
...
+	{
+		if res.AuthMode == VenafiCloudVenafiConnection || res.AuthMode == VenafiCloudKeypair {
```
inteon, line 440:
```suggestion

```

inteon:
@maelvls I think instead of `"github.com/hashicorp/go-multierror"` we should use https://pkg.go.dev/errors#Join.

inteon (approved):
LGTM
Thanks for making the requested changes.

TODO list:
- deprecate UploadPath
- replace "github.com/hashicorp/go-multierror"
- deduplicate some of the client logic (oauth vs ven cloud vs venconn)
- deprecate the JSS logic

pkg/agent/config.go
```diff
@@ -74,33 +86,44 @@ type AgentCmdFlags struct {
...
+	// instead uploading them to the Venafi Cloud API.
 	OutputPath string
 
-	// InputPath (--input-path) is where the agent will read data from instead
-	// of gathering data from clusters.
+	// InputPath (--input-path) is used for testing purposes. In conjunction
+	// with --one-shot, it allows you to push manually crafted data readings (in
+	// JSON format) to the Venafi Cloud API without the need to connect to a
+	// Kubernetes cluster. See the jscp-testing-cli's README for more info:
+	// https://gitlab.com/venafi/vaas/applications/tls-protect-for-k8s/cloud-services/-/tree/master/jscp-testing-cli
```
maelvls, line 126: As indicated by Olu in https://github.com/jetstack/jetstack-secure/pull/575#issuecomment-2363335206, `--input-path` is being used internally at Venafi to test out the Venafi Cloud API. See the jscp-testing-cli's README for more info: https://gitlab.com/venafi/vaas/applications/tls-protect-for-k8s/cloud-services/-/tree/master/jscp-testing-cli

I've added "Used for testing purposes" to make it clearer that it isn't intended for general use.

I thought about hiding `--one-shot`, `--output-path`, and `--input-path` with

```go
c.PersistentFlags().MarkHidden("input-path")
c.PersistentFlags().MarkHidden("output-path")
c.PersistentFlags().MarkHidden("one-shot")
```

but I've decided to just mark them with "for testing purposes" since it achieves the same outcome: avoid confusion for end-users as to whether this flag would be useful to them or not.

@tfadeyi Does it make sense?

pkg/client/client_venafi_cloud.go
maelvls: I've had a conversation with ChatGPT and I'm still not convinced that returning an error is better than `(bool, string)`. In short: `false` reports a "valid" state rather than a failure. It is just used to check that all the credential values (client ID, etc) are present. Example:

```go
ok, _ := credentials.IsClientSet()
	if !ok {
		credentials.ClientID = ClientID
		credentials.ClientSecret = ClientSecret
		credentials.AuthServerDomain = AuthServerDomain
	}
```

That's why a boolean makes more sense in my mind. Returning an error also feels off because when running `IsClientSet` you expect `false` to be returned.

That said, I do agree that the `why` value feels a bit off...

maelvls:
Before merging, I'll proceed with a two tests to increase my confidence that I won't be breaking anyone with this PR:

- Kind + Helm + image + venafi cloud with key pair service account using venctl's `VENAFI_KUBERNETES_AGENT_CLIENT_ID` (manual),
- Kind + Helm + image + venafi cloud with VenafiConnection (I'll use Richard's `test.sh` for that).

pkg/client/client_unauthenticated.go
maelvls: More context [from Slack](https://venafi.slack.com/archives/C0738V9QK61/p1726824739362869?thread_ts=1726824352.735729&cid=C0738V9QK61):

Olu: It was used used mainly for local development when we could run the saas locally. I don't think it will be needed anymore to be honest. It wasn't requested by customers.
David: Should we consider deprecating first?

Olu: I don't mind, if it's easier to deprecating I think it's fine, but afaik no JSS customer is using the agent in that way.


Tim:  Atm. it is easier to remove it directly (that is Mael already did in his PR).
Since no customer is using it, I think that means we can skip deprecating it...

Maël: And unauthenticated is a catch-all mode that only kicks in when you forget to put some flag like —venafi-cloud, so no one would have thought of using this mode intentionally since there is no explicit flag such as —noauth to enable it. I’m very confident that removing this will have no negative impact on existing deployments.

inteon (changes requested):
After testing using VenafiConnection authentication, I found the following error: `after error: post to server failed: received response with status code 400. Body: [{"message":"Bad Request"}`

maelvls:
> status code 400. Body: `[{"message":"Bad Request"}]`

I introduced a bug in this PR. In Venafi Cloud Key Pair Service Account mode, the `uploader_id` path segment became missing: the call is made to `/v1/tlspk/upload/clusterdata` instead of `/v1/tlspk/upload/clusterdata/no`. I've written a unit test in 4f9c5bc236b7b2e5f7b84ba04c93e871590024c6 to pin point the bug:

```
--- FAIL: Test_ValidateAndCombineConfig_VenafiCloudKeyPair (0.00s)
    --- FAIL: Test_ValidateAndCombineConfig_VenafiCloudKeyPair/server,_uploader_id,_and_cluster_name_are_correctly_passed (0.00s)
        envtest.go:188: fake api.venafi.cloud received request: POST /v1/oauth/token/serviceaccount
        envtest.go:188: fake api.venafi.cloud received request: POST /v1/tlspk/upload/clusterdata
        config_test.go:586:
                Error:          Not equal:
                                expected: "/v1/tlspk/upload/clusterdata/no"
                                actual  : "/v1/tlspk/upload/clusterdata"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -/v1/tlspk/upload/clusterdata/no
                                +/v1/tlspk/upload/clusterdata
                Test:           Test_ValidateAndCombineConfig_VenafiCloudKeyPair/server,_uploader_id,_and_cluster_name_are_correctly_passed
```

I'll now proceed to fix the bug in another commit.

inteon (approved):
see https://github.com/jetstack/jetstack-secure/pull/575#pullrequestreview-2318046395

We will further test these changes when we work on a new tools and testing framework.

maelvls:
I've rebased this (after all these years I'm still confused by "ours" and "theirs" when rebasing... I'll never get confortable resolving conflicts).

This is finally ready to get merged..

PR #585, opened 2024-10-03
README: Release process: push git tag before editing the release

maelvls:
As mentioned earlier today, I forgot to update the release process with the latest changes: we no longer need to create a GitHub Release. We now have to push a tag and wait for the GitHub Release to appear.


inteon (approved):
LGTM

PR #594, opened 2024-10-15
VC-36510: Key Pair and Venafi Connection modes now use gzip compression

maelvls:
**Summary:**

- This PR is a remediation for the 413s returned by our NGINX gateway due to the fact that the request limit was set too low (smth around 100MB). See [VC-36510](https://venafi.atlassian.net/browse/VC-36510) to know more.
- On 26 Sept, as a first attempt at remediating this, we bumped NGINX's [`client_max_body_size`](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size) from 256 MB to 1024 MB, and [`client_body_buffer_size`](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_buffer_size) from 128 MB to 750 MB ([gitlab MR](https://gitlab.com/venafi/vaas/applications/tls-protect-for-k8s/cloud-services/-/merge_requests/1237/diffs), [message](https://venafi.slack.com/archives/C04RXLJ99AB/p1727338288590219?thread_ts=1727331258.837779&cid=C04RXLJ99AB)).
  - Before this fix, NGINX used to return a 413 as soon as the `Content-Length` was above 256 MB. When the request body goes above 750 MB, the body is written to a temporary file to disk rather than buffered in memory.
  - After this this, NGINX hasn't returned any 413s, even when stress-testing it.
- Side note: **buffering in NGINX is still happening**. Keeping buffering on puts stress in terms of memory and disk on NGINX, and I don’t think there is a reason in buffering these requests. The setting in question is `proxy_request_buffering` and is `on` by default.
- The Venafi Control Plane API **doesn't publicly state that GZIP compression is possible**; I have tested it and it works, but we shall maybe talk to Prod Engineering.
- Compression hasn't been added for Standalone (Jetstack Secure OAuth and Jetstack Secure API token modes). In this PR, I only added compression for the Key Pair and Venafi Connection modes.
- Compression aims to lower the load on our intermediate NGNIX (because it buffers requests) as well as lower the egress cost for customers. Sending over 100MBs over the wire looked somewhat bad, enabling compression will improve that.
- To ease the pain that might be caused by adding compression, we decided to add `--disable-compression` and let the Support team tell customers. I'm also in favor of adding `--disable-compression` for development purposes (e.g., when using mitmproxy).

**Testing**

- [x] **Test 1:** (manual) test that the compression works in Venafi Connection mode by using `./hack/e2e/test.sh`. (see results below)
- [x] **Test 2:** (manual) test that the compression works in Key Pair mode. (see results below)
- [x] **Test 3:** test that the headers are correctly set in Key Pair mode. → unit test written.
- [x] **Test 4:** test that the headers are correctly set in Venafi Connection mode. → unit test written.
- [x] **Test 5:** test that no compression takes place when `--disable-compression` is used in Key Pair mode → unit test written
- [x] **Test 6:** test that no compression takes place when `--disable-compression` is used in Venafi Connection mode. → unit test written
- [x] **Test 7:** (manual) test that the data is actually in the S3 bucket. @tfadeyi will check this.

**Rollout Plan:**

- If Olu or somebody in Cloud Services is taking over the feature and releasing v1.2.0, they will assess how confident they are about releasing. Although this feature is important, it isn't critical, so they might wait until Richard is back (29th Oct) or Mael is back (4th Nov).
- Release v1.2.0 for all customers,
- `venctl` will not immediately get v1.2.0 for the reason explained in the `#venctl` channel. You can contact @SgtCoDFish or @tfadeyi to know more about that.
- Let the Support Team (@hawksight) know that this change might affect customers, and that they can tell them to use `--disable-compression` if a problem occurs.

**Before:** request body weights 49 MB:

```
POST https://api.venafi.cloud/v1/tlspk/upload/clusterdata/no?description=S2luZCBjbHVzdGVyIG9uIE1hZWwmIzM5O3MgQW9ydXMgaG9tZSBtYWNoaW5l&name=kind-mael HTTP/2.0
accept: application/json
authorization: Bearer REDACTED
content-length: 49000269
accept-encoding: gzip

{"agent_metadata":{...

HTTP/2.0 200 
date: Tue, 15 Oct 2024 08:42:45 GMT
www-authenticate: Bearer realm="Echo"
vary: Accept-Encoding
server: envoy
content-length: 70

{"status":"ok","organization":"756db001-280e-11ee-84fb-991f3177e2d0"}
```

**After:** request body weights 5 MB:

```
POST https://api.venafi.cloud/v1/tlspk/upload/clusterdata/no?description=S2luZCBjbHVzdGVyIG9uIE1hZWwmIzM5O3MgQW9ydXMgaG9tZSBtYWNoaW5l&name=kind-mael HTTP/2.0
authorization: Bearer REDACTED
accept: application/json
content-length: 5125690
accept-encoding: gzip

\x8b������\xff̽ͮh\xb9\x8d...

HTTP/2.0 200 
date: Tue, 15 Oct 2024 08:42:45 GMT
www-authenticate: Bearer realm="Echo"
vary: Accept-Encoding
server: envoy
content-length: 70

{"status":"ok","organization":"756db001-280e-11ee-84fb-991f3177e2d0"}
```

^ note that the above request is missing the `Content-Encoding: gzip` header. I fixed this in 22def464afb6f5d80f5b8aa68ad0fe058c6d37dc.

(see the section below to know how to reproduce this)

### Test 1

For this test, I've used the tenant https://ven-tlspk.venafi.cloud/. To access the API key, use the user `system.admin@tlspk.qa.venafi.io` and the password is visible in the page [Production Accounts](https://venafi.atlassian.net/wiki/spaces/CT/pages/2115404149) (private to Venafi). Then go to the settings and find the API key, and set it in the env var `APIKEY`.

The tenant https://ven-tlspk.venafi.cloud/ doesn't have the right tier to pull images, so I use an API key from the tenant https://glow-in-the-dark.venafi.cloud/, that's why I set the env var `APIKEY_GLOW_IN_THE_DARK`. Ask Atanas to get access to the tenant https://glow-in-the-dark.venafi.cloud/.

```sh
export APIKEY=...
export APIKEY_GLOW_IN_THE_DARK=...
```

Then:

```sh
export \
 OCI_BASE=ttl.sh/maelvls \
 VEN_API_KEY=$APIKEY \
 VEN_API_KEY_PULL=$APIKEY_GLOW_IN_THE_DARK \
 VEN_API_HOST=api.venafi.cloud \
 VEN_VCP_REGION=us \
 VEN_ZONE='tlspk-bench\Default' \
 CLOUDSDK_CORE_PROJECT=jetstack-mael-valais \
 CLOUDSDK_COMPUTE_ZONE=europe-west1-b \
 CLUSTER_NAME=test-secretless
```

Finally, run the smoke test Venafi Connection mode:

```sh
./hack/e2e/test.sh
```

Result:

```
2024/10/18 19:18:35 Posting data to:
2024/10/18 19:18:36 Data sent successfully.
```

### Test 2

For this test, I've decided to use mitmproxy to see exactly what is being sent on the wire.

First, copy the following script to a file called `bigjsonbody.go` (written in Go because bash is super slow):

```go
///usr/bin/true; exec /usr/bin/env go run "$0" "$@"

package main

import (
	"encoding/json"
	"flag"
	"fmt"
	"time"

	"github.com/google/uuid"
)

var (
	count = flag.Int("count", 5000, "Number of items to generate")
)

type Metadata struct {
	Name              string            `json:"name"`
	Namespace         string            `json:"namespace"`
	UID               string            `json:"uid"`
	CreationTimestamp string            `json:"creationTimestamp"`
	Labels            map[string]string `json:"labels"`
}

type Resource struct {
	Kind       string   `json:"kind"`
	APIVersion string   `json:"apiVersion"`
	Metadata   Metadata `json:"metadata"`
}

type Item struct {
	Resource Resource `json:"resource"`
}

type Data struct {
	Items []Item `json:"items"`
}

type Output struct {
	ClusterID    string    `json:"cluster_id"`
	DataGatherer string    `json:"data-gatherer"`
	Timestamp    time.Time `json:"timestamp"`
	Data         Data      `json:"data"`
}

func generateItem() Item {
	randomUID := uuid.New().String()
	return Item{
		Resource: Resource{
			Kind:       "Pod",
			APIVersion: "v1",
			Metadata: Metadata{
				Name:              "test-" + randomUID,
				Namespace:         "test",
				UID:               randomUID,
				CreationTimestamp: "2024-09-20T18:03:51Z",
				Labels: map[string]string{
					"k8s-app": "test",
				},
			},
		},
	}
}

func main() {
	flag.Parse()
	items := make([]Item, *count)

	for i := 0; i < *count; i++ {
		items[i] = generateItem()
	}

	output := Output{
		ClusterID:    "mael temp",
		DataGatherer: "k8s/pods",
		Timestamp:    time.Date(2024, time.September, 30, 16, 47, 32, 0, time.FixedZone("CET", 2*3600)),
		Data:         Data{Items: items},
	}

	outputJSON, err := json.MarshalIndent(output, "", "  ")
	if err != nil {
		fmt.Println("Error:", err)
		return
	}

	fmt.Println(string(outputJSON))
}
```
Then:

```
chmod +x bigjsonbody.go
```

Finally:

```
$ ./bigjsonbody.go --count=200000 >input-data.json
du -h input-data.json
87M    input-data.json
```

For this tests, I've used the tenant https://ven-tlspk.venafi.cloud/. To access the API key, use the user `system.admin@tlspk.qa.venafi.io` and the password is visible in the page [Production Accounts](https://venafi.atlassian.net/wiki/spaces/CT/pages/2115404149) (private to Venafi). Then go to the settings and find the API key, and set it as an env var:

```sh
export APIKEY=...
```

Create the service account and key pair:

```bash
venctl iam service-account agent create --name "$USER temp" \
  --vcp-region US \
  --output json \
  --owning-team $(curl -sS https://api.venafi.cloud/v1/teams -H "tppl-api-key: $APIKEY" | jq '.teams[0].id') \
  --api-key $APIKEY
```

Now, make sure to have `127.0.0.1 me` in your `/etc/hosts`.

Then, run mitmproxy with:

```sh
mitmproxy --mode regular@9090 --ssl-insecure -s /tmp/watch-stream.py --set client_certs=$(kubectl incluster --print-client-cert >/tmp/me.pem && echo /tmp/me.pem)
```

Run this:

```bash
cat <<EOF >minimal-config.yaml
cluster_id: "kind-mael"
cluster_description: "Kind cluster on Mael's machine"
venafi-cloud:
  uploader_id: "no"
data-gatherers: []
period: 3m
EOF
```

Finally, run the Agent with:

```sh
go install github.com/maelvls/kubectl-incluster@latest
export HTTPS_PROXY=http://localhost:9090 KUBECONFIG=/tmp/kube && KUBECONFIG= HTTPS_PROXY= kubectl incluster --replace-ca-cert ~/.mitmproxy/mitmproxy-ca-cert.pem --sa=venafi/venafi-kubernetes-agent | sed 's|127.0.0.1|me|' >/tmp/kube

go run . agent -c minimal-config.yaml \
  --client-id $(jq -r .client_id /tmp/agent-credentials.json) \
  --private-key-path <(jq -r .private_key /tmp/agent-credentials.json) \
  --install-namespace=venafi \
  --input-path=input-data.json \
  --one-shot
```

Result:

```
2024/10/18 19:18:35 Posting data to: https://api.venafi.cloud/
2024/10/18 19:18:36 Data sent successfully.
```

[VC-36510]: https://venafi.atlassian.net/browse/VC-36510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/client/client_venafi_cloud.go
+++ b/pkg/client/client_venafi_cloud.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"bytes"
+	"compress/gzip"
 	"crypto"
 	"crypto/ecdsa"
 	"crypto/ed25519"
@@ -50,6 +51,8 @@ type (
 		jwtSigningAlg jwt.SigningMethod
 		lock          sync.RWMutex
 
+		disableCompression bool
+
 		// Made public for testing purposes.
 		Client *http.Client
 	}
@@ -84,7 +87,7 @@ const (
 
 // NewVenafiCloudClient returns a new instance of the VenafiCloudClient type that will perform HTTP requests using a bearer token
 // to authenticate to the backend API.
-func NewVenafiCloudClient(agentMetadata *api.AgentMetadata, credentials *VenafiSvcAccountCredentials, baseURL string, uploaderID string, uploadPath string) (*VenafiCloudClient, error) {
+func NewVenafiCloudClient(agentMetadata *api.AgentMetadata, credentials *VenafiSvcAccountCredentials, baseURL string, uploaderID string, uploadPath string, disableCompression bool) (*VenafiCloudClient, error) {
 	if err := credentials.Validate(); err != nil {
 		return nil, fmt.Errorf("cannot create VenafiCloudClient: %w", err)
 	}
@@ -107,15 +110,16 @@ func NewVenafiCloudClient(agentMetadata *api.AgentMetadata, credentials *VenafiS
 	}
 
 	return &VenafiCloudClient{
-		agentMetadata: agentMetadata,
-		credentials:   credentials,
-		baseURL:       baseURL,
-		accessToken:   &venafiCloudAccessToken{},
-		Client:        &http.Client{Timeout: time.Minute},
-		uploaderID:    uploaderID,
-		uploadPath:    uploadPath,
-		privateKey:    privateKey,
-		jwtSigningAlg: jwtSigningAlg,
+		agentMetadata:      agentMetadata,
+		credentials:        credentials,
+		baseURL:            baseURL,
+		accessToken:        &venafiCloudAccessToken{},
+		Client:             &http.Client{Timeout: time.Minute},
+		uploaderID:         uploaderID,
+		uploadPath:         uploadPath,
+		privateKey:         privateKey,
+		jwtSigningAlg:      jwtSigningAlg,
+		disableCompression: disableCompression,
 	}, nil
 }
 
@@ -260,11 +264,39 @@ func (c *VenafiCloudClient) Post(path string, body io.Reader) (*http.Response, e
 		return nil, err
 	}
 
-	req, err := http.NewRequest(http.MethodPost, fullURL(c.baseURL, path), body)
+	var encodedBody io.Reader
+	if c.disableCompression {
+		encodedBody = body
+	} else {
+		compressed := new(bytes.Buffer)
+		gz := gzip.NewWriter(compressed)
+		if _, err := io.Copy(gz, body); err != nil {
+			return nil, err
+		}
+		err := gz.Close()
+		if err != nil {
+			return nil, err
+		}
+		encodedBody = compressed
+	}
+
+	req, err := http.NewRequest(http.MethodPost, fullURL(c.baseURL, path), encodedBody)
 	if err != nil {
 		return nil, err
 	}
 
+	// We have noticed that NGINX, which is Venafi Control Plane's API gateway,
+	// has a limit on the request body size we can send (client_max_body_size).
+	// On large clusters, the agent may exceed this limit, triggering the error
+	// "413 Request Entity Too Large". Although this limit has been raised to
+	// 1GB, NGINX still buffers the requests that the agent sends because
+	// proxy_request_buffering isn't set to off. To reduce the strain on NGINX'
+	// memory and disk, to avoid further 413s, and to avoid reaching the maximum
+	// request body size of customer's proxies, we have decided to enable GZIP
+	// compression. Ref: https://venafi.atlassian.net/browse/VC-36434.
+	if !c.disableCompression {
+		req.Header.Set("Content-Encoding", "gzip")
+	}
 	req.Header.Set("Accept", "application/json")
 	req.Header.Set("Content-Type", "application/json")
 
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -23,7 +23,8 @@ import (
 	"github.com/jetstack/preflight/pkg/version"
 )
 
-// Config wraps the options for a run of the agent.
+// Config defines the YAML configuration file that you can pass using
+// `--config-file` or `-c`.
 type Config struct {
 	// Deprecated: Schedule doesn't do anything. Use `period` instead.
 	Schedule string        `yaml:"schedule"`
@@ -159,6 +160,11 @@ type AgentCmdFlags struct {
 
 	// Prometheus (--enable-metrics) enables the Prometheus metrics server.
 	Prometheus bool
+
+	// DisableCompression (--disable-compression) disables the GZIP compression
+	// when uploading the data. Useful for debugging purposes, or when an
+	// intermediate proxy doesn't like compressed data.
+	DisableCompression bool
 }
 
 func InitAgentCmdFlags(c *cobra.Command, cfg *AgentCmdFlags) {
@@ -285,6 +291,12 @@ func InitAgentCmdFlags(c *cobra.Command, cfg *AgentCmdFlags) {
 		"Enables Prometheus metrics server on the agent (port: 8081).",
 	)
 
+	c.PersistentFlags().BoolVar(
+		&cfg.DisableCompression,
+		"disable-compression",
+		false,
+		"Disables GZIP compression when uploading the data. Useful for debugging purposes or when an intermediate proxy doesn't like compressed data.",
+	)
 }
 
 type AuthMode string
@@ -325,6 +337,9 @@ type CombinedConfig struct {
 	// VenafiCloudVenafiConnection mode only.
 	VenConnName string
 	VenConnNS   string
+
+	// VenafiCloudKeypair and VenafiCloudVenafiConnection modes only.
+	DisableCompression bool
 
 	// Only used for testing purposes.
 	OutputPath string
@@ -558,6 +573,14 @@ func ValidateAndCombineConfig(log *log.Logger, cfg Config, flags AgentCmdFlags) 
 		}
 	}
 
+	// Validation of --disable-compression.
+	{
+		if flags.DisableCompression && res.AuthMode != VenafiCloudKeypair && res.AuthMode != VenafiCloudVenafiConnection {
+			errs = multierror.Append(errs, fmt.Errorf("--disable-compression can only be used with the %s and %s modes", VenafiCloudKeypair, VenafiCloudVenafiConnection))
+		}
+		res.DisableCompression = flags.DisableCompression
+	}
+
 	if errs != nil {
 		return CombinedConfig{}, nil, errs
 	}
@@ -652,7 +675,7 @@ func validateCredsAndCreateClient(log *log.Logger, flagCredentialsPath, flagClie
 			break // Don't continue with the client if kubeconfig wasn't loaded.
 		}
 
-		preflightClient, err = client.NewVenConnClient(restCfg, metadata, cfg.InstallNS, cfg.VenConnName, cfg.VenConnNS, nil)
+		preflightClient, err = client.NewVenConnClient(restCfg, metadata, cfg.InstallNS, cfg.VenConnName, cfg.VenConnNS, nil, cfg.DisableCompression)
 		if err != nil {
 			errs = multierror.Append(errs, err)
 		}
@@ -710,7 +733,7 @@ func createCredentialClient(log *log.Logger, credentials client.Credentials, cfg
 			log.Println("Loading upload_path from \"venafi-cloud\" configuration.")
 			uploadPath = cfg.UploadPath
 		}
-		return client.NewVenafiCloudClient(agentMetadata, creds, cfg.Server, uploaderID, uploadPath)
+		return client.NewVenafiCloudClient(agentMetadata, creds, cfg.Server, uploaderID, uploadPath, cfg.DisableCompression)
 
 	case *client.OAuthCredentials:
 		return client.NewOAuthClient(agentMetadata, creds, cfg.Server)
--- a/pkg/agent/config_test.go
+++ b/pkg/agent/config_test.go
@@ -383,7 +383,7 @@ func Test_ValidateAndCombineConfig(t *testing.T) {
 				organization_id: foo
 				cluster_id: bar
 				`)),
-			withCmdLineFlags("--disable-compression", "--credentials-file", path))
+			withCmdLineFlags("--disable-compression", "--credentials-file", path, "--install-namespace", "venafi"))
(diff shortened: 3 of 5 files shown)


maelvls:
By the way, I don't understand why we have set a client timeout of 1 minute in Key Pair service account mode... Timeouts are a good practice in most cases, but I don't think this is such a case:

https://github.com/jetstack/jetstack-secure/blob/64a925584cc91e6b4ab977f70788bac0a9db67be/pkg/client/client_venafi_cloud.go#L114

pkg/client/client_venafi_cloud.go
```diff
@@ -260,13 +261,20 @@ func (c *VenafiCloudClient) Post(path string, body io.Reader) (*http.Response, e
 		return nil, err
 	}
 
-	req, err := http.NewRequest(http.MethodPost, fullURL(c.baseURL, path), body)
+	encodedBody := &bytes.Buffer{}
+	gz := gzip.NewWriter(encodedBody)
```
inteon, line 265: I thought that compression is handled automatically by the Go http client: https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/net/http/transport.go;l=194-202

pkg/client/client_venafi_cloud.go
```diff
@@ -260,13 +261,20 @@ func (c *VenafiCloudClient) Post(path string, body io.Reader) (*http.Response, e
...
+	}
+
+	req, err := http.NewRequest(http.MethodPost, fullURL(c.baseURL, path), encodedBody)
 	if err != nil {
 		return nil, err
 	}
 
 	req.Header.Set("Accept", "application/json")
 	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Content-Encoding", "gzip")
```
inteon, line 277: Is there a reason this header is missing in the example you show in the PR description?

wallrj (dismissed):
1. Add some comments in the code, explaining why the collected data is compressed using GZIP....to reduce the likelihood of exceeding the maximum request body size limit imposed by the server or by loadbalancers and proxies in between.
2. Show evidence that this still works with the old Jetstack TLSPK API server (does it handle requests with Content-Encoding: gzip?), or explain why that is not important.
3. Did you consider adding a flag to allow users to turn off the gzip content encoding, in case it doesn't work with a corporate proxy?
4. Or did you consider checking for HTTP server error 415 Unsupported Media Type and resending the request uncompressed?
5. Does the API documentation explain that the API supports gzip encoded requests?

pkg/client/client_venafi_cloud.go
inteon: I did some searching and could not find an automatic implementation for http requests on the client side. This seems to be implemented only for responses and when receiving requests.

pkg/client/client_venafi_cloud.go
maelvls: Well caught, I hadn't noticed that.

No idea why it doesn't show. This seems like a problem, I'll double check.

maelvls:
> Add some comments in the code, explaining why the collected data is compressed
> using GZIP....to reduce the likelihood of exceeding the maximum request body
> size limit imposed by the server or by loadbalancers and proxies in between.

Good point, I forgot to explain why `Content-Encoding` was added to the Key Pair
mode and the VenafiConnection mode, but not the Jetstack Secure OAuth mode nor
the Jetstack Secure API Token mode.

I've added a few comments in 22def46 to explain that.

> Show evidence that this still works with the old Jetstack TLSPK API server
> (does it handle requests with Content-Encoding: gzip?), or explain why that is
> not important.

I haven't added `Content-Encoding: gzip` to the Jetstack Secure OAuth mode nor
the Jetstack Secure API Token mode. I haven't added them in these two modes
because I have heard that the 413 Entity Too Large error was only happening with
the VCP API. I don't think it's worth changing how the way data is pushed to the
Jetstack Secure API ("Standalone") when we found no issue with it.

> Did you consider adding a flag to allow users to turn off the gzip content
> encoding, in case it doesn't work with a corporate proxy?

Good question... I haven't thought about that. I've looked around and found one
instance of a person struggling with using IIS as a reverse proxy in front of
SABnzbd; the problem seemed to be that SABnzbd's responses are GZIP-encoded; IIS
would respond "406 Not Acceptable":
https://github.com/sabnzbd/sabnzbd/issues/915.

I haven't found other evidence of people struggling with GZIP-encoded requests
being a problem in proxies. I conclude that the large majority of transparent
proxies, HTTP proxies, or SOCKS proxies, and the like found in companies handle
GZIP-encoded requests well.

I don't think there is a need to add a flag to disable GZIP encoding.

> Did you consider checking for HTTP server error 415 Unsupported Media Type
> and resending the request uncompressed?

I haven't considered this. It seems like some work to add this logic, and I'm
not sure how useful it will be.

**Note:** Looks like 415 Unsupported Media Type isn't the only possible error
that may show up: 406 Not Acceptable may also be one (as seen in the SABnzbd
issue I linked above).

> Does the API documentation explain that the API supports gzip encoded
> requests?

I had not thought of doing that. I just tried the google search [site:developer.venafi.com gzip](https://www.google.com/search?q=site%3Adeveloper.venafi.com+gzip) and didn't find anything; I also tried [site:docs.venafi.cloud gzip](https://www.google.com/search?q=site:docs.venafi.cloud+gzip), and I didn't find anything either.

There is also the internal documentation [API Documentation for /v1/tlspk/upload/clusterdata](https://venafi.atlassian.net/wiki/x/MgDdlg) but nothing there either.

I conclude that the VCP API doesn't officially support gzip-encoded requests... should I let the Production Engineering team know about this? I'm not sure what to do about it.

maelvls:
I realize that this PR is trickier than anticipated. @j-fuentes told "if it is low hanging fruit, just enable compression; if it is not low hanging fruit, discuss further actions". I'll have a chat with Jose to know what to do next.

maelvls:
@tfadeyi I'm done with the PR! I have added the necessary unit tests and have run a smoke test in Venafi Connection mode as well as in Key Pair mode. I am confident that this will work and will not break customers. You can proceed with the review and with merging and releasing v1.2.0 once you feel confident.

Before releasing, can you proceed with the "Test 7"? I've sent data to the tenant https://ven-tlspk.venafi.cloud/ at around 2024/10/18 19:21:37 UTC, can you check that the bucket contains the data?

pkg/client/client_venafi_cloud.go
maelvls: I've added four unit tests to make sure this is not a problem anymore:

- a unit test checks that the headers are correctly set in Key Pair mode..
- a unit test checks that the headers are correctly set in Venafi Connection mode..
- a unit test checks that no compression takes place when `--disable-compression` is used in Key Pair mode.
- a unit test checks that no compression takes place when `--disable-compression` is used in Venafi Connection mode.

tfadeyi:
> Before releasing, can you proceed with the "Test 7"? I've sent data to the tenant https://ven-tlspk.venafi.cloud/ at around 2024/10/18 19:21:37 UTC, can you check that the bucket contains the data?

I've tested e2e (pushing data from this branch of the agent to checking the blob storage) and the data is successfully uploaded to the backend side, with both `--disable-compression` and without.
**Note**: the backend already compresses data before storing it to the storage.

image

I'm going to look into the failing unit tests.

One thing I noticed is the behaviour of `POD_NAME` and `--install-namespace`, without either one of them the agent will be in a error state. This flag should be added to the helm chart as a value so users can change it. The flag/envvar should be set by default to be `venafi` as that is the default namespace,

 When setting `--install-namespace venafi` I still see the following line at the start of the agent logs:
 
 `2024/10/23 13:57:21 error messages will not show in the pod's events because the POD_NAME environment variable is empty`

tfadeyi (approved):

PR #596, opened 2024-10-16
[VC-35738] Log with klog to stdout and stderr in Kubernetes text format

wallrj:
I've settled on using klog and kubernetes component-base module for the reasons given in the comments in the logs.go file.

~~I think this logging UX is most likely to be familiar to Kubernetes platform admins who are likely to be deploying and debugging the venafi-kubernetes-agent.~~
> After discussion with @maelvls we settled on the flags: `-v`, `--log-level`, `--logging-format`, and `--vmodule`
> for maximum consistency with the other TLSPK components.

I gave up on Slog because it seemed to have unfamiliar severity / levels and I couldn't get the klog client-go logs to have the right levels with slog.

Using klog allows us to debug problems in client-go.
For example, you can view the API requests and responses using `--vmodule=round_tripper=6`.

In followup PRs I will convert the existing log lines to use context derived logr loggers.

```console
$ go test ./pkg/logs/... -v -run TestLogs -count 1

...
=== RUN   TestLogs/vmodule-level-3
    logs_test.go:291: FLAGS
        --vmodule=logs_test=3
    logs_test.go:295: STDOUT
        I1025 17:39:47.529791   62093 logs.go:161] "log Print" source="vcert"
        I1025 17:39:47.529998   62093 logs_test.go:43] "slog Info"
        I1025 17:39:47.530019   62093 logs_test.go:46] klog Info
        I1025 17:39:47.530031   62093 logs_test.go:49] "klog InfoS" key="value"
        I1025 17:39:47.530035   62093 logs_test.go:51] "Contextual Info Level 3" logger="foo" key="value"

    logs_test.go:296: STDERR
        W1025 17:39:47.530012   62093 logs_test.go:44] "slog Warn"
        E1025 17:39:47.530015   62093 logs_test.go:45] "slog Error"
        W1025 17:39:47.530023   62093 logs_test.go:47] klog Warning
        E1025 17:39:47.530027   62093 logs_test.go:48] "klog Error" err="fake-error"
        E1025 17:39:47.530039   62093 logs_test.go:52] "Contextual error" err="fake-error" logger="foo" key="value"

--- PASS: TestLogs (0.07s)
    --- PASS: TestLogs/help (0.01s)
    --- PASS: TestLogs/unrecognized-flag (0.01s)
    --- PASS: TestLogs/v-long-form-not-available (0.01s)
    --- PASS: TestLogs/logging-format-unrecognized (0.01s)
    --- PASS: TestLogs/original-defaults (0.01s)
    --- PASS: TestLogs/modified-defaults (0.01s)
    --- PASS: TestLogs/logging-format-json (0.01s)
    --- PASS: TestLogs/log-json-split-stream-false (0.01s)
    --- PASS: TestLogs/logging-format-text (0.01s)
    --- PASS: TestLogs/log-text-split-stream-false (0.01s)
    --- PASS: TestLogs/v-level-3 (0.01s)
    --- PASS: TestLogs/log-level-3 (0.01s)
    --- PASS: TestLogs/vmodule-level-3 (0.01s)
```

```console
$ ./hack/e2e/test.sh
...
{
  "ts": 1729108594511.0774,
  "caller": "log/log.go:245",
  "msg": "successfully gathered 30 items from \"k8s/replicasets\" datagatherer",
  "v": 0
}
{
  "ts": 1729108594511.1047,
  "caller": "log/log.go:245",
  "msg": "Posting data to: ",
  "v": 0
}
{"ts":1729108595121.0984,"caller":"log/log.go:245","msg":"Data sent successfully.","v":0}
```

image


hack/e2e/test.sh
```diff
@@ -191,14 +191,11 @@ EOF
...
 # See https://unix.stackexchange.com/questions/274120/pipe-fail-141-when-piping-output-into-tee-why
-set +o pipefail
 kubectl logs deployments/venafi-kubernetes-agent \
-  --follow \
-  --namespace venafi \
-  | tee >(grep -v -e "reflector\.go" -e "datagatherer" -e "data gatherer" >/dev/stderr) \
-  | grep -q "Data sent successfully"
+        --follow \
+        --namespace venafi \
+    | timeout 60 jq 'if .msg | test("Data sent successfully") then . | halt_error(0) end'
```
wallrj, line 201: This is a hack but it halts the pipeline when the expected `msg` value is observed and prints the matching record to stderr:
 * https://jqlang.github.io/jq/manual/#halt_error

The timeout stops the pipeline after 60 seconds if the expected msg is not observed.

pkg/logs/logs.go
```diff
@@ -1,8 +1,107 @@
...
 	"log"
 	"os"
+
+	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/component-base/featuregate"
+	"k8s.io/component-base/logs"
+	logsapi "k8s.io/component-base/logs/api/v1"
+	_ "k8s.io/component-base/logs/json/register"
```
wallrj, line 14: Importing this module, is what makes the json log format option available.

james-w:
Hi, does cert-manager and the other components use klog? Do they use the conventions that this MR is adopting?

maelvls:
> Does cert-manager and the other components use klog? Do they use the conventions that this MR is adopting?

To the question of consistency, here is a picture of what we have right now:

To the question of consistency, here is a picture of what we have right now:

| project                                  | library                        | flag                | values                   | default severity(level) | supports json format                                             |
| ---------------------------------------- | ------------------------------ | ------------------- | ------------------------ | ----------------------- | ---------------------------------------------------------------- |
| cert-manager                             | klog                           | `-v`, `--v`         | 0 to 9                   | info(4)                 | yes with `--logging-format=json`, default to klog's text format  |
| cert-manager webhook                     | klog                           | `-v`, `--v`         | 0 to 9                   | info(4)                 | yes with `--logging-format=json`, default to klog's text format  |
| openshift-routes                         | klogr                          | `-v`, `--log-level` | 0 to 9                   | info(1)                 | no, uses klog's text format                                      |
| csi-driver                               | klogr                          | `-v`, `--log-level` | 0 to 9                   | info(1)                 | no, uses klog's text format                                      |
| trust-manager                            | klogr                          | `-v`, `--log-level` | 0 to 9                   | info(1)                 | yes with `--log-format=json`, defaults to klog's text format     |
| approver-policy                          | klogr                          | `-v`, `--log-level` | 0 to 9                   | info(1)                 | yes with `--log-format=json`, defaults to klog's text format     |
| venafi-enhanced-issuer                   | zap                            | `-zap-log-level`    | debug,info,error, 1 to 9 | info                    | yes with `--zap-encoder=json`, defaults to zap's text format     |
| Venafi Kubernetes Agent (before this PR) | std's log and klog (client-go) | none                |                          | info(4)                 | yes with `--logging-format=json`, defaults to klog's text format |
| Venafi Kubernetes Agent (with this PR)   | klog                           | `-v`, `--v`         | 1 to 9                   | ?                       | no, uses log's and klog's text format                            |
| venctl                                   | internal based on slog         | `--log-level`       | debug,info,warn,error    | info                    | yes with `--log-format=json`, defaults to slog's text format     |
| vcert                                    | std's log                      | `--verbose`         |                          |                         | no, uses log's text format                                       |

Summary:

- The library convention is klog for older projects such as cert-manager and klogr for newer projects,
- The default format convention seems to be klog's text format.

james-w:
Hi, yep, consistency was exactly my concern, thanks for the detailed response. In my opinion it's favourable to have a smaller set to make updates easier down the line, even if that smaller set doesn't include the current best practice yet. If this project were to adopt the conventions of the majority of the other projects then we would have more consistency, and could figure out how/when to migrate them en-masse. Alternatively there may be an argument to update to the latest/greatest now, it depends on the details.

It sounds like you think that `log/slog` would be preferable to `klog` at this moment in this project?

maelvls:
**Regarding the flags**, it seems we (the cert-manager project) have initiated the convention of the flags `-v` and `--log-level`. It has nothing to do with klogr; klogr itself doesn't have a convention around the flag names. Example: https://github.com/cert-manager/openshift-routes/blob/0da68aed69ef66c8c9cefde45d855410813a6b3c/internal/cmd/app/options/options.go#L114-L116

I propose that we only expose the `-v` flag for now since it is common to both the cert-manager project and approver-policy, trust-manager, openshift-routes, and csi-driver. 

**Regarding the output format**, @wallrj chose to make it JSON even though none of our projects currently output JSON by default. Richard mentioned the [Kubernetes logging conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md), which says that the JSON format must be enabled with `--logging-format`, which means that if not passed, the output shouldn't be JSON:

> Requires passing `--logging-format=json` to enable.

I will thus set the default output to "text".

Regarding the slog severity problem:

> I gave up on Slog because it seemed to have unfamiliar severity / levels and I couldn't get the klog client-go logs to have the right levels with slog.

I've investigated and was able to configure an slog logger. To avoid lots of unrelated changes, the agent logs are still using the standard library's `*log.Logger`.

pkg/logs/logs_test.go
```diff
@@ -0,0 +1,453 @@
...
+I0000 00:00:00.000000   00000 logs_test.go:000] "klog InfoS" key="value"
+E0000 00:00:00.000000   00000 logs_test.go:000] "Contextual error" err="fake-error" logger="foo" key="value"
+`,
+		},
+		{
+			name:  "modified-defaults",
+			flags: "",
+			expectStdout: `
+{"ts":0000000000000.000,"caller":"logs/logs.go:000","msg":"log Print","source":"vcert","v":0}
+{"ts":0000000000000.000,"caller":"logs/logs_test.go:000","msg":"slog Info","v":0}
```
maelvls, line 144: I wish we were able to show the real source of the log line rather than the generic `logs/logs_test.go`. I guess we can improve this once we have migrated the `logs.Log.Printf` calls to `slog.Info`.

wallrj:
> @wallrj IMO we should set the default output to "text"... WDYT? I haven't reverted to "text". I'll let you decide.

I agree that text format by default is more consistent with other TLSPK components.  Done.

> @wallrj Feel free to merge this PR once you are happy with it. I'll approve it now.

I'll merge this after the 1.2.0 release because I'd like to finish the other logging improvements before releasing to customers.

pkg/logs/logs.go
```diff
@@ -1,8 +1,164 @@
...
+		fmt.Fprintf(os.Stderr, "Error in logging configuration: %v\n", err)
+		os.Exit(2)
+	}
+
+	// Thanks to logs.InitLogs, slog.Default now uses klog as its backend. Thus,
+	// the client-go library, which relies on klog.Info, has the same logger as
+	// the agent, which still uses log.Printf.
+	slog := slog.Default()
+
+	Log = &log.Logger{}
```
inteon, line 128: Can we just use `slog.SetDefault(...)` here?
Using `slog.New(logr.ToSlogHandler(logger))` we can convert the klog logger to an slog logger.

pkg/logs/logs.go
```diff
@@ -1,8 +1,164 @@
...
+		}
+
+		// The original usage string includes details about how
+		// JSON logging is only available when BETA logging features are
+		// enabled, but that's not relevant here because the feature is enabled
+		// by default.
+		if f.Name == "logging-format" {
+			f.Usage = `Sets the log format. Permitted formats: "json", "text".`
+		}
+		if f.Name == "log-text-split-stream" {
```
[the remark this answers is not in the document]
wallrj: Here's my current understanding:
1. I enabled `log-json-split-stream` by default to solve the problem that Google Logs Explorer otherwise can't distinguish between JSON formatted errors and info messages.
    > Severities: By default, logs written to the standard output are on the INFO level and logs written to the standard error are on the ERROR level. Structured logs can include a severity field, which defines the log's severity.
    > -- https://cloud.google.com/kubernetes-engine/docs/concepts/about-logs#best_practices
3. `log-text-split-stream` is not technically necessary to solve that problem because Google Logs Explorer seems to parse the `E`, `W`, and `I` prefix of the Kubernetes text format. So yes, I could remove this change.
4. Kubernetes introduced the split-stream options to solve performance problems. Info and warn messages can be buffered and flushed periodically while errors are flushed immediately.
    * https://github.com/kubernetes/kubernetes/pull/104873
    * https://github.com/kubernetes/enhancements/pull/2912#discussion_r703346009
6. By default logs are not buffered and are flushed to stdout and stderr immediately, so there shouldn't be any out-of-order log messages.
7. If log buffering is enabled, then I *think* that the buffered info logs will be flushed whenever an error is logged, so there should be any out-of-order log messages in that case either.
8. Kubernetes plan to make split-stream JSON logging the default for the Kubernetes components.
   > Ability to route logs based on type/verbosity will be replaced with default splitting info and errors logs to stdout and stderr. We will make this change only in alternative logging formats (like JSON) as we don't want to introduce breaking change in default configuration. Splitting stream will allow treating info and errors with different priorities. It will unblock efforts like https://github.com/kubernetes/klog/issues/209 to make info logs non-blocking.
   > 
   > -- https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/2845-deprecate-klog-specific-flags-in-k8s-components#logging-defaults

pkg/logs/logs.go
wallrj: Maybe. I'll consider that in a follow up.

PR #598, opened 2024-10-24
VC-36547: Remove gabs library as a dependency

inteon:
Instead of using the `gabs` library and an unnecessary marshal + unmarshal roundtrip, we can just use the `unstructured` helpers.

This is the first step in the plan outlined here: https://venafi.atlassian.net/browse/VC-36216?focusedCommentId=439503

### DIFF
--- a/pkg/datagatherer/k8s/fieldfilter.go
+++ b/pkg/datagatherer/k8s/fieldfilter.go
@@ -1,143 +1,88 @@
 package k8s
 
 import (
-	"fmt"
-	"strings"
-
-	"github.com/Jeffail/gabs/v2"
-	json "github.com/json-iterator/go"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
 // SecretSelectedFields is the list of fields sent from Secret objects to the
 // backend
-var SecretSelectedFields = []string{
-	"kind",
-	"apiVersion",
-	"metadata.annotations",
-	"metadata.labels",
-	"metadata.name",
-	"metadata.namespace",
-	"metadata.ownerReferences",
-	"metadata.selfLink",
-	"metadata.uid",
-	"type",
-	"/data/tls.crt",
-	"/data/ca.crt",
+var SecretSelectedFields = []FieldPath{
+	{"kind"},
+	{"apiVersion"},
+	{"metadata", "annotations"},
+	{"metadata", "labels"},
+	{"metadata", "name"},
+	{"metadata", "namespace"},
+	{"metadata", "ownerReferences"},
+	{"metadata", "selfLink"},
+	{"metadata", "uid"},
+
+	{"type"},
+	{"data", "tls.crt"},
+	{"data", "ca.crt"},
 }
 
 // RouteSelectedFields is the list of fields sent from OpenShift Route objects to the
 // backend
-var RouteSelectedFields = []string{
-	"kind",
-	"apiVersion",
-	"metadata.annotations",
-	"metadata.name",
-	"metadata.namespace",
-	"metadata.ownerReferences",
-	"metadata.selfLink",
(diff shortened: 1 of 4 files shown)


pkg/datagatherer/k8s/fieldfilter_test.go
```diff
@@ -18,10 +18,10 @@ func TestSelect(t *testing.T) {
 			"metadata": map[string]interface{}{
 				"name":      "example",
 				"namespace": "example",
-				"annotations": map[string]string{
+				"annotations": map[string]interface{}{
```
inteon, line 21: None of the maps should be of type `map[string]string{}`, kubernetes uses yaml.v2 which always produces `map[string]interface{}{}` maps.

yaml.v3 would produce `map[string]string{}`, but Kubernetes does not use that version and we expect the unstructured helpers to get updated once they switch.

inteon, line 21: We can expect all maps to be of type `map[string]interface{}`, that is the behavior for Kubernetes objects unmarshalled in unstructured objects (eg. https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/unstructured/helpers.go#L259).

pkg/datagatherer/k8s/fieldfilter.go
```diff
@@ -1,143 +1,88 @@
...
-	"metadata.annotations",
-	"metadata.labels",
-	"metadata.name",
-	"metadata.namespace",
-	"metadata.ownerReferences",
-	"metadata.selfLink",
-	"metadata.uid",
-	"type",
-	"/data/tls.crt",
-	"/data/ca.crt",
```
wallrj, line 26: Why did this original code use `/` as path separator for these two fields?

pkg/datagatherer/k8s/fieldfilter.go
```diff
@@ -1,143 +1,88 @@
...
+	{"spec", "tls", "destinationCACertificate"},
+	{"spec", "tls", "insecureEdgeTerminationPolicy"},
+	{"spec", "wildcardPolicy"},
+	{"status"},
 }
 
 // RedactFields are removed from all objects
-var RedactFields = []string{
-	"metadata.managedFields",
-	"/metadata/annotations/kubectl.kubernetes.io~1last-applied-configuration",
```
wallrj, line 57: Oh, maybe something to do with handling field path segments that may contain `/`.

wallrj (approved):
Thanks @inteon 

* I see that this is part of a plan that you and @maelvls agreed upon.
* Good to remove this unnecessary dependency and the resulting code is more readable.
* I ran the `./hack/e2e/test.sh` script and it passed.
* I noticed some comments and code in `pkg.datagatherere.k8s.dynamic.go:redactList` which might be obsoleted by this change. Ignore this comment if you disagree.
    ```
    // all objects fetched from sharedIndexInformers is now redacted
    // removing the managedFields and `kubectl.kubernetes.io/last-applied-configuration` annotation
    ```
    ```
    
    // During the internal marshal/unmarshal the runtime.Object the metav1.TypeMeta seems to be lost
    // this section reassigns the TypeMeta to the resource
    ```


https://github.com/jetstack/jetstack-secure/blob/2ebec9ef6066c4ad9f0b7e216a0d62f7e4babf39/pkg/datagatherer/k8s/dynamic.go#L349-L409

/approve
/lgtm

inteon:
It is still useful to remove annotations and managed fields here: https://github.com/jetstack/jetstack-secure/blob/2ebec9ef6066c4ad9f0b7e216a0d62f7e4babf39/pkg/datagatherer/k8s/dynamic.go#L383-L384

Because the `Select` and `Redact` functions are only applied on unstructured resources (vs typed resources).

PR #601, opened 2024-10-30
[VC-35738] Use errgroup for all go routines

wallrj:
This is part of the effort to remove the use of the legacy `log` module and replace it with `klog`.

Instead of creating free goroutines and relying on `log.Fatal` to terminate the process if any of them fail,
this PR instead adds those goroutines to the existing `errgroup` and replaces those instances of `log.Fatal` with returned errors which can be bubbled up to the parent Cobra command and handled centrally, in https://github.com/jetstack/jetstack-secure/pull/599.

These changes have been extracted from https://github.com/jetstack/jetstack-secure/pull/593 for easier review.

## Testing

Occupy the server listening port and observe that the process exits with a meaningful error message.

```console
python3 -m http.server 8081 &

$ POD_NAMESPACE=venafi ./preflight agent --output-path /dev/null --api-token should-not-be-required  --agent-config-file examples/cert-manager-agent.yaml --enable-pprof --period 1s
2024/10/30 15:41:00 Preflight agent version: development ()
2024/10/30 15:41:00 Using the Jetstack Secure API Token auth mode since --api-token was specified.
2024/10/30 15:41:00 Using deprecated Endpoint configuration. User Server instead.
2024/10/30 15:41:00 pprof profiling was enabled.
2024/10/30 15:41:00 error messages will not show in the pod's events because the POD_NAME environment variable is empty
2024/10/30 15:41:00 starting "k8s/secrets.v1" datagatherer
2024/10/30 15:41:00 starting "k8s/certificates.v1alpha2.cert-manager.io" datagatherer
2024/10/30 15:41:00 starting "k8s/ingresses.v1beta1.networking.k8s.io" datagatherer
2024/10/30 15:41:00 starting "k8s/certificaterequests.v1alpha2.cert-manager.io" datagatherer
2024/10/30 15:41:00 failed to complete initial sync of "k8s-dynamic" data gatherer "k8s/secrets.v1": timed out waiting for Kubernetes caches to sync
2024/10/30 15:41:00 failed to complete initial sync of "k8s-dynamic" data gatherer "k8s/certificates.v1alpha2.cert-manager.io": timed out waiting for Kubernetes caches to sync
2024/10/30 15:41:00 failed to complete initial sync of "k8s-dynamic" data gatherer "k8s/ingresses.v1beta1.networking.k8s.io": timed out waiting for Kubernetes caches to sync
2024/10/30 15:41:00 failed to complete initial sync of "k8s-dynamic" data gatherer "k8s/certificaterequests.v1alpha2.cert-manager.io": timed out waiting for Kubernetes caches to sync
2024/10/30 15:41:00 successfully gathered 0 items from "k8s/secrets.v1" datagatherer
2024/10/30 15:41:00 successfully gathered 0 items from "k8s/certificates.v1alpha2.cert-manager.io" datagatherer
2024/10/30 15:41:00 successfully gathered 0 items from "k8s/ingresses.v1beta1.networking.k8s.io" datagatherer
2024/10/30 15:41:00 successfully gathered 0 items from "k8s/certificaterequests.v1alpha2.cert-manager.io" datagatherer
2024/10/30 15:41:00 Data saved to local file: /dev/null
2024/10/30 15:41:00 failed to wait for controller-runtime component to stop: failed to run the health check server: listen tcp :8081: bind: address already in use
```

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -74,7 +74,16 @@ func Run(cmd *cobra.Command, args []string) {
 		logs.Log.Fatalf("While evaluating configuration: %v", err)
 	}
 
-	go func() {
+	group, gctx := errgroup.WithContext(ctx)
+	defer func() {
+		// TODO: replace Fatalf log calls with Errorf and return the error
+		cancel()
+		if err := group.Wait(); err != nil {
+			logs.Log.Fatalf("failed to wait for controller-runtime component to stop: %v", err)
+		}
+	}()
+
+	group.Go(func() error {
 		server := http.NewServeMux()
 
 		if Flags.Profiling {
@@ -105,21 +114,25 @@ func Run(cmd *cobra.Command, args []string) {
 
 		err := http.ListenAndServe(":8081", server)
 		if err != nil && !errors.Is(err, http.ErrServerClosed) {
-			logs.Log.Fatalf("failed to run the health check server: %s", err)
+			return fmt.Errorf("failed to run the health check server: %s", err)
 		}
-	}()
+		// The agent must stop if the management server stops
+		cancel()
+		return nil
+	})
 
 	_, isVenConn := preflightClient.(*client.VenConnClient)
 	if isVenConn {
-		go func() {
-			err := preflightClient.(manager.Runnable).Start(ctx)
+		group.Go(func() error {
+			err := preflightClient.(manager.Runnable).Start(gctx)
 			if err != nil {
-				logs.Log.Fatalf("failed to start a controller-runtime component: %v", err)
+				return fmt.Errorf("failed to start a controller-runtime component: %v", err)
 			}
 
 			// The agent must stop if the controller-runtime component stops.
 			cancel()
-		}()
+			return nil
+		})
 	}
 
 	// To help users notice issues with the agent, we show the error messages in
@@ -130,15 +143,6 @@ func Run(cmd *cobra.Command, args []string) {
 	}
 
 	dataGatherers := map[string]datagatherer.DataGatherer{}
-	group, gctx := errgroup.WithContext(ctx)
-
-	defer func() {
-		// TODO: replace Fatalf log calls with Errorf and return the error
-		cancel()
-		if err := group.Wait(); err != nil {
-			logs.Log.Fatalf("failed to wait for controller-runtime component to stop: %v", err)
-		}
-	}()
 
 	// load datagatherer config and boot each one
 	for _, dgConfig := range config.DataGatherers {
@@ -160,6 +164,8 @@ func Run(cmd *cobra.Command, args []string) {
 			if err := newDg.Run(gctx.Done()); err != nil {
 				return fmt.Errorf("failed to start %q data gatherer %q: %v", kind, dgConfig.Name, err)
 			}
+			// The agent must stop if any of the data gatherers stops
+			cancel()
 			return nil
 		})
 
@@ -192,7 +198,12 @@ func Run(cmd *cobra.Command, args []string) {
 
 	// begin the datagathering loop, periodically sending data to the
 	// configured output using data in datagatherer caches or refreshing from
-	// APIs each cycle depending on datagatherer implementation
+	// APIs each cycle depending on datagatherer implementation.
+	// If any of the go routines exit (with nil or error) the main context will
+	// be cancelled, which will cause this blocking loop to exit
+	// instead of waiting for the time period.
+	// TODO(wallrj): Pass a context to gatherAndOutputData, so that we don't
+	// have to wait for it to finish before exiting the process.
 	for {
 		gatherAndOutputData(eventf, config, preflightClient, dataGatherers)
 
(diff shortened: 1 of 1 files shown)


pkg/agent/run.go
```diff
@@ -74,7 +74,16 @@ func Run(cmd *cobra.Command, args []string) {
...
 
-	go func() {
+	group, gctx := errgroup.WithContext(ctx)
+	defer func() {
+		// TODO: replace Fatalf log calls with Errorf and return the error
+		cancel()
+		if err := group.Wait(); err != nil {
+			logs.Log.Fatalf("failed to wait for controller-runtime component to stop: %v", err)
+		}
+	}()
```
wallrj, line 84: This code was cut and pasted earlier in the file so that the errgroup can be initialized and used for the management server (`/metrics`, `/healthz`, `/debug/pprof`) and for the VenafiConnection manager; both of which were previously run in unsupervised goroutines.

pkg/agent/run.go
```diff
@@ -200,7 +206,11 @@ func Run(cmd *cobra.Command, args []string) {
 			break
 		}
 
-		time.Sleep(config.Period)
+		select {
+		case <-gctx.Done():
+			return
+		case <-time.After(config.Period):
+		}
```
wallrj, line 213: If any of the errgroup go routines exit (with nil or error) the main context will be cancelled, which will cause this blocking loop to exit immediately instead of waiting for the time period.

pkg/agent/run.go
```diff
@@ -105,21 +114,25 @@ func Run(cmd *cobra.Command, args []string) {
...
 		err := http.ListenAndServe(":8081", server)
 		if err != nil && !errors.Is(err, http.ErrServerClosed) {
-			logs.Log.Fatalf("failed to run the health check server: %s", err)
+			return fmt.Errorf("failed to run the health check server: %s", err)
 		}
-	}()
+		// The agent must stop if the management server stops
+		cancel()
+		return nil
+	})
```
wallrj, line 122: In a future PR we should add a signal handler to catch sigterm (ctrl-c) and cancel the parent context, and add another Go routine which calls `server.Stop` after the context is cancelled.

As it stands, sigterm will cause the process (all threads) to immediately  exit without any graceful shutdown.
 * https://pkg.go.dev/os/signal#hdr-Default_behavior_of_signals_in_Go_programs

pkg/agent/run.go
```diff
@@ -192,15 +198,24 @@ func Run(cmd *cobra.Command, args []string) {
 
 	// begin the datagathering loop, periodically sending data to the
 	// configured output using data in datagatherer caches or refreshing from
-	// APIs each cycle depending on datagatherer implementation
+	// APIs each cycle depending on datagatherer implementation.
+	// If any of the go routines exit (with nil or error) the main context will
+	// be cancelled, which will cause this blocking loop to exit
+	// instead of waiting for the time period.
+	// TODO(wallrj): Pass a context to gatherAndOutputData, so that we don't
+	// have to wait for it to finish before exiting the process.
```
wallrj, line 206: This can be done in another PR.

pkg/agent/run.go
```diff
@@ -105,21 +114,25 @@ func Run(cmd *cobra.Command, args []string) {
 
 		err := http.ListenAndServe(":8081", server)
 		if err != nil && !errors.Is(err, http.ErrServerClosed) {
-			logs.Log.Fatalf("failed to run the health check server: %s", err)
+			return fmt.Errorf("failed to run the health check server: %s", err)
 		}
-	}()
+		// The agent must stop if the management server stops
+		cancel()
```
inteon, line 120: Instead of calling `cancel`, can we return an error when the server is stopped unexpectedly early?
Normally when the context is not canceled, the server should keep running, and when the context is canceled; we do not have to call `cancel`.

inteon: Also, when we return an error, the group context will be automatically canceled.

tfadeyi (approved):
lgtm

In the a future PR it might be good to also move the main server definition in a different function/file, and just have the server's start and stop calls in the go routines.

pkg/agent/run.go
wallrj: I think I agree, that an error should be returned. 
Something like `fmt.Errorf("the API server stopped unexpectedly")`.
But I'll come back to that in another PR.
This `cancel` will work for now.

pkg/agent/run.go
```diff
@@ -160,6 +164,8 @@ func Run(cmd *cobra.Command, args []string) {
 			if err := newDg.Run(gctx.Done()); err != nil {
 				return fmt.Errorf("failed to start %q data gatherer %q: %v", kind, dgConfig.Name, err)
 			}
+			// The agent must stop if any of the data gatherers stops
+			cancel()
```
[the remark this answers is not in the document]
wallrj: Fixed in https://github.com/jetstack/jetstack-secure/pull/605
 * https://github.com/jetstack/jetstack-secure/pull/605

PR #607, opened 2024-11-06
Allow specification of client_id by env var

aidy:
Specifying client_id by env var allows us to pre-populate a namespace with all the relevant authentication details.

Currently, we have to populate a secret with a private key, and then create a deployment given the client id. Specifying via an env var would allow us to create/rotate deployments and credentials independently of each other.

e.g.

```
env:
 - name: CLIENT_ID
   valueFrom:
     configMapKeyRef:
       name: venafi-agent-svc-account
       key: client_id
```


aidy:
Forgot I could just push to `jetstack/` I'll recreate.

PR #608, opened 2024-11-06
Allow specification of client_id by env var

aidy:
Specifying client_id by env var allows us to pre-populate a namespace with all the relevant authentication details.

Currently, we have to populate a secret with a private key, and then create a deployment given the client id. Specifying via an env var would allow us to create/rotate deployments and credentials independently of each other.

e.g.

```
env:
 - name: CLIENT_ID
   valueFrom:
     configMapKeyRef:
       name: venafi-agent-svc-account
       key: client_id
```

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -201,10 +201,10 @@ func InitAgentCmdFlags(c *cobra.Command, cfg *AgentCmdFlags) {
 		&cfg.ClientID,
 		"client-id",
 		"",
-		"",
+		os.Getenv("CLIENT_ID"),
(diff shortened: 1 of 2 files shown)


maelvls:
I wasn't aware of these env variable either. Thanks for pointing them out, Richard. @aidy Does `PREFLIGHT_CLIENT_ID` satisfy your needs?

(We should somehow have these env vars documented somewhere, though)

maelvls:
Hey, I'll close this PR for now.

@aidy let me know if you want to re-open it 👍

PR #611, opened 2024-11-07
[VC-35738] Remove os.Exit from the logs module

wallrj:
Instead of calling os.exit directly after parsing and finding an error in the logging flags, the error is now returned and handled the same way as errors that come from the `Run` function....they are logged using structured logging and the logs are flushed before the process exits.

I've adjusted the logs tests accordingly and modified them so that the log flag parsing and error handling in the tests matches the behaviour of the actual command.

Before:
```console
$ ./preflight agent --logging-format=foo
Error in logging configuration: format: Invalid value: "foo": Unsupported log format

$ echo $?
2
```

After:
```console
$ ./preflight agent --logging-format=foo
E1107 13:41:31.063915  281828 root.go:51] "Exiting due to error" err="Error in logging configuration: format: Invalid value: \"foo\": Unsupported log format" exit-code=1

$ echo $?
1
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -20,8 +20,8 @@ var rootCmd = &cobra.Command{
 configuration checks using Open Policy Agent (OPA).
 
 Preflight checks are bundled into Packages`,
-	PersistentPreRun: func(cmd *cobra.Command, args []string) {
-		logs.Initialize()
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		return logs.Initialize()
 	},
 	// SilenceErrors and SilenceUsage prevents this command or any sub-command
 	// from printing arbitrary text to stderr.
--- a/pkg/logs/logs_test.go
+++ b/pkg/logs/logs_test.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
+	"io"
 	"log"
 	"log/slog"
 	"os"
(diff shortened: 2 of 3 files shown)


cmd/root.go
```diff
@@ -20,8 +20,8 @@ var rootCmd = &cobra.Command{
 configuration checks using Open Policy Agent (OPA).
 
 Preflight checks are bundled into Packages`,
-	PersistentPreRun: func(cmd *cobra.Command, args []string) {
-		logs.Initialize()
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		return logs.Initialize()
 	},
```
wallrj, line 25: Returning the error here means that the error can be logged in the same format as all the other errors that are returned by the command.

pkg/logs/logs_test.go
```diff
@@ -33,10 +35,25 @@ import (
...
+					klog.ErrorS(err, "Exiting due to error", "exit-code", exitCode)
+					klog.FlushAndExit(time.Second, exitCode)
+				}
+			}
+			if err := logs.Initialize(); err != nil {
+				exitCode := 1
+				klog.ErrorS(err, "Exiting due to error", "exit-code", exitCode)
+				klog.FlushAndExit(time.Second, exitCode)
+			}
 		}
```
wallrj, line 57: These changes are intended to make the logs test fixture parse the flags and behave the same way as Cobra.
* --help / -h results in stdout and exit code 0
* Unrecognized flags result in an error which is logged to stderr with exit code 1

maelvls: Thanks for the extra context. I think it would be worth adding a comment above to keep that in mind in the future. I'm already imagining myself in a year: why do we do this weird dance with the exit codes in the test code?

maelvls (approved):
This sounds like a good change. I haven't tested the change manually nor have I run the automated tests, I trust you.

pkg/logs/logs_test.go
wallrj: I'll merge this now and add the comment in one of the followup PRS.

PR #614, opened 2024-11-08
VC-36950: It is now possible to exclude labels and annotations

maelvls:
Some annotations and labels contain sensitive information. A well-known example is the `kubectl.kubernetes.io/last-applied-configuration` annotation. We already exclude this annotations from Secret resources, but we have found that customers use other sensitive annotations and labels that they would like to not be sent to the Venafi Control Plane API.

A realistic example lies in the Kapp project. The Kapp project uses four annotations that all start with `kapp.k14s.io/original*` as seen in [1]. These annotations are similar to `kubectl.kubernetes.io/last-applied-configuration` in that they may contain sensitive information. The annotations would look like this:

```yaml
annotations:
  kapp.k14s.io/original: |
    {"apiVersion":"v1","kind":"Secret","spec":{"data": {"password": "cGFzc3dvcmQ=","username": "bXl1c2VybmFtZQ=="}}}
  kapp.k14s.io/original-diff: |
    - type: test
      path: /data
      value:
      password: cygpcGVyUzNjcmV0UEBhc3N3b3JkIQ==
      username: bXl1c2VybmFtZQ==
```

[1]: https://github.com/carvel-dev/kapp/issues/90#issuecomment-602074356

In this case, customers will be suggested to use the following exclusion setting in `values.yaml` to exclude both `kapp.k14s.io/original` and `kapp.k14s.io/original-diff`:

```yaml
config:
  excludeAnnotationKeysRegex:
    - ^kapp\.k14s\.io/original
```

For labels, it looks like this:

```yaml
config:
  excludeLabelKeysRegex:
    - \.company\.com/
```

The documentation for this feature is being worked on in https://gitlab.com/venafi/vaas/ua/clouddocs/-/merge_requests/1220.

**Ref:** [VC-36950](https://venafi.atlassian.net/browse/VC-36950 "Step 2: Add the regex logic and expose the config fields excludeAnnotationKeysRegex and excludeLabelKeysRegex")

## Why regex, why not fixed string search with wildcards?

We have chosen to go with regexes instead of fixed-string search or wildcards. We know that using regexes will be confusing, but we don't expect this feature to be used so often, and since we don't know exactly the type of filtering that customers will want, we have decided to use the most versatile one : regexes.

## Manual Testing

Although I've added a couple of unit tests, I haven't had time to add an automated end-to-end test. In particular, my PR is **lacking a smoke test to check that the Helm chart value `excludeAnnotationKeysRegex` does what it claims it does**.

Here is the two manual tests that make me confident that this feature works as expected.

### Test 1: Using `--output-path` + fake Jetstack Secure API token

I used an empty Kind cluster for this test.

First, run this:

```bash
cat <<EOF >minimal-config.yaml
period: 5m
cluster_id: "kind-mael"
organization_id: foo
venafi-cloud:
exclude-annotation-keys-regex:
  - ".*password.*"
  - ".*apikey.*"
exclude-label-keys-regex:
  - ".*password.*"
  - ".*apikey.*"
data-gatherers:
  - kind: "k8s-dynamic"
    name: "k8s/namespaces"
    config:
      resource-type:
        resource: namespaces
        version: v1
  - kind: "k8s-dynamic"
    name: "k8s/secrets"
    config:
      resource-type:
        version: v1
        resource: secrets
EOF
```

Then:

```sh
openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout tls.key -out tls.crt -subj "/CN=mydomain.com/O=myorganization"
kubectl create secret tls my-tls-secret --cert=tls.crt --key=tls.key
kubectl annotate secret my-tls-secret some-secret-apikey=1234
kubectl label secret my-tls-secret some-secret-password=1234

kubectl create ns my-ns
kubectl annotate ns my-ns some-secret-apikey=1234
kubectl label ns my-ns some-secret-password=1234
```

Finally, run the agent in one-shot mode. You should see no output when running this:

```console
$ go run . agent -c minimal-config.yaml \
  --api-token should-not-be-required \
  --install-namespace=venafi \
  --output-path /dev/stdout \
  --one-shot | grep -B5 1234
2024/11/08 16:46:56 Preflight agent version: development ()
2024/11/08 16:46:56 Using the Jetstack Secure API Token auth mode since --api-token was specified.
2024/11/08 16:46:56 error messages will not show in the pod's events because the POD_NAME environment variable is empty
2024/11/08 16:46:56 starting "k8s/namespaces" datagatherer
2024/11/08 16:46:56 starting "k8s/secrets" datagatherer
2024/11/08 16:46:56 successfully gathered 7 items from "k8s/namespaces" datagatherer
2024/11/08 16:46:56 successfully gathered 5 items from "k8s/secrets" datagatherer
2024/11/08 16:46:56 Data saved to local file: /dev/stdout
```

Before this PR, the output would look like this:

```console
$ go run . agent -c minimal-config.yaml \
  --api-token should-not-be-required \
  --install-namespace=venafi \
  --output-path /dev/stdout \
  --one-shot | grep -B5 1234
2024/11/08 16:47:26 Preflight agent version: development ()
2024/11/08 16:47:26 Using the Jetstack Secure API Token auth mode since --api-token was specified.
2024/11/08 16:47:26 error messages will not show in the pod's events because the POD_NAME environment variable is empty
2024/11/08 16:47:26 starting "k8s/namespaces" datagatherer
2024/11/08 16:47:26 starting "k8s/secrets" datagatherer
2024/11/08 16:47:26 successfully gathered 7 items from "k8s/namespaces" datagatherer
2024/11/08 16:47:26 successfully gathered 5 items from "k8s/secrets" datagatherer
2024/11/08 16:47:26 Data saved to local file: /dev/stdout
          "resource": {
            "apiVersion": "v1",
            "kind": "Namespace",
            "metadata": {
              "annotations": {
                "some-secret-apikey": "1234"
              },
              "creationTimestamp": "2024-11-08T15:37:03Z",
              "labels": {
                "kubernetes.io/metadata.name": "my-ns",
                "some-secret-password": "1234"
--
              "tls.crt": "..."
            },
            "kind": "Secret",
            "metadata": {
              "annotations": {
                "some-secret-apikey": "1234"
              },
              "labels": {
                "some-secret-password": "1234"
```

### Test 2: Real API + Venafi Cloud Key Pair Service Account auth

In this test, I check that agent's HTTP request doesn't contain the annotations and labels using mitmproxy.

For this test, I've used the tenant https://ven-tlspk.venafi.cloud/. To access the API key, use the user `system.admin@tlspk.qa.venafi.io` and the password is visible in the page [Production Accounts](https://venafi.atlassian.net/wiki/spaces/CT/pages/2115404149) (private to Venafi). Then go to the settings and find the API key, and set it as an env var:

```sh
APIKEY=...
```

Create the service account and key pair:

```bash
venctl iam service-account agent create --name "$USER temp" \
  --vcp-region US \
  --output json \
  --owning-team $(curl -sS https://api.venafi.cloud/v1/teams -H "tppl-api-key: $APIKEY" | jq '.teams[0].id') \
  --api-key $APIKEY
```

Now, make sure to have `127.0.0.1 me` in your `/etc/hosts`.

Then, run mitmproxy with:

```sh
mitmproxy --mode regular@9090 --ssl-insecure -s /tmp/watch-stream.py --set client_certs=$(kubectl incluster --print-client-cert >/tmp/me.pem && echo /tmp/me.pem)
```

Run this:

```bash
cat <<EOF >minimal-config.yaml
period: 5m
cluster_id: "kind-mael"
venafi-cloud:
exclude-annotation-keys-regex:
  - ".*password.*"
  - ".*apikey.*"
exclude-label-keys-regex:
  - ".*password.*"
  - ".*apikey.*"
data-gatherers:
  - kind: "k8s-dynamic"
    name: "k8s/namespaces"
    config:
      resource-type:
        resource: namespaces
        version: v1
  - kind: "k8s-dynamic"
    name: "k8s/secrets"
    config:
      resource-type:
        version: v1
        resource: secrets
EOF
```

Now, don't forget to create a Kind cluster. Then:

```sh
openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout tls.key -out tls.crt -subj "/CN=mydomain.com/O=myorganization"
kubectl create secret tls my-tls-secret --cert=tls.crt --key=tls.key
kubectl annotate secret my-tls-secret some-secret-apikey=1234
kubectl label secret my-tls-secret some-secret-password=1234

kubectl create ns my-ns
kubectl annotate ns my-ns some-secret-apikey=1234
kubectl label ns my-ns some-secret-password=1234
```

Finally, run the Agent with:

```sh
go install github.com/maelvls/kubectl-incluster@latest
export HTTPS_PROXY=http://localhost:9090 KUBECONFIG=/tmp/kube && KUBECONFIG= HTTPS_PROXY= kubectl incluster --replace-ca-cert ~/.mitmproxy/mitmproxy-ca-cert.pem --sa=venafi/venafi-kubernetes-agent | sed 's|127.0.0.1|me|' >/tmp/kube

go run . agent -c minimal-config.yaml \
  --client-id $(jq -r .client_id /tmp/agent-credentials.json) \
  --private-key-path <(jq -r .private_key /tmp/agent-credentials.json) \
  --install-namespace=venafi \
  --one-shot
```

Look at the mitmproxy logs and look for the entry that has the path

```
/v1/tlspk/upload/clusterdata/no?name=...
```

Open it (enter) and type `/` and 1234. You should not see any instance of "1234", proving that the annotations and labels have been filtered out.

When looking at the JSON blob, all of the annotations and labels containing `apikey` and `password` are gone:

| type | name | namespace | annotations | labels |
|------|------|-----------|-------------|--------|
| Namespace | local-path-storage | - | {} | {“kubernetes.io/metadata.name”: “local-path-storage”} |
| Namespace | my-ns | - | {} | {“kubernetes.io/metadata.name”: “my-ns”} |
| Namespace | venafi | - | {} | {“kubernetes.io/metadata.name”: “venafi”, “name”: “venafi”} |
| Namespace | default | - | {} | {“kubernetes.io/metadata.name”: “default”} |
| Namespace | kube-node-lease | - | {} | {“kubernetes.io/metadata.name”: “kube-node-lease”} |
| Namespace | kube-public | - | {} | {“kubernetes.io/metadata.name”: “kube-public”} |
| Namespace | kube-system | - | {} | {“kubernetes.io/metadata.name”: “kube-system”} |
| Secret | agent-credentials | venafi | {} | - |
| Secret | sh.helm.release.v1.venafi-kubernetes-agent.v1 | venafi | - | {“modifiedAt”: “1728905433”, “name”: “venafi-kubernetes-agent”, “owner”: “helm”,  |“status”: “superseded”, “version”: “1”}
| Secret | sh.helm.release.v1.venafi-kubernetes-agent.v2 | venafi | - | {“modifiedAt”: “1728905433”, “name”: “venafi-kubernetes-agent”, “owner”: “helm”,  |“status”: “superseded”, “version”: “2”}
| Secret | sh.helm.release.v1.venafi-kubernetes-agent.v3 | venafi | - | {“modifiedAt”: “1728905433”, “name”: “venafi-kubernetes-agent”, “owner”: “helm”,  |“status”: “deployed”, “version”: “3”}
| Secret | my-tls-secret | default | {} | {} |

Here is the output that I got without this PR:

| type | name | namespace | annotations | labels |
|------|------|-----------|-------------|--------|
| Namespace | default | - | - | {“kubernetes.io/metadata.name”: “default”} |
| Namespace | kube-node-lease | - | - | {“kubernetes.io/metadata.name”: “kube-node-lease”} |
| Namespace | kube-public | - | - | {“kubernetes.io/metadata.name”: “kube-public”} |
| Namespace | kube-system | - | - | {“kubernetes.io/metadata.name”: “kube-system”} |
| Namespace | local-path-storage | - | {} | {“kubernetes.io/metadata.name”: “local-path-storage”} |
| Namespace | my-ns | - | {“some-secret-apikey”: “1234”} | {“kubernetes.io/metadata.name”: “my-ns”, “some-secret-password”: “1234”} |
| Namespace | venafi | - | - | {“kubernetes.io/metadata.name”: “venafi”, “name”: “venafi”} |
| Secret | agent-credentials | venafi | {} | - |
| Secret | sh.helm.release.v1.venafi-kubernetes-agent.v1 | venafi | - | {“modifiedAt”: “1728905433”, “name”: “venafi-kubernetes-agent”, “owner”: “helm”,  |“status”: “superseded”, “version”: “1”}
| Secret | sh.helm.release.v1.venafi-kubernetes-agent.v2 | venafi | - | {“modifiedAt”: “1728905433”, “name”: “venafi-kubernetes-agent”, “owner”: “helm”,  |“status”: “superseded”, “version”: “2”}
| Secret | sh.helm.release.v1.venafi-kubernetes-agent.v3 | venafi | - | {“modifiedAt”: “1728905433”, “name”: “venafi-kubernetes-agent”, “owner”: “helm”,  |“status”: “deployed”, “version”: “3”}
| Secret | my-tls-secret | default | {“some-secret-apikey”: “1234”} | {“some-secret-password”: “1234”} |

[VC-36950]: https://venafi.atlassian.net/browse/VC-36950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deploy/charts/venafi-kubernetes-agent/values.yaml
+++ b/deploy/charts/venafi-kubernetes-agent/values.yaml
@@ -238,6 +238,23 @@ config:
   - bootstrap.kubernetes.io/token
   - helm.sh/release.v1
 
+  # You can configure Venafi Kubernetes Agent to exclude some annotations or
+  # labels from being pushed to the Venafi Control Plane. All Kubernetes objects
+  # are affected. The objects are still pushed, but the specified annotations
+  # and labels are removed before being sent to the Venafi Control Plane.
+  #
+  # If you would like to exclude annotations keys that contain the word
+  # `secret`, use the regular expression `.*secret.*`. The leading and ending .*
+  # are important if you want to filter out keys that contain `secret` anywhere
+  # in the key string.
+  #
+  # Note that the annotation `kubectl.kubernetes.io/last-applied-configuration`
+  # is already excluded by default, you don't need to exclude it explicitly.
+  #
+  # Example: excludeAnnotationKeysRegex: [".*secret.*"]
+  excludeAnnotationKeysRegex: []
+  excludeLabelKeysRegex: []
+
   # Specify ConfigMap details to load config from an existing resource.
   # This should be blank by default unless you have you own config.
   configmap:
--- a/deploy/charts/venafi-kubernetes-agent/values.yaml
+++ b/deploy/charts/venafi-kubernetes-agent/values.yaml
@@ -114,7 +114,7 @@ podSecurityContext: {}
 securityContext:
   capabilities:
     drop:
-    - ALL
+      - ALL
   readOnlyRootFilesystem: true
   runAsNonRoot: true
 
@@ -230,28 +230,24 @@ config:
   # * https://kubernetes.io/docs/concepts/configuration/secret/#secret-types
   # * https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/#list-of-supported-fields
   ignoredSecretTypes:
-  - kubernetes.io/service-account-token
-  - kubernetes.io/dockercfg
-  - kubernetes.io/dockerconfigjson
-  - kubernetes.io/basic-auth
-  - kubernetes.io/ssh-auth
-  - bootstrap.kubernetes.io/token
-  - helm.sh/release.v1
+    - kubernetes.io/service-account-token
+    - kubernetes.io/dockercfg
+    - kubernetes.io/dockerconfigjson
+    - kubernetes.io/basic-auth
+    - kubernetes.io/ssh-auth
+    - bootstrap.kubernetes.io/token
+    - helm.sh/release.v1
 
   # You can configure Venafi Kubernetes Agent to exclude some annotations or
   # labels from being pushed to the Venafi Control Plane. All Kubernetes objects
   # are affected. The objects are still pushed, but the specified annotations
   # and labels are removed before being sent to the Venafi Control Plane.
   #
-  # If you would like to exclude annotations keys that contain the word `word`,
-  # use the regular expression `.*word.*`. The leading and ending .* are
-  # important if you want to filter out keys that contain `word` anywhere in the
-  # key string.
+  # Dots is the only character that needs to be escaped in the regex. Use either
+  # double quotes with escaped single quotes or unquoted strings for the regex
+  # to avoid YAML parsing issues with `\.`.
   #
-  # Note that the annotation `kubectl.kubernetes.io/last-applied-configuration`
-  # is already excluded by default, you don't need to exclude it explicitly.
-  #
-  # Example: excludeAnnotationKeysRegex: ["kapp\.k14s\.io\/original.*"]
+  # Example: excludeAnnotationKeysRegex: ['^kapp\.k14s\.io/original.*']
   excludeAnnotationKeysRegex: []
   excludeLabelKeysRegex: []
 
--- a/deploy/charts/venafi-kubernetes-agent/values.yaml
+++ b/deploy/charts/venafi-kubernetes-agent/values.yaml
@@ -243,15 +243,15 @@ config:
   # are affected. The objects are still pushed, but the specified annotations
   # and labels are removed before being sent to the Venafi Control Plane.
   #
-  # If you would like to exclude annotations keys that contain the word
-  # `secret`, use the regular expression `.*secret.*`. The leading and ending .*
-  # are important if you want to filter out keys that contain `secret` anywhere
-  # in the key string.
+  # If you would like to exclude annotations keys that contain the word `word`,
+  # use the regular expression `.*word.*`. The leading and ending .* are
+  # important if you want to filter out keys that contain `word` anywhere in the
+  # key string.
   #
   # Note that the annotation `kubectl.kubernetes.io/last-applied-configuration`
   # is already excluded by default, you don't need to exclude it explicitly.
   #
-  # Example: excludeAnnotationKeysRegex: [".*secret.*"]
+  # Example: excludeAnnotationKeysRegex: ["kapp\.k14s\.io\/original.*"]
   excludeAnnotationKeysRegex: []
   excludeLabelKeysRegex: []
 
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -34,6 +34,7 @@ import (
 	"github.com/jetstack/preflight/api"
 	"github.com/jetstack/preflight/pkg/client"
 	"github.com/jetstack/preflight/pkg/datagatherer"
+	"github.com/jetstack/preflight/pkg/datagatherer/k8s"
 	"github.com/jetstack/preflight/pkg/kubeconfig"
 	"github.com/jetstack/preflight/pkg/logs"
 	"github.com/jetstack/preflight/pkg/version"
@@ -174,6 +175,12 @@ func Run(cmd *cobra.Command, args []string) (returnErr error) {
 		newDg, err := dgConfig.Config.NewDataGatherer(gctx)
 		if err != nil {
 			return fmt.Errorf("failed to instantiate %q data gatherer  %q: %v", kind, dgConfig.Name, err)
+		}
+
+		dynDg, isDynamicGatherer := newDg.(*k8s.DataGathererDynamic)
+		if isDynamicGatherer {
+			dynDg.ExcludeAnnotKeys = config.ExcludeAnnotationKeysRegex
+			dynDg.ExcludeLabelKeys = config.ExcludeLabelKeysRegex
 		}
 
 		log.V(logs.Debug).Info("Starting DataGatherer", "name", dgConfig.Name)
--- a/pkg/datagatherer/k8s/dynamic.go
+++ b/pkg/datagatherer/k8s/dynamic.go
@@ -3,6 +3,7 @@ package k8s
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
@@ -260,6 +261,9 @@ type DataGathererDynamic struct {
 	// informer watches the events around the targeted resource and updates the cache
 	informer     k8scache.SharedIndexInformer
 	registration k8scache.ResourceEventHandlerRegistration
+
+	ExcludeAnnotKeys []*regexp.Regexp
+	ExcludeLabelKeys []*regexp.Regexp
 }
 
 // Run starts the dynamic data gatherer's informers for resource collection.
@@ -338,7 +342,7 @@ func (g *DataGathererDynamic) Fetch() (interface{}, int, error) {
 	}
 
 	// Redact Secret data
-	err := redactList(items)
+	err := redactList(items, g.ExcludeAnnotKeys, g.ExcludeLabelKeys)
 	if err != nil {
 		return nil, -1, errors.WithStack(err)
 	}
@@ -349,7 +353,7 @@ func (g *DataGathererDynamic) Fetch() (interface{}, int, error) {
 	return list, len(items), nil
 }
 
-func redactList(list []*api.GatheredResource) error {
+func redactList(list []*api.GatheredResource, excludeAnnotKeys, excludeLabelKeys []*regexp.Regexp) error {
 	for i := range list {
 		if item, ok := list[i].Resource.(*unstructured.Unstructured); ok {
 			// Determine the kind of items in case this is a generic 'mixed' list.
@@ -374,6 +378,43 @@ func redactList(list []*api.GatheredResource) error {
 
 			// remove managedFields from all resources
 			Redact(RedactFields, resource)
+
+			annotsRaw, ok, err := unstructured.NestedFieldNoCopy(resource.Object, "metadata", "annotations")
+			if err != nil {
+				return fmt.Errorf("wasn't able to find the metadata.annotations field: %w", err)
+			}
+			if ok {
+				annots, ok := annotsRaw.(map[string]interface{})
+				if !ok {
+					return fmt.Errorf("metadata.annotations isn't a map on the resource %s in namespace %s: %w", resource.GetName(), resource.GetNamespace(), err)
+				}
+				for key := range annots {
+					for _, excludeAnnotKey := range excludeAnnotKeys {
+						if excludeAnnotKey.MatchString(key) {
+							delete(annots, key)
+						}
+					}
+				}
+			}
+
+			labelsRaw, ok, err := unstructured.NestedFieldNoCopy(resource.Object, "metadata", "labels")
+			if err != nil {
+				return fmt.Errorf("wasn't able to find the metadata.labels field for the resource %s in namespace %s: %w", resource.GetName(), resource.GetNamespace(), err)
+			}
+			if ok {
+				labels, ok := labelsRaw.(map[string]interface{})
+				if !ok {
+					return fmt.Errorf("metadata.labels isn't a map on the resource %s in namespace %s: %w", resource.GetName(), resource.GetNamespace(), err)
+				}
+				for key := range labels {
+					for _, excludeLabelKey := range excludeLabelKeys {
+						if excludeLabelKey.MatchString(key) {
+							delete(labels, key)
+						}
+					}
+				}
+			}
+
 			continue
 		}
 
@@ -385,6 +426,24 @@ func redactList(list []*api.GatheredResource) error {
 		if item, ok := list[i].Resource.(objectMeta); ok {
 			item.GetObjectMeta().SetManagedFields(nil)
 			delete(item.GetObjectMeta().GetAnnotations(), "kubectl.kubernetes.io/last-applied-configuration")
+
+			annots := item.GetObjectMeta().GetAnnotations()
+			for key := range annots {
+				for _, excludeAnnotKey := range excludeAnnotKeys {
+					if excludeAnnotKey.MatchString(key) {
+						delete(annots, key)
+					}
+				}
+			}
+
+			labels := item.GetObjectMeta().GetLabels()
+			for key := range labels {
+				for _, excludeLabelKey := range excludeLabelKeys {
+					if excludeLabelKey.MatchString(key) {
+						delete(labels, key)
+					}
+				}
+			}
 
 			resource := item.(runtime.Object)
 			gvks, _, err := scheme.Scheme.ObjectKinds(resource)
--- a/pkg/datagatherer/k8s/dynamic.go
+++ b/pkg/datagatherer/k8s/dynamic.go
@@ -379,41 +379,8 @@ func redactList(list []*api.GatheredResource, excludeAnnotKeys, excludeLabelKeys
 			// remove managedFields from all resources
 			Redact(RedactFields, resource)
 
-			annotsRaw, ok, err := unstructured.NestedFieldNoCopy(resource.Object, "metadata", "annotations")
-			if err != nil {
-				return fmt.Errorf("wasn't able to find the metadata.annotations field: %w", err)
-			}
-			if ok {
-				annots, ok := annotsRaw.(map[string]interface{})
-				if !ok {
-					return fmt.Errorf("metadata.annotations isn't a map on the resource %s in namespace %s: %w", resource.GetName(), resource.GetNamespace(), err)
-				}
-				for key := range annots {
-					for _, excludeAnnotKey := range excludeAnnotKeys {
-						if excludeAnnotKey.MatchString(key) {
-							delete(annots, key)
-						}
-					}
-				}
-			}
-
-			labelsRaw, ok, err := unstructured.NestedFieldNoCopy(resource.Object, "metadata", "labels")
-			if err != nil {
-				return fmt.Errorf("wasn't able to find the metadata.labels field for the resource %s in namespace %s: %w", resource.GetName(), resource.GetNamespace(), err)
-			}
-			if ok {
-				labels, ok := labelsRaw.(map[string]interface{})
-				if !ok {
-					return fmt.Errorf("metadata.labels isn't a map on the resource %s in namespace %s: %w", resource.GetName(), resource.GetNamespace(), err)
-				}
-				for key := range labels {
-					for _, excludeLabelKey := range excludeLabelKeys {
-						if excludeLabelKey.MatchString(key) {
-							delete(labels, key)
-						}
-					}
-				}
-			}
+			RemoveUnstructuredKeys(excludeAnnotKeys, resource, "metadata", "annotations")
+			RemoveUnstructuredKeys(excludeLabelKeys, resource, "metadata", "labels")
 
 			continue
 		}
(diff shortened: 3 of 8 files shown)


pkg/agent/run.go
```diff
@@ -154,6 +155,12 @@ func Run(cmd *cobra.Command, args []string) {
 			logs.Log.Fatalf("failed to instantiate %q data gatherer  %q: %v", kind, dgConfig.Name, err)
 		}
 
+		dynDg, isDynamicGatherer := newDg.(*k8s.DataGathererDynamic)
+		if isDynamicGatherer {
+			dynDg.ExcludeAnnotKeys = config.ExcludeAnnotationKeysRegex
+			dynDg.ExcludeLabelKeys = config.ExcludeLabelKeysRegex
+		}
```
maelvls, line 162: This is super hacky... I tried finding a cleaner way of "injecting" the ExcludeAnnotKeys and ExcludeLabelKeys, but gave up.

deploy/charts/venafi-kubernetes-agent/values.yaml
```diff
@@ -238,6 +238,23 @@ config:
...
 
+  # You can configure Venafi Kubernetes Agent to exclude some annotations or
+  # labels from being pushed to the Venafi Control Plane. All Kubernetes objects
+  # are affected. The objects are still pushed, but the specified annotations
+  # and labels are removed before being sent to the Venafi Control Plane.
+  #
+  # If you would like to exclude annotations keys that contain the word
+  # `secret`, use the regular expression `.*secret.*`. The leading and ending .*
+  # are important if you want to filter out keys that contain `secret` anywhere
+  # in the key string.
```
wallrj, line 249:
```suggestion
  # If you would like to exclude annotation keys that contain the word
  # `secret`, use the regular expression `.*secret.*`. The leading and ending .*
  # are important if you want to filter out keys that contain `secret` anywhere
  # in the key string.
```

pkg/datagatherer/k8s/dynamic.go
```diff
@@ -371,6 +375,43 @@ func redactList(list []*api.GatheredResource) error {
 
 			// remove managedFields from all resources
 			Redact(RedactFields, resource)
+
+			annotsRaw, ok, err := unstructured.NestedFieldNoCopy(resource.Object, "metadata", "annotations")
+			if err != nil {
+				return fmt.Errorf("wasn't able to find the metadata.annotations field: %w", err)
+			}
```
wallrj, line 382: I don't think this should return an error, because the object might not have any annotations. I guess it depends whether the `metadata.annotations` field has an `omitempty` marker.

wallrj (changes requested):
@maelvls I read the code but didn't test it yet.
1. The unit tests are failing
2. The example regex patterns don't seem very realistic. Why would a user put "secrets" in the annotations of labels?
3. And if it is realistic, shouldn't the venafi-kubernetes-agent redact  labels containing those terms by default...like a log sanitizer or like GitHub actions attempts to do to prevent accidentally leaking these things.
4. The original Jira ticket says: "Note: sensitive data could be something like organizational structure or team ownership of a resource, which could be defined through labels." but does not give any real world examples. I imagine something like `^.*\.example\.com/.*` to filter out all the labels with the well known organization prefix.

maelvls:
@wallrj FYI, the documentation for this feature is being worked on in https://gitlab.com/venafi/vaas/ua/clouddocs/-/merge_requests/1220.

maelvls:
I've fixed the unit tests. The issue was each test case focused on a single data gatherer, and one data gatherer only ever deals with one resource at a time. And the test case in question was set up with the `Secret` resource. I could have created a different test case... But I've decided to remove the `Route` resource altogether since it doesn't increase my level of confidence.

The tests are now passing. You can go ahead and try the feature out. Thanks! @wallrj

maelvls:
> 2. The example regex patterns don't seem very realistic. Why would a user put "secrets" in the annotations of labels?

You are right. 

To give a realistic regex in this test case, let’s use the example of the Kapp project that uses four annotations that all start with `kapp.k14s.io/original*`. These annotations are similar to `kubectl.kubernetes.io/last-applied-configuration` in that they may contain sensitive information. From [1], they may look like this:

```yaml
annotations:
  kapp.k14s.io/original: |
    {"apiVersion":"v1","kind":"Secret","spec":{"data": {"password": "cGFzc3dvcmQ=","username": "bXl1c2VybmFtZQ=="}}}
  kapp.k14s.io/original-diff: |
    - type: test
      path: /data
      value:
      password: cygpcGVyUzNjcmV0UEBhc3N3b3JkIQ==
      username: bXl1c2VybmFtZQ==
```

[1]: https://github.com/carvel-dev/kapp/issues/90#issuecomment-602074356

In this case, I'd suggest using the following exclusion setting in `values.yaml`:

```yaml
excludeAnnotationKeysRegex:
  - ".*kapp\.k14s\.io\/original.*"
```

> 3. And if it is realistic, shouldn't the venafi-kubernetes-agent redact  labels containing those terms by default...like a log sanitizer or like GitHub actions attempts to do to prevent accidentally leaking these things.

Good point, that'd be perfect. I don't know how to get there though, so until we get there, I like the idea of providing a way of letting customers exclude specific annotations.

> 4. The original Jira ticket says: "Note: sensitive data could be something like organizational structure or team ownership of a resource, which could be **defined through labels**." but does not give any real world examples. I imagine something like `^.*\.example\.com/.*` to filter out all the labels with the well known organization prefix.

I think your example is sensible, and I'll change my examples in code and documentation to match what you proposed. The `values.yaml` would look like this:

```yaml
config:
  excludeLabelKeysRegex:
    - ^.*\.company\.com/.*
```

It reminds me that I need to clarify a few things in the documentation:

- Which characters need to be escaped. It seems like `.` is the only character that people should worry about since annotation keys and label keys must be of the form `[prefix/]name`, where prefix is a DNS subdomain and name is a DNS label.
- Which characters don't need to be escaped (notably, `/` doesn't need to be escaped).

pkg/datagatherer/k8s/dynamic_test.go
```diff
@@ -582,19 +598,55 @@ func TestDynamicGatherer_Fetch(t *testing.T) {
...
+			//    {"apiVersion":"v1","kind":"Secret","spec":{"data": {"password": "cGFzc3dvcmQ=","username": "bXl1c2VybmFtZQ=="}}}
+			//  kapp.k14s.io/original-diff: |
+			//    - type: test
+			//      path: /data
+			//      value:
+			//      password: cygpcGVyUzNjcmV0UEBhc3N3b3JkIQ==
+			//      username: bXl1c2VybmFtZQ==
+			//
+			//  [1]: https://github.com/carvel-dev/kapp/issues/90#issuecomment-602074356
+			excludeAnnotsKeys: []string{`kapp\.k14s\.io\/original.*`},
```
maelvls, line 621: I don't know why I escaped the character `/`. I'll un-escape it.

```suggestion
			excludeAnnotsKeys: []string{`kapp\.k14s\.io/original.*`},
```

maelvls: I also forgot to use `^`. I'll add it too.

```suggestion
			excludeAnnotsKeys: []string{`^kapp\.k14s\.io\/original.*`},
```

wallrj (changes requested):
@maelvls I tested it locally and it does strip the labels and annotations from the metdata of objects. 
But it misses the labels and annotations that are added to PodTemplate in Deployment, StatefulSet, Job etc.
So I don't think it's going to satisfy users who want to prevent the sharing of sensitive labels and annotations.

Happy to approve this if you consider it to be a step in the right direction, but I think a general solution which scrubs all sensitive fields, will look different to this.  
I suppose it might be applied to all keys in the venafi-kubernetes-agent JSON output,
regardless of which DataGatherers have been used.

```
helm template deploy/charts/venafi-kubernetes-agent \
  --show-only templates/configmap.yaml \
  --set fullnameOverride=venafi-kubernetes-agent \
  --set config.excludeAnnotationKeysRegex={'.*'} \
  --set config.excludeLabelKeysRegex={'.*'} \
| yq '.data."config.yaml" \
        | @yamld \
        | .organization_id |= "example.com" \
        | .cluster_id |= "cluster-1.example.com"' \
> examples/venafi-kubernetes-agent.yaml

go run ./ agent \
  --one-shot \
  --api-token should-not-be-required \
  --install-namespace venafi \
  --output-path /dev/stdout \
  --agent-config-file examples/venafi-kubernetes-agent.yaml 2>/dev/null \
|  grep -E -C 5 -e '"(labels|annotations)": {$'
```

```json
...
--
                }
              },
              "template": {
                "metadata": {
                  "creationTimestamp": null,
                  "labels": {
                    "app.kubernetes.io/instance": "venafi-enhanced-issuer",
                    "app.kubernetes.io/managed-by": "Helm",
                    "app.kubernetes.io/name": "venafi-enhanced-issuer",
                    "app.kubernetes.io/version": "v0.14.0",
                    "helm.sh/chart": "venafi-enhanced-issuer-v0.14.0",
                    "pod-template-hash": "d7496dc68"
                  },
                  "annotations": {
                    "kubectl.kubernetes.io/default-container": "venafi-enhanced-issuer"
                  }
                },
                "spec": {
                  "containers": [
```

pkg/datagatherer/k8s/dynamic.go
maelvls: You are right. I will rework this and add more unit tests to feel a little more confident.

maelvls:
> But it misses the labels and annotations that are added to PodTemplate in Deployment, StatefulSet, Job etc.

I totally forgot about these. I think it's worth investigating this further in a separate PR, but I'd be OK shipping the next version of Venafi Kubernetes Agent with the feature as presented in this PR.

pkg/datagatherer/k8s/dynamic_test.go
maelvls: Done.

pkg/datagatherer/k8s/dynamic_test.go
```diff
@@ -1009,3 +1086,200 @@ func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
...
+				"annotations": map[string]interface{}{
+					"key1": "value1",
+					"key2": "value2",
+				},
+			},
+		},
+		expectObj: map[string]interface{}{
+			"metadata": map[string]interface{}{
+				"name": "foo",
+			},
```
wallrj, line 1106: Shouldn't `key2` remain ?
```suggestion
			"metadata": map[string]interface{}{
				"name": "foo",
				"annotations": map[string]interface{}{
					"key2": "value2",
				},
			},
```

wallrj (changes requested):

pkg/datagatherer/k8s/dynamic_test.go
```diff
@@ -1009,3 +1086,200 @@ func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
...
+	givenExclude []string
+	givenObj     map[string]interface{}
+	givenPath    []string
+	expectObj    map[string]interface{}
+}
+
+func run_TestRemoveUnstructuredKeys(tc tc_RemoveUnstructuredKeys) func(*testing.T) {
+	return func(t *testing.T) {
+		t.Helper()
+		RemoveUnstructuredKeys(toRegexps(tc.givenExclude), &unstructured.Unstructured{Object: tc.givenObj}, tc.givenPath...)
```
wallrj, line 1230:
```suggestion
		RemoveUnstructuredKeys(toRegexps(tc.givenExclude), &unstructured.Unstructured{Object: tc.givenObj}, tc.givenPath...)
		assert.Equal(t, tc.expectObj, tc.givenObj)
```

wallrj: After adding the missing assertion the test cases needed some adjustment.

```diff
diff --git a/pkg/datagatherer/k8s/dynamic_test.go b/pkg/datagatherer/k8s/dynamic_test.go
index 8002a15..fcf0071 100644
--- a/pkg/datagatherer/k8s/dynamic_test.go
+++ b/pkg/datagatherer/k8s/dynamic_test.go
@@ -1103,6 +1103,9 @@ func TestRemoveUnstructuredKeys(t *testing.T) {
                expectObj: map[string]interface{}{
                        "metadata": map[string]interface{}{
                                "name": "foo",
+                               "annotations": map[string]interface{}{
+                                       "key2": "value2",
+                               },
                        },
                },
        }))
@@ -1121,7 +1124,8 @@ func TestRemoveUnstructuredKeys(t *testing.T) {
                },
                expectObj: map[string]interface{}{
                        "metadata": map[string]interface{}{
-                               "name": "foo",
+                               "name":        "foo",
+                               "annotations": map[string]interface{}{},
                        },
                },
        }))
@@ -1140,7 +1144,8 @@ func TestRemoveUnstructuredKeys(t *testing.T) {
                },
                expectObj: map[string]interface{}{
                        "metadata": map[string]interface{}{
-                               "name": "foo",
+                               "name":        "foo",
+                               "annotations": map[string]interface{}{},
                        },
                },
        }))
@@ -1187,7 +1192,12 @@ func TestRemoveUnstructuredKeys(t *testing.T) {
                                "annotations": nil,
                        },
                },
-               expectObj: map[string]interface{}{"metadata": map[string]interface{}{"name": "foo"}},
+               expectObj: map[string]interface{}{
+                       "metadata": map[string]interface{}{
+                               "name":        "foo",
+                               "annotations": nil,
+                       },
+               },
        }))

        t.Run("works when leaf field is unexpectedly not nil and not a known map", run_TestRemoveUnstructuredKeys(tc_RemoveUnstructuredKeys{
@@ -1208,6 +1218,7 @@ func TestRemoveUnstructuredKeys(t *testing.T) {
        t.Run("works when the intermediate field is nil", run_TestRemoveUnstructuredKeys(tc_RemoveUnstructuredKeys{
                givenPath: []string{"metadata", "annotations"},
                givenObj:  map[string]interface{}{"metadata": nil},
+               expectObj: map[string]interface{}{"metadata": nil},
        }))

        t.Run("works when the intermediate field is unexpectedly not nil and not a map", run_TestRemoveUnstructuredKeys(tc_RemoveUnstructuredKeys{
@@ -1228,6 +1239,7 @@ func run_TestRemoveUnstructuredKeys(tc tc_RemoveUnstructuredKeys) func(*testing.
        return func(t *testing.T) {
                t.Helper()
                RemoveUnstructuredKeys(toRegexps(tc.givenExclude), &unstructured.Unstructured{Object: tc.givenObj}, tc.givenPath...)
+               assert.Equal(t, tc.expectObj, tc.givenObj)
        }
 }
```

wallrj (commented):
I noticed some problems with the new tests, and suggested a diff.
I didn't push the changes, you'd better sanity check it first.

pkg/datagatherer/k8s/dynamic_test.go
maelvls: Thanks for noticing this... I rushed this part of the testing, it clearly shows. Thanks for proposing a fix!

I've adjusted the tests.

pkg/datagatherer/k8s/dynamic_test.go
maelvls: Thanks for noticing this; I had forgotten to write the assertion, which led to this 🤦  I fixed this and renamed `key1` to `toexclude` to make it clear which fields are meant to disappear in each test.

wallrj (approved):
Thanks @maelvls 

Thanks for making the new unit tests so easily readable.
You've acknowledged that this only catches the labels in the object metadata,
not in the PodTemplates of Deployment, StatefulSet etc..
But we've agreed that you'll address that in another PR.
The user-facing configuration makes sense and is very well documented.
But we've also acknowledged that it is going to be quite difficult for the user to know whether their supplied exclusion regular expressions are actually working.
They don't have an easy way to see the data that is being collected by the agent before it is posted to the Venafi API.

I imagine that users may ask for a label / annotation (or even a general field)  whitelist feature in future, so that they can declare exactly which fields and values are shared with Venafi....but hopefully this new feature will help us to validate that. 

/approve
/lgtm

wallrj (approved):

PR #617, opened 2024-11-12
[VC-35738] Define well known log verbosity values and log less by default

wallrj:
The aims of this PR are:
 * Make it clearer to the user what the `--verbosity` values mean in practice.
 * Make it clear what the default `--verbosity` value is.
 * Make it clearer to developers which `V()` value to use when they add new log messages.
 * Make the `V()` values more consistent.
 * Reduce the number of Info log messages that the agent prints by default.

### DIFF
--- a/pkg/logs/logs.go
+++ b/pkg/logs/logs.go
@@ -49,6 +49,14 @@ var (
 	features = featuregate.NewFeatureGate()
 )
 
+const (
+	// Standard log verbosity levels.
+	// Use these instead of integers in venafi-kubernetes-agent code.
+	Info  = 0
+	Debug = 1
+	Trace = 2
+)
+
 func init() {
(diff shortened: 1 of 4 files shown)


pkg/logs/logs.go
```diff
@@ -94,6 +102,7 @@ func AddFlags(fs *pflag.FlagSet) {
 		if f.Name == "v" {
 			f.Name = "log-level"
 			f.Shorthand = "v"
+			f.Usage = fmt.Sprintf("%s. 0=Info, 1=Debug, 2=Trace. Use 3-10 for even greater detail. (default: 0)", f.Usage)
```
maelvls, line 105: Thanks for documenting this in `--help`, this is much appreciated!

PR #612, opened 2024-11-07
[VC-35738] Replace logs.Log with logr.Logger in the remaining code

wallrj:
The aim of this PR is to remove the remaining references to the `logs.Log` variable and replace them with context derived `logr.Logger` calls.

I haven't given much thought to the log levels because I think we're going to be removing most of the messages about missing CRDs and permissions errors in another feature.

```console
$ go test ./pkg/datagatherer/k8s/... -v -run TestNoneCache -count 1
=== RUN   TestNoneCache
    cache.go:53: E1107 17:57:50.964580] Cache update failure err="not a cacheResource type: *k8s.notCachable missing metadata/uid field" operation="add"
    cache.go:66: E1107 17:57:50.964680] Cache update failure err="not a cacheResource type: *k8s.notCachable missing metadata/uid field" operation="update"
    cache.go:80: E1107 17:57:50.964689] Cache update failure err="not a cacheResource type: *k8s.notCachable missing metadata/uid field" operation="delete"
--- PASS: TestNoneCache (0.00s)
PASS
ok      github.com/jetstack/preflight/pkg/datagatherer/k8s      0.025s
```

```bash
 ./hack/e2e/test.sh
```

```json
{
  "ts": 1731002359611.0227,
  "caller": "cache/reflector.go:561",
  "msg": "k8s.io/client-go@v0.31.1/tools/cache/reflector.go:243: failed to list firefly.venafi.com/v1, Resource=issuers: issuers.firefly.venafi.com is forbidden: User \"system:serviceaccount:venafi:venafi-kubernetes-agent\" cannot list resource \"issuers\" in API group \"firefly.venafi.com\" at the cluster scope",
  "v": 0
}
{
  "ts": 1731002359611.0823,
  "caller": "k8s/dynamic.go:278",
  "msg": "datagatherer informer has failed and is backing off",
  "v": 0,
  "groupVersionResource": "firefly.venafi.com/v1, Resource=issuers",
  "reason": "failed to list firefly.venafi.com/v1, Resource=issuers: issuers.firefly.venafi.com is forbidden: User \"system:serviceaccount:venafi:venafi-kubernetes-agent\" cannot list resource \"issuers\" in API group \"firefly.venafi.com\" at the cluster scope"
}
```

```json
{
  "ts": 1731002360493.6667,
  "caller": "k8s/dynamic.go:276",
  "msg": "server missing resource for datagatherer",
  "v": 0,
[...]

PR #619, opened 2024-11-14
VC-35738: 

maelvls:
In #612, it is written:

```go
// TestNoneCache demonstrates that the cache helpers do not crash if passed a
// non-cachable object, but log an error with a reference to the object type.
```

It says that "an error is logged", but the logs aren't asserted. I found ktesting's documentation (https://pkg.go.dev/k8s.io/klog/v2/ktesting) that it is possible to assert the logs. I've tried doing that in this PR.

WDYT?


wallrj (changes requested):
I've stated my case below. Interested to hear your counter arguments.

PR #624, opened 2024-11-19
WIP: [VC-35738] Hide the client-go reflector warning logs

wallrj:
The aim of this PR is to filter out the [warnings logged in client-go/tools/cache/reflector.go](https://github.com/kubernetes/client-go/blob/37045084c2aa82927b0e5ffc752861430fd7e4ab/tools/cache/reflector.go#L569) when a resource group version is not installed in the Kubernetes API server.

Hiding the warnings is achieved by creating a filtering logr.LogSink and assigning that to the default klog logger.

However it does not seem possible to maintain the correct file and line numbers when wrapping the logger this way.
I discussed this with the klog authors and they confirm that there are probably bugs in the way the klog functions unwrap the stack:
 * https://kubernetes.slack.com/archives/C20HH14P7/p1732035524601449

In that thread on Slack, I learned from the klog authors that there is an in progress PR which will remove this distracting warning message:
* https://github.com/kubernetes/kubernetes/pull/126387#pullrequestreview-2448130478

So I will close this PR.

### DIFF
--- a/pkg/datagatherer/k8s/dynamic.go
+++ b/pkg/datagatherer/k8s/dynamic.go
@@ -277,7 +277,7 @@ func (g *DataGathererDynamic) Run(stopCh <-chan struct{}) error {
 
 	// attach WatchErrorHandler, it needs to be set before starting an informer
 	err := g.informer.SetWatchErrorHandler(func(r *k8scache.Reflector, err error) {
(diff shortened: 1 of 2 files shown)


maelvls:
Seems like controller-runtime already bumped client-go to v0.33.0-alpha.1, which contains the fix that removes the log line "the server could not find the requested resource". I'll open a PR to see if it does the trick.

PR #625, opened 2024-11-20
[VC-36032] Log the client-id when VenafiCloudKeypair authentication is used

wallrj:
In https://github.com/jetstack/jetstack-secure/issues/549 @hawksight wrote:
> ...we are missing one important piece [of logged information] when connecting to TLSPK on VCP, the client-id

I've added client-id to the log messages when the configuration resolver selects Venafi Cloud Authentication.

But in the interests of reducing the overwhelming quantity of log messages, I've also changed that log message to only be shown at debug log level `--log-level=1`.

I've also associated this PR with `[VC-36032] AGENT: Enhanced troubleshooting through improved logging`.

## Testing

```console
$ go run ./ agent --install-namespace default --private-key-path /dev/null --client-id foo --agent-config-file examples/cert-manager-agent.yaml  --output-path /dev/null -v1
I1120 15:50:50.273235  185320 run.go:59] "Starting" logger="Run" version="development" commit=""
I1120 15:50:50.280681  185320 config.go:410] "Authentication mode" logger="Run" clientID="foo" privateKeyPath="/dev/null" mode="Venafi Cloud Key Pair Service Account" reason="--client-id and --private-key-path were specified"
I1120 15:50:50.280752  185320 config.go:428] "Using deprecated Endpoint configuration. User Server instead." logger="Run"
E1120 15:50:50.280799  185320 root.go:53] "Exiting due to error" err=<
        While evaluating configuration: 2 errors occurred:
                * the venafi-cloud.upload_path field is required when using the Venafi Cloud Key Pair Service Account mode
                * period must be set using --period or -p, or using the 'period' field in the config file

 > exit-code=1
exit status 1
```

```console
$ go test -v ./pkg/agent/... -run  Test_ValidateAndCombineConfig | grep "Authentication mode"
    config.go:410: I1120 15:54:40.370580] Authentication mode mode="Jetstack Secure OAuth" reason="--credentials-file was specified without --venafi-cloud"
    config.go:410: I1120 15:54:40.415486] Authentication mode venConnName="venafi-components" mode="Venafi Cloud VenafiConnection" reason="--venafi-connection was specified"
```

[VC-36032]: https://venafi.atlassian.net/browse/VC-36032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -21,6 +21,7 @@ import (
 	"github.com/jetstack/preflight/pkg/datagatherer/k8s"
 	"github.com/jetstack/preflight/pkg/datagatherer/local"
 	"github.com/jetstack/preflight/pkg/kubeconfig"
+	"github.com/jetstack/preflight/pkg/logs"
 	"github.com/jetstack/preflight/pkg/version"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/agent/config.go
```diff
@@ -401,7 +406,8 @@ func ValidateAndCombineConfig(log logr.Logger, cfg Config, flags AgentCmdFlags)
 				" - Use --api-token if you want to use the " + string(JetstackSecureAPIToken) + " mode.\n")
 		}
 		res.AuthMode = mode
-		log.Info(reason)
+		keysAndValues = append(keysAndValues, "mode", mode, "reason", reason)
+		log.V(logs.Debug).Info("Authentication mode", keysAndValues...)
```
wallrj, line 410: I changed this to a Debug log, so it will not be shown by default. Why? Because I thought it might be desirable to keep the logs as quiet as possible by default, while allowing these details to be logged in the case of problems.

I'm also not sure if the client-id is sensitive information...is there any risk in logging it?

maelvls: I don't think there is any risk printing out the cliend ID. The client ID is just like a username, the difference being that it isn't a personal information. So I'd say that there is no problem displaying the client ID.

hawksight (approved):
The changes look good, and thank you for taking into account the log volume.
That said, it is only 1 log line and would only write once at startup, so perhaps it could just be an info level log? I'm not too fusses, just a suggestion :)

What is that default `--log-level` set to? Or is it not set on deployment? 

I am just trying to think how a user would enable this, either directly through helm or `venctl`.

wallrj:
> What is that default `--log-level` set to? Or is it not set on deployment?
> I am just trying to think how a user would enable this, either directly through helm or `venctl`.

I added some notes to the Helm chart values and README in #627 

https://github.com/jetstack/jetstack-secure/blob/98afe3b70434d1f64881d7c1b97e5e9a1cbedece/deploy/charts/venafi-kubernetes-agent/values.yaml#L148-L159

Those updated values will eventually find there way into the the docs at:
 * https://docs.venafi.cloud/vaas/k8s-components/c-vka-helmvalues/#extraargs

PR #626, opened 2024-11-20
[VC-36032] Configure controller-runtime to log with klog

wallrj:
I assume that the controller-runtime logs were being discarded.
With this change I hope to find some controller-runtime logs among the e2e test logs.

## Testing

```console
$ make test-e2e-gke
...
{
  "ts": 1732123466929.3813,
  "caller": "metrics/metrics.go:97",
  "msg": "cache not synced yet, skipping metrics venaficonnection_*",
  "source": "controller-runtime",
  "v": 0
}
...
{"ts":1732123472568.4834,"caller":"agent/run.go:409","msg":"Data sent successfully","v":0,"logger":"Run.gatherAndOutputData.postData"}
```

That single message seems to come from venafi-connection-lib, so perhaps it's misleading to label these logs with `source: "controller-runtime"`:
 * https://github.com/jetstack/venafi-connection-lib/pull/146
 * https://github.com/jetstack/venafi-connection-lib/blob/13c2342fe0140ff084d2aabfd29ae3d10721691b/internal/metrics/metrics.go#L92-L102

@maelvls Does that log message indicate a bug in the way we're instantiating the venafi-connection-lib client?

### DIFF
--- a/hack/e2e/values.venafi-kubernetes-agent.yaml
+++ b/hack/e2e/values.venafi-kubernetes-agent.yaml
@@ -10,3 +10,4 @@ authentication:
 
 extraArgs:
 - --logging-format=json
+- --log-level=2
--- a/pkg/client/client_venconn.go
+++ b/pkg/client/client_venconn.go
@@ -12,7 +12,6 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/go-logr/logr"
 	venapi "github.com/jetstack/venafi-connection-lib/api/v1alpha1"
 	"github.com/jetstack/venafi-connection-lib/venafi_client"
 	"github.com/jetstack/venafi-connection-lib/venafi_client/auth"
@@ -22,7 +21,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
-	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/jetstack/preflight/api"
 	"github.com/jetstack/preflight/pkg/version"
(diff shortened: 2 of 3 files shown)


pkg/client/client_venconn.go
```diff
@@ -57,11 +55,6 @@ type VenConnClient struct {
 // `restcfg` is not mutated. `trustedCAs` is only used for connecting to Venafi
 // Cloud and Vault and can be left nil.
 func NewVenConnClient(restcfg *rest.Config, agentMetadata *api.AgentMetadata, installNS, venConnName, venConnNS string, trustedCAs *x509.CertPool, disableCompression bool) (*VenConnClient, error) {
-	// TODO(mael): The rest of the codebase uses the standard "log" package,
-	// venafi-connection-lib uses "go-logr/logr", and client-go uses "klog". We
-	// should standardize on one of them, probably "slog".
-	ctrlruntimelog.SetLogger(logr.Logger{})
```
wallrj, line 63: This was from https://github.com/jetstack/jetstack-secure/pull/552:
 * https://github.com/jetstack/jetstack-secure/pull/552

hack/e2e/values.venafi-kubernetes-agent.yaml
```diff
@@ -10,3 +10,4 @@ authentication:
 
 extraArgs:
 - --logging-format=json
+- --log-level=2
```
wallrj, line 13: I changed this so that could see trace level logs in the e2e test.
As it happens, the single controller-runtime log message from venafi-connection-lib is at level 0, so this change is not strictly necessary.

maelvls:
> cache not synced yet, skipping metrics venaficonnection_*

I remember adding this log line in venafi-connection-lib's [`metrics.go`](https://github.com/jetstack/venafi-connection-lib/blob/13c2342fe0140ff084d2aabfd29ae3d10721691b/internal/metrics/metrics.go#L93-L99) (internal link). I remember writing this:

```go
// We found a niche problem where `/metrics` would hang forever in case of a
// misconfigured RBAC. This was due to `cache.List` hanging until the cache
// is synced. To prevent that, we skip reporting this subset of the metrics.
if !hasSynced(cc.cache) {
	cc.log.Info("cache not synced yet, skipping metrics venaficonnection_*")
	return
}
```

I don't know why `WaitForCache` is taking so long to return. The Helm chart's RBAC is correct, so it shouldn't be this... This needs debugging, but we can do that in the next PR.

maelvls (approved):

PR #628, opened 2024-11-22
VC-37264: Disable compression to give us time to work on a fix

maelvls:
Ref: [VC-37264](https://venafi.atlassian.net/browse/VC-37264 "Disable agent's compression")

## Why?

On Nov 21, 2024, the Production Engineering team found that the test pipeline was failing due to venctl 1.15.3. They rolled back to venctl 1.15.2, which made the pipeline work again.

After investigating, we found that the Agent's compression feature broke the Agent. The Agent sends its upload request using `Content-Encoding: gzip`, but the server doesn't process this header and doesn't decompress the body. It copies the contents of the request's body to a bucket without checking its validity. When the backend asynchronously tries to process the contents of the upload, it fails with the error:

```
json: invalid character \u001f as token
```

## Solution

After some consideration, we decided to remove entirely the compression logic that was added in https://github.com/jetstack/jetstack-secure/pull/594. The `--disable-compression` flag **will no longer do anything**, and a log line will be printed when using the flag:

```
INFO The flag --disable-compression has been deprecated an no longer has any effect.
```

Previously, we wanted to simply switch the default value of the `--disable-compression` flag from `true` to `false`; we found the new logic clunky due to the fact that users can set `--disable-compression=false`, so we voted against it. We had also considered adding the flag `--disable-compression` to the Helm chart but realized that it would impact users that don't use the Venafi Connection mode nor the Key Pair Service Account mode.

## Manual test

For this test, I've used the tenant https://ven-tlspk.venafi.cloud/. To access the API key, use the user `system.admin@tlspk.qa.venafi.io` and the password is visible in the page [Production Accounts](https://venafi.atlassian.net/wiki/spaces/CT/pages/2115404149) (private to Venafi). Then go to the settings and find the API key, and set it in the env var `APIKEY`.

The tenant https://ven-tlspk.venafi.cloud/ doesn't have the right tier to pull images, so I use an API key from the tenant https://glow-in-the-dark.venafi.cloud/, that's why I set the env var `APIKEY_GLOW_IN_THE_DARK`. Ask Atanas to get access to the tenant https://glow-in-the-dark.venafi.cloud/.

```sh
export APIKEY=...
export APIKEY_GLOW_IN_THE_DARK=...
```

Then:

```sh
export \
 OCI_BASE=ttl.sh/maelvls \
 VEN_API_KEY=$APIKEY \
 VEN_API_KEY_PULL=$APIKEY_GLOW_IN_THE_DARK \
[...]

PR #631, opened 2024-11-29
[VC-36032] Set User-Agent header containing the agent version in all HTTP requests

wallrj:
In [VC-36032](https://venafi.atlassian.net/browse/VC-36032) we aim to provide:
> Enhanced troubleshooting through improved logging

By adding user-agent headers to *all* HTTP requests, it will allow the Venafi platform team to know which version of the agent is making requests to the resource and auth APIs, by analysing the request headers among the server logs.

I tried writing tests for this, but the it would have required some significant refactoring of the client code,
for example there's no easy way to intercept the  Auth0 client request made by the Jetstack-Secure client in client_oauth.go.
That's a bit of a lame excuse, but I would like to propose that we refactor the Venafi client code in a followup PR and at the same time add tests which verify that all the flavours of Venafi client do include user agent headers.

[VC-36032]: https://venafi.atlassian.net/browse/VC-36032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/agent/config_test.go
+++ b/pkg/agent/config_test.go
@@ -178,7 +178,7 @@ func Test_ValidateAndCombineConfig(t *testing.T) {
 
 		// The log line printed by pflag is not captured by the log recorder.
 		assert.Equal(t, testutil.Undent(`
-			INFO Using the Jetstack Secure OAuth auth mode since --credentials-file was specified without --venafi-cloud.
+			INFO Authentication mode mode="Jetstack Secure OAuth" reason="--credentials-file was specified without --venafi-cloud"
 			INFO Using period from config period="1h0m0s"
 		`), b.String())
 	})
--- a/pkg/client/client_api_token.go
+++ b/pkg/client/client_api_token.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/jetstack/preflight/api"
+	"github.com/jetstack/preflight/pkg/version"
 	"k8s.io/client-go/transport"
 )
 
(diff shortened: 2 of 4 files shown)


pkg/agent/config_test.go
```diff
@@ -178,7 +178,7 @@ func Test_ValidateAndCombineConfig(t *testing.T) {
 
 		// The log line printed by pflag is not captured by the log recorder.
 		assert.Equal(t, testutil.Undent(`
-			INFO Using the Jetstack Secure OAuth auth mode since --credentials-file was specified without --venafi-cloud.
+			INFO Authentication mode mode="Jetstack Secure OAuth" reason="--credentials-file was specified without --venafi-cloud"
```
wallrj, line 181: Another drive-by fix. This test was introduced in #628 but #625 which altered the logging format (and which was merged afterwards) had not been rebased and did not alter the expected log output in this test.

maelvls (approved):
Thanks for fixing this, I wish we had done it sooner. I've realized the importance of the user agent header when responding to last weeks venctl/agent outage.

I haven't tested the PR, but I'm very confident that nothing will break. Nothing in the PR could break anything, I'd say.

pkg/agent/config_test.go
wallrj: The tests are not being run on the master branch and the green badge on the README file does not reflect the status of the current branch:
 * https://github.com/jetstack/jetstack-secure/actions/workflows/tests.yaml?query=branch%3Amaster
 * https://github.com/jetstack/jetstack-secure/blob/master/README.md

PR #635, opened 2024-12-11
Upgrade Makefile modules

inteon:
Manual Makefile modules upgrade.


inteon:
Self-approving this PR since this would have been auto-merged if it wasn't for the breaking change.

PR #640, opened 2025-03-14
Fix govulncheck job by pulling more data from git

SgtCoDFish:
This is similar to an issue in other repos. I coped the job from [trust-manager](https://github.com/cert-manager/trust-manager/blob/735febe3d3777233c497b99ce1362eef207a0f04/.github/workflows/govulncheck.yaml) to make this commit.


maelvls (approved):
thanks for spending the time fixing this!

maelvls:
I see. We are not complying with two of the policies: disallow-privilege-escalation and restrict-seccomp-strict. Do you need help fixing these?

SgtCoDFish:
> Thanks for adding the table. I see that we are not complying with two of the policies: disallow-privilege-escalation and restrict-seccomp-strict. Do you need help fixing these?

I think they're easy to fix, but I don't have the bandwidth today to ensure that they're safe to change for the agent.

If you know that the agent doesn't require privilege escalation / if we can adhere to the strict seccomp policy, then by all means we should fix that straight away!

PR #644, opened 2025-04-16
Make `venafiConnection` a root-level Helm option

inteon:
Move the `venafiConnection` option out of the `crds` sub-object. This matches the config structure we have in other projects.
Next, in a future PR, we can add the `serviceAccountNamespace` option to the `venafiConnection` object.

This is a breaking change, we might want to still allow the old field to be used, but I think it is ok to change the location since users will get an error when they try to use the old field thanks to the json validation.


inteon:
This has been implemented in the new discovery agent chart instead.

PR #646, opened 2025-04-17
feat: add client for cyberark service discovery

SgtCoDFish:
Currently, the only aim of this is to fetch the URL of the Identity API. A future PR will use this client to fetch that API URL and log in via Identity.

I've written this to be specific to Identity, but to be easily extensible later if we want to discover other services.


pkg/internal/cyberark/servicediscovery/discovery.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/jetstack/preflight/pkg/version"
+)
+
+const (
+	prodDiscoveryEndpoint        = "https://platform-discovery.cyberark.cloud/api/v2/"
+	integrationDiscoveryEndpoint = "https://platform-discovery.integration-cyberark.cloud/api/v2/"
+
+	// identityServiceName is the name of the identity service we're looking for in responses from the Service Discovery API
+	// TODO: confirm that this is definitely the one to use
```
SgtCoDFish, line 20: nb: this is a TODO, but I don't think it blocks merging this. We'll confirm ~next week, and update then.

SgtCoDFish: Had the correct one confirmed, so this is now done. I've updated the PR

pkg/internal/cyberark/servicediscovery/testdata/README.md
```diff
@@ -0,0 +1,3 @@
+# Test data for CyberArk Discovery
+
+All data in this folder is derived from an unauthenticated endpoint accessible from the public internet.
```
wallrj, line 3: I'd like to know how to recreate or update this data.
I tried using curl , but got an empty response.

```
$ curl https://platform-discovery.integration-cyberark.cloud/api/v2/services/subdomain/venafi-test; echo
{}
```

pkg/internal/cyberark/servicediscovery/discovery_test.go
```diff
@@ -0,0 +1,153 @@
...
+		return
+	}
+
+	if !strings.HasPrefix(r.URL.String(), "/services/subdomain/") {
+		// This was observed by making a request to /api/v2/services/asd
+		// Normally, we'd expect 404 Not Found but we match the observed response here
+		w.WriteHeader(http.StatusForbidden)
+		_, _ = w.Write([]byte(`{"message":"Missing Authentication Token"}`))
+		return
+	}
```
wallrj, line 36: I get a different response when I use curl to GET that endpoint:

```
$ curl -v https://platform-discovery.integration-cyberark.cloud/api/v2/services/asd | jq
...
> GET /api/v2/services/asd HTTP/2
...
< HTTP/2 400
...
{
  "error": "input validation error"
}
```

wallrj (approved):
I haven't been involved in the discussions around this new feature so I'm a bit out of the loop and don't know the plan.

But it looks like a nice self-contained start. 

Some of the canned responses in the test server didn't match what I got from the live server, but  that might be because I'm not making the right requests.

pkg/internal/cyberark/servicediscovery/testdata/README.md
SgtCoDFish: That tenant was decommissioned and replaced with a new one I'm using to test the Identity client. There's a different example tenant which can be used for getting this kind of data: `curl -v https://platform-discovery.integration-cyberark.cloud/api/v2/services/subdomain/beyonceknowles`

I didn't want to add the old one or the one above to that README because I don't know if they're intended to be long lived, so my instinct is to avoid referring to either in code right now!

pkg/internal/cyberark/servicediscovery/discovery_test.go
SgtCoDFish: Ah that's my fault for just making up a random identifier; I'll update this in the Identity client which is to follow!

SgtCoDFish:
Thanks for the review Richard, really appreciate it!

PR #655, opened 2025-05-16
feat: Initial work on CyberArk Identity client

SgtCoDFish:
This includes a bunch of testing data obtained experimentally from testing against an integration-environment tenant.  

I've confirmed that I'm able to log in using this code in some scenarios; *more work is needed* on this client before it can be used but this is quite a lot of lines added as-is and it's worth getting an initial review of the logic so far.

Most docs are linked in comments, but the most important are:

- [StartAuthentication](https://api-docs.cyberark.com/docs/identity-api-reference/authentication-and-authorization/operations/create-a-security-start-authentication)
- [AdvanceAuthentication](https://api-docs.cyberark.com/docs/identity-api-reference/authentication-and-authorization/operations/create-a-security-advance-authentication)


pkg/internal/cyberark/identity/identity.go
```diff
@@ -0,0 +1,440 @@
...
+	response.MechanismID = mechanismID
+	response.SessionID = startAuthResponse.Result.SessionID
+	response.TenantID = c.subdomain
+	response.PersistantLogin = true
+
+	return response, nil
+}
+
+func (c *Client) doAdvanceAuthentication(ctx context.Context, username string, requestBody *advanceAuthenticationRequestBody) error {
+	bodyJSON, err := json.Marshal(requestBody)
```
djivko, line 377: nit: Seems like doStartAuthentication and doAdvanceAuthentication methods have quite a lot of common code. Does it make sense to extract common template methods?

SgtCoDFish: My instinct is that it's not worth extracting it right now - there are enough differences between the two that it would be a bit harder to read if I did that, IMO. If we had to add a third method which was similar I'd definitely start to consider though!

pkg/internal/cyberark/identity/cmd/testidentity/main.go
wallrj, line 1: I wanted to try this out, but it's not obvious what URL or  credentials to use.
Suggest adding an explanatory comment with an example of usage.

pkg/internal/cyberark/identity/identity.go
```diff
@@ -0,0 +1,441 @@
...
+// identityResponseBody generically wraps a response from the Identity server; the Result will differ for
+// responses from different endpoint, but the other fields are similar.
+// Not all fields in the JSON returned from the server are replicated here, since we only need a subset.
+type identityResponseBody[T any] struct {
+	// Success is a simple boolean indicator from the server of success.
+	// NB: The JSON key is lowercase, in contrast to other JSON keys in the response.
+	Success bool `json:"success"`
+
+	// Result holds the information we need to parse from successful responses
+	Result T `json:"Result"`
```
wallrj, line 77: This is cool. I'd never seen this use of generics before.

pkg/internal/cyberark/identity/identity.go
```diff
@@ -0,0 +1,441 @@
...
+// LoginUsernamePassword performs a blocking call to fetch an auth token from CyberArk Identity using the given username and password.
+// The password is zeroed after use.
+// Tokens are cached internally and are not directly accessible to code; use Client.AuthenticatedHTTPClient to add credentials
+// to an *http.Client.
+func (c *Client) LoginUsernamePassword(ctx context.Context, username string, password []byte) error {
+	defer func() {
+		for i := range password {
+			password[i] = 0x00
+		}
+	}()
```
wallrj, line 234: It's cute, but isn't it a bit theatrical since we've already copied it to a string and assigned it to a struct, below.

pkg/internal/cyberark/identity/identity.go
```diff
@@ -0,0 +1,441 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// doStartAuthentication performs the initial request to start the login process using a username and password.
+// It returns a partially initialized advanceAuthenticationRequestBody ready to send to the server to complete
+// the login. To avoid copying passwords, this function doesn't set the password in the Answer field, and
+// callers must set that field before making the request to AdvanceAuthentication.
```
wallrj, line 259: I don't understand the "avoid copying passwords" comment. This function doesn't have or need the password so why would it set it in the Answer field?
I would find it more useful to have a package level `doc.go` file or README or CONTRIBUTING doc, explaining how future maintainers should write code to minimise secrets getting copied and lingering in memory.

pkg/internal/cyberark/identity/identity.go
```diff
@@ -0,0 +1,441 @@
...
+	advanceRequestBody, err := c.doStartAuthentication(ctx, username)
+	if err != nil {
+		return err
+	}
+
+	// We can't skip AdvanceAuthentication so we need to add the password to the body
+	// and send it off to complete the login process
+	advanceRequestBody.Answer = string(password)
+
+	err = c.doAdvanceAuthentication(ctx, username, &advanceRequestBody)
```
wallrj, line 248: I think it would make more sense if the password bytes were supplied as an argument to this function.
Then the password bytes could be zeroed immediately after they are copied to the request body.

pkg/internal/cyberark/identity/identity.go
SgtCoDFish: I'll try to improve this, thanks!

pkg/internal/cyberark/identity/cmd/testidentity/main.go
SgtCoDFish: I can provide credentials in another place - because this PR is public I don't want to share any details of our private test env for this. I'm sure it would be fine to do so, but I'm being _super_ careful about what I share and don't share in public.

pkg/internal/cyberark/identity/identity.go
SgtCoDFish: >  I would find it more useful to have a package level doc.go file or README or CONTRIBUTING doc, explaining how future maintainers should write code to minimise secrets getting copied and lingering in memory.

I'm gonna save this idea as something which would be useful for us later, but I'm not going to write this now. I've tried to improve the comments a bit for now!

PR #660, opened 2025-06-10
Linter fixes (part 2)

inteon:
depends on https://github.com/jetstack/jetstack-secure/pull/659

Fixes issues found by the linters & enables the linters to prevent these issues from being re-introduced.


pkg/agent/config.go
```diff
@@ -345,7 +345,7 @@ func InitAgentCmdFlags(c *cobra.Command, cfg *AgentCmdFlags) {
 		false,
 		"Enables the MachineHub mode. The agent will push data to CyberArk MachineHub.",
 	)
-	c.PersistentFlags().MarkHidden("machine-hub")
+	_ = c.PersistentFlags().MarkHidden("machine-hub")
```
wallrj: It might be better to panic if there's an error to avoid the hidden flags getting out of sync with the actual flag names.

pkg/datagatherer/k8s/dynamic.go
```diff
@@ -76,27 +76,27 @@ func (c *ConfigDynamic) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-			errors = append(errors, fmt.Sprintf("invalid field selector %d: %s", i, err))
+			errs = append(errs, fmt.Sprintf("invalid field selector %d: %s", i, err))
 		}
 	}
 
-	if len(errors) > 0 {
-		return fmt.Errorf(strings.Join(errors, ", "))
+	if len(errs) > 0 {
+		return errors.New(strings.Join(errs, ", "))
 	}
```
wallrj, line 100: I'm surprised it didn't suggest [`errors.Join`](https://pkg.go.dev/errors#Join).

wallrj (approved):
Thanks. 

/approve
/lgtm

pkg/datagatherer/k8s/dynamic.go
inteon: I think `errors.Join(...).Error()` will print all error messages separated by newlines. This would be different that the current behavior.

pkg/agent/config.go
inteon: Good point. I'll fix that.

PR #661, opened 2025-06-10
Cleanup: config.go

inteon:
- simplified control flow
- deduplicated code
- removed `createCredentialClient`


pkg/agent/config.go
```diff
@@ -779,8 +776,16 @@ func validateCredsAndCreateClient(log logr.Logger, flagCredentialsPath, flagClie
 			return nil, fmt.Errorf("programmer mistake: --client-id and --private-key-path or --credentials-file must have been provided")
 		}
 
+		// The uploader ID isn't actually used in the backend, let's use an
+		// arbitrary value.
+		uploaderID := "no"
+
+		// We don't do this for the VenafiCloudVenafiConnection mode because
+		// the upload_path field is ignored in that mode.
+		log.Info("Loading upload_path from \"venafi-cloud\" configuration.")
```
maelvls, line 785: Why is this unconditionally logged? It used to be only logged in "key pair" mode.

maelvls (approved):
Seems like innocuous cosmetic changes. Not sure why this is being refactored, but OK with that.

Just one nit about one log line, the rest is fine.

pkg/agent/config.go
inteon: It still is only logged in "key pair" mode: https://github.com/jetstack/jetstack-secure/pull/661/files/84570e79542e103e7997c509f9d198052da5f97c#diff-4c826b1b395b31baef6313ac64f7d6e2283739d884418144b0ff908302af6eebR740

PR #662, opened 2025-06-10
Cleanup client interface

inteon:
- Removed `PostDataReadings` function from the client interface, since `PostDataReadingsWithOptions` can be used instead.
- Removed `Post` function from the client interface, since it was not used (`config.OrganizationID` cannot be `""`)

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -451,7 +451,7 @@ func postData(ctx context.Context, config CombinedConfig, preflightClient client
 		return err
 
 	default:
-		return fmt.Errorf("not implemented")
+		return fmt.Errorf("not implemented for mode %s", config.TLSPKMode)
 	}
 }
 


pkg/agent/run.go
```diff
@@ -437,55 +437,22 @@ func postData(ctx context.Context, config CombinedConfig, preflightClient client
 		log.Info("Data sent successfully")
 
 		return nil
-	}
-
-	if config.OrganizationID == "" {
```
inteon, line 442: This will not be true when `TLSPKMode` is `JetstackSecureOAuth` or `JetstackSecureAPIToken`, since this is checked here: https://github.com/jetstack/jetstack-secure/blob/5aab7339bbd84cdcb623db2f8fcea5969d65363a/pkg/agent/config.go#L598-L600

inteon: And the other two modes (`VenafiCloudKeypair` and `VenafiCloudVenafiConnection` are handled above already).

pkg/client/client_venafi_cloud.go
```diff
@@ -205,43 +205,7 @@ func (c *VenafiCloudClient) PostDataReadingsWithOptions(ctx context.Context, rea
...
-	data, err := json.Marshal(payload)
-	if err != nil {
-		return err
-	}
-
-	if !strings.HasSuffix(c.uploadPath, "/") {
-		c.uploadPath = fmt.Sprintf("%s/", c.uploadPath)
-	}
-	res, err := c.Post(ctx, filepath.Join(c.uploadPath, c.uploaderID), bytes.NewBuffer(data))
+	res, err := c.post(ctx, venafiCloudUploadURL.String(), bytes.NewBuffer(data))
```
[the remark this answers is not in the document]
inteon: AI is wrong, the line is there: https://github.com/jetstack/jetstack-secure/pull/662/files#diff-d2c34084bb344da75f4b7852cf9b462d07dcdb162fd08b808dbf6a17a100bf92L248

pkg/agent/run.go
maelvls: I imagine that the test coverage would show that this code path is never run? Either way, this doesn't seem to be tested anyways. I'm fine with removing this.

wallrj (commented):
Yes, I find the new code easier to understand.

/approve
/lgtm

wallrj (approved):

PR #663, opened 2025-06-11
Use new client-go functions for contextual logging

inteon:
Use `Context`s instead of `stopCh` when working with informers, such that client-go can use contextual logging.

See https://pkg.go.dev/k8s.io/client-go@v0.33.1/tools/cache#SharedInformer for more info.

### DIFF
--- a/pkg/datagatherer/k8s/dynamic.go
+++ b/pkg/datagatherer/k8s/dynamic.go
@@ -181,7 +181,6 @@ func (c *ConfigDynamic) newDataGathererWithClient(ctx context.Context, cl dynami
 	dgCache := cache.New(5*time.Minute, 30*time.Second)
 
 	newDataGatherer := &DataGathererDynamic{
-		ctx:                  ctx,
 		groupVersionResource: c.GroupVersionResource,
 		fieldSelector:        fieldSelector.String(),
 		namespaces:           c.IncludeNamespaces,
@@ -217,7 +216,7 @@ func (c *ConfigDynamic) newDataGathererWithClient(ctx context.Context, cl dynami
 		newDataGatherer.informer = factory.ForResource(c.GroupVersionResource).Informer()
 	}
 
-	registration, err := newDataGatherer.informer.AddEventHandler(k8scache.ResourceEventHandlerFuncs{
+	registration, err := newDataGatherer.informer.AddEventHandlerWithOptions(k8scache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
 			onAdd(log, obj, dgCache)
 		},
@@ -227,6 +226,8 @@ func (c *ConfigDynamic) newDataGathererWithClient(ctx context.Context, cl dynami
 		DeleteFunc: func(obj interface{}) {
 			onDelete(log, obj, dgCache)
 		},
+	}, k8scache.HandlerOptions{
+		Logger: &log,
 	})
 	if err != nil {
 		return nil, err
@@ -243,7 +244,6 @@ func (c *ConfigDynamic) newDataGathererWithClient(ctx context.Context, cl dynami
 // This is to allow us to support arbitrary CRDs and resources that Preflight
 // does not have registered as part of its `runtime.Scheme`.
 type DataGathererDynamic struct {
-	ctx context.Context
 	// groupVersionResource is the name of the API group, version and resource
 	// that should be fetched by this data gatherer.
 	groupVersionResource schema.GroupVersionResource
@@ -269,8 +269,8 @@ type DataGathererDynamic struct {
 // Run starts the dynamic data gatherer's informers for resource collection.
 // Returns error if the data gatherer informer wasn't initialized, Run blocks
 // until the stopCh is closed.
-func (g *DataGathererDynamic) Run(stopCh <-chan struct{}) error {
-	log := klog.FromContext(g.ctx)
+func (g *DataGathererDynamic) Run(ctx context.Context) error {
+	log := klog.FromContext(ctx)
 	if g.informer == nil {
 		return fmt.Errorf("informer was not initialized, impossible to start")
 	}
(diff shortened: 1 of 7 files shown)


pkg/datagatherer/k8s/dynamic.go
```diff
@@ -227,6 +226,8 @@ func (c *ConfigDynamic) newDataGathererWithClient(ctx context.Context, cl dynami
 		DeleteFunc: func(obj interface{}) {
 			onDelete(log, obj, dgCache)
 		},
+	}, k8scache.HandlerOptions{
+		Logger: &log,
 	})
```
wallrj, line 231: Paste an example of these new log messages, so that we can be sure that we're not re-opening "flood of log messages" problem:
 * https://github.com/jetstack/jetstack-secure/pull/639
 * https://github.com/jetstack/jetstack-secure/pull/617

pkg/datagatherer/k8s/dynamic.go
```diff
@@ -288,7 +288,7 @@ func (g *DataGathererDynamic) Run(stopCh <-chan struct{}) error {
 	}
 
 	// start shared informer
-	g.informer.Run(stopCh)
+	g.informer.RunWithContext(ctx)
```
wallrj, line 291: Does this change result in additional log messages? If so, paste an example so that we can be sure we're not logging too much.

pkg/datagatherer/k8s/dynamic.go
```diff
@@ -298,8 +298,9 @@ var ErrCacheSyncTimeout = fmt.Errorf("timed out waiting for Kubernetes cache to
 // WaitForCacheSync waits for the data gatherer's informers cache to sync before
 // collecting the resources. Use errors.Is(err, ErrCacheSyncTimeout) to check if
 // the cache sync failed.
-func (g *DataGathererDynamic) WaitForCacheSync(stopCh <-chan struct{}) error {
-	if !k8scache.WaitForCacheSync(stopCh, g.registration.HasSynced) {
+func (g *DataGathererDynamic) WaitForCacheSync(ctx context.Context) error {
+	// Don't use WaitForNamedCacheSync, since we don't want to log extra messages.
+	if !k8scache.WaitForCacheSync(ctx.Done(), g.registration.HasSynced) {
 		return ErrCacheSyncTimeout
 	}
```
wallrj, line 305: The docs say: https://pkg.go.dev/k8s.io/client-go/tools/cache#WaitForNamedCacheSyncWithContext

> // Contextual logging can be used to identify the caller in those log messages. The log level is zero,
> // the same as in [WaitForNamedCacheSync]. If this is too verbose, then store a logger with an increased
> // threshold in the context:

Why not do that?

pkg/datagatherer/k8s/dynamic.go
```diff
@@ -227,6 +226,8 @@ func (c *ConfigDynamic) newDataGathererWithClient(ctx context.Context, cl dynami
 		DeleteFunc: func(obj interface{}) {
 			onDelete(log, obj, dgCache)
 		},
+	}, k8scache.HandlerOptions{
+		Logger: &log,
```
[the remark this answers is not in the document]
inteon: Incorrect, Logger is of type `*klog.Logger` and log is `klog.Logger`.

pkg/datagatherer/k8s/dynamic.go
inteon: I think this might cause an issue with too many extra log messages: https://github.com/kubernetes/client-go/blob/v0.33.1/tools/cache/shared_informer.go#L374

I added a comment above this line to explain this.

wallrj: I'm confused. The function comment suggests how to avoid "too many extra log messages":

> // If this is too verbose, then store a logger with an increased
> // threshold in the context.

Does that work?


....oh!, i think  you mean that the WaitFor**Named**CacheSync functions additionally log **errors** (by way of [utilruntime.HandleErrorWithContext](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/runtime#HandleErrorWithContext)), and errors are always logged, regardless of the log verbosity setting.

pkg/datagatherer/k8s/dynamic.go
inteon: The log messages should be the same, but now we can pass a name etc. to the logger/ tune it.
eg. when we pass `RunWithContext(klog.NewContext(ctx, klog.LoggerWithName(log, "informer")))` and run with `-v 3`:
```
...
I0611 16:27:47.800354 1325202 reflector.go:403] "Listing and watching" logger="informer" type="networking.istio.io/v1alpha3, Resource=gateways" reflector="pkg/mod/k8s.io/client-go@v0.33.0-alpha.1/tools/cache/reflector.go:285"
I0611 16:27:47.804793 1325202 dynamic.go:281] "Server missing resource for datagatherer" groupVersionResource="networking.istio.io/v1alpha3, Resource=gateways"
I0611 16:27:47.820909 1325202 reflector.go:403] "Listing and watching" logger="informer" type="awspca.cert-manager.io/v1beta1, Resource=awspcaclusterissuers" reflector="pkg/mod/k8s.io/client-go@v0.33.0-alpha.1/tools/cache/reflector.go:285"
I0611 16:27:47.824537 1325202 reflector.go:403] "Listing and watching" logger="informer" type="cert-manager.io/v1, Resource=issuers" reflector="pkg/mod/k8s.io/client-go@v0.33.0-alpha.1/tools/cache/reflector.go:285"
I0611 16:27:47.824725 1325202 reflector.go:403] "Listing and watching" logger="informer" type="cert-manager.io/v1, Resource=clusterissuers" reflector="pkg/mod/k8s.io/client-go@v0.33.0-alpha.1/tools/cache/reflector.go:285"
...
```

pkg/datagatherer/k8s/dynamic.go
inteon: This logger is only used in case of issues (eg. incorrect resync period or a crash in the `OnUpdate/ OnAdd ...` functions. Afaik.

inteon:
This change has no effects on the logs when there are not critical errors (crashes in OnAdd etc). Also, since we were already setting the global logger, json logging was already working for the messages that were logged by the informer.
This PR future proofs us a bit and allows us to add more attributes to the informer logs in the future.

PR #664, opened 2025-06-11
Add CyberArk client and mock server

inteon:
Adds client to upload resources to future CyberArk API.
Additionally, adds a mock server to test the upload.

Closes https://venafi.atlassian.net/browse/VC-41697

### DIFF
--- a/pkg/internal/cyberark/identity/authenticated_http_client.go
+++ b/pkg/internal/cyberark/identity/authenticated_http_client.go
@@ -1,0 +1,19 @@
+package identity
+
+import (
+	"fmt"
+	"net/http"
+)
+
+func (c *Client) AuthenticateRequest(req *http.Request) error {
+	c.tokenCachedMutex.Lock()
+	defer c.tokenCachedMutex.Unlock()
+
+	if len(c.tokenCached) == 0 {
+		return fmt.Errorf("no token cached")
+	}
+
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", string(c.tokenCached)))
+
+	return nil
+}
--- a/pkg/internal/cyberark/identity/identity.go
+++ b/pkg/internal/cyberark/identity/identity.go
@@ -174,15 +174,15 @@ type advanceAuthenticationResponseResult struct {
 }
 
 // Client is an client for interacting with the CyberArk Identity API and performing a login using a username and password.
-// For context on the behaviour of this client, see the Pytho SDK: https://github.com/cyberark/ark-sdk-python/blob/3be12c3f2d3a2d0407025028943e584b6edc5996/ark_sdk_python/auth/identity/ark_identity.py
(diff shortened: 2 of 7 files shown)


pkg/internal/cyberark/identity/identity.go
```diff
@@ -174,15 +174,15 @@ type advanceAuthenticationResponseResult struct {
...
-// For context on the behaviour of this client, see the Pytho SDK: https://github.com/cyberark/ark-sdk-python/blob/3be12c3f2d3a2d0407025028943e584b6edc5996/ark_sdk_python/auth/identity/ark_identity.py
+// For context on the behaviour of this client, see the Python SDK: https://github.com/cyberark/ark-sdk-python/blob/3be12c3f2d3a2d0407025028943e584b6edc5996/ark_sdk_python/auth/identity/ark_identity.py
 type Client struct {
 	client *http.Client
 
 	endpoint  string
 	subdomain string
 
-	tokenCache      map[string]token
-	tokenCacheMutex sync.Mutex
```
SgtCoDFish, line 185: suggestion: This change means that if the auth credentials change we'll continue to send the token for the old username, right?

It seems reasonable for us to handle a change of credentials while the agent is running; this would require that the username is passed in when the client is doing an upload, but that seems doable, right?

Was there a different motivation for this change?

pkg/internal/cyberark/identity/authenticated_http_client.go
```diff
@@ -0,0 +1,19 @@
...
+func (c *Client) AuthenticateRequest(req *http.Request) error {
+	c.tokenCachedMutex.Lock()
+	defer c.tokenCachedMutex.Unlock()
+
+	if len(c.tokenCached) == 0 {
+		return fmt.Errorf("no token cached")
+	}
+
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", string(c.tokenCached)))
+
```
SgtCoDFish, line 17: suggestion (non-blocking): I think this would be fine in the existing identity.go file, but that's a matter of taste and it's not important.

SgtCoDFish (changes requested):

pkg/internal/cyberark/identity/identity.go
inteon: It was not clear to me in what situation the auth credentials would change.
As you said, I was trying to not have to pass the username.

I don't understand what the flow would look like, the auth credentials would change but would that not result in the username passed to this function also changing?

SgtCoDFish: > It was not clear to me in what situation the auth credentials would change.

If they're provided as a secret which we read at runtime. It doesn't really matter I guess.

SgtCoDFish (approved):
/lgtm
/approve

it's progress, thanks!

PR #669, opened 2025-06-18
[VC-42222] Proxy settings always ignored

hawksight:
- **fix: Only provide one env key in deployment.yaml**
- **chore: Add two tests for env key and fix failing test for helm chart**

This fixes an issue that meant the proxy settings were always ignored. The `spec.template.spec.containers[0].env` key was specified twice in the template:

1) https://github.com/jetstack/jetstack-secure/blob/v1.5.0/deploy/charts/venafi-kubernetes-agent/templates/deployment.yaml#L36
2) https://github.com/jetstack/jetstack-secure/blob/v1.5.0/deploy/charts/venafi-kubernetes-agent/templates/deployment.yaml#L92

Because the second invocation was never gated, it always took effect. This meant that even when specified, the proxy settings would be overridden.

Here is an example of what happened when set:

```sh
> helm template venafi-kubernetes-agent oci://registry.venafi.cloud/charts/venafi-kubernetes-agent --version v1.5.0 \
   -n venafi --set http_proxy="http://testing.me" \
   --set no_proxy="i_am_not_a_proxy.url" \
   --set https_proxy="https://testing.ssl" \
   | yq '. | select(.kind == "Deployment") | .spec.template.spec.containers[0].env'

Pulled: registry.venafi.cloud/charts/venafi-kubernetes-agent:v1.5.0
Digest: sha256:3a062cbe791dad88a5b858ca621bbccde956bfbf3e80ecd1a95646db1d26db2f
- name: POD_NAMESPACE
  valueFrom:
    fieldRef:
      fieldPath: metadata.namespace
- name: POD_NAME
  valueFrom:
    fieldRef:
      fieldPath: metadata.name
- name: POD_UID
  valueFrom:
    fieldRef:
      fieldPath: metadata.uid
- name: POD_NODE
  valueFrom:
    fieldRef:
      fieldPath: spec.nodeName
```

What happens now:

```sh
> helm template venafi-kubernetes-agent deploy/charts/venafi-kubernetes-agent --version v1.5.0 -n venafi \
   --set http_proxy="http://testing.me" \
   --set no_proxy="i_am_not_a_proxy_url" \
   --set https_proxy="https://testing.ssl" \
   | yq '. | select(.kind == "Deployment") | .spec.template.spec.containers[0].env'

- name: POD_NAMESPACE
  valueFrom:
    fieldRef:
      fieldPath: metadata.namespace
- name: POD_NAME
  valueFrom:
    fieldRef:
      fieldPath: metadata.name
- name: POD_UID
  valueFrom:
    fieldRef:
      fieldPath: metadata.uid
- name: POD_NODE
  valueFrom:
    fieldRef:
      fieldPath: spec.nodeName
- name: HTTP_PROXY
  value: http://testing.me
- name: HTTPS_PROXY
  value: https://testing.ssl
- name: NO_PROXY
  value: i_am_not_a_proxy_url
```

### DIFF
--- a/deploy/charts/venafi-kubernetes-agent/templates/deployment.yaml
+++ b/deploy/charts/venafi-kubernetes-agent/templates/deployment.yaml
@@ -32,8 +32,24 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          {{- if or .Values.http_proxy .Values.https_proxy .Values.no_proxy }}
           env:
+          - name: POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          - name: POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+          - name: POD_UID
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.uid
+          - name: POD_NODE
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+          {{- if or .Values.http_proxy .Values.https_proxy .Values.no_proxy }}
           {{- with .Values.http_proxy }}
           - name: HTTP_PROXY
             value: {{ . }}
@@ -89,23 +105,6 @@ spec:
             {{- with .Values.volumeMounts }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
-          env:
-          - name: POD_NAMESPACE
-            valueFrom:
-              fieldRef:
-                fieldPath: metadata.namespace
-          - name: POD_NAME
-            valueFrom:
-              fieldRef:
-                fieldPath: metadata.name
-          - name: POD_UID
-            valueFrom:
-              fieldRef:
-                fieldPath: metadata.uid
-          - name: POD_NODE
-            valueFrom:
-              fieldRef:
-                fieldPath: spec.nodeName
           {{- if .Values.metrics.enabled }}
           ports:
             - containerPort: 8081
--- a/deploy/charts/venafi-kubernetes-agent/templates/deployment.yaml
+++ b/deploy/charts/venafi-kubernetes-agent/templates/deployment.yaml
@@ -49,7 +49,6 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
-          {{- if or .Values.http_proxy .Values.https_proxy .Values.no_proxy }}
           {{- with .Values.http_proxy }}
           - name: HTTP_PROXY
             value: {{ . }}
@@ -61,7 +60,6 @@ spec:
           {{- with .Values.no_proxy }}
           - name: NO_PROXY
             value: {{ . }}
-          {{- end }}
           {{- end }}
           {{- if not (empty .Values.command) }}
           command:
--- a/deploy/charts/venafi-kubernetes-agent/tests/deployment_test.yaml
+++ b/deploy/charts/venafi-kubernetes-agent/tests/deployment_test.yaml
@@ -103,12 +103,13 @@ tests:
               - key: cabundle
                 path: ca-certificates.crt
 
-  # Check proxy settings are additive not overriding
+  # Check proxy settings are additive not overriding and set to correct values.
+  # Values from our documentation: https://docs.venafi.cloud/vaas/k8s-components/c-vcp-network-requirements/#modifying-network-settings-for-kubernetes
   - it: All environment variables present when all proxy settings are supplied
     set:
-      http_proxy: "http://testing.proxy"
-      no_proxy: "127.0.0.1,localhost"
-      https_proxy: "https://testing.proxy"
+      http_proxy: "http://<proxy server>:<port>"
+      https_proxy: "https://<proxy server>:<port>"
+      no_proxy: "127.0.0.1,localhost,kubernetes.default.svc,kubernetes.default.svc.cluster.local"
     template: deployment.yaml
     asserts:
       - isKind:
@@ -116,7 +117,17 @@ tests:
       - lengthEqual :
           path: spec.template.spec.containers[0].env
           count: 7
+      - equal:
+          path: spec.template.spec.containers[0].env[?(@.name == "NO_PROXY")].value
+          value: "127.0.0.1,localhost,kubernetes.default.svc,kubernetes.default.svc.cluster.local"
+      - equal:
+          path: spec.template.spec.containers[0].env[?(@.name == "HTTPS_PROXY")].value
+          value: "https://<proxy server>:<port>"
+      - equal:
+          path: spec.template.spec.containers[0].env[?(@.name == "HTTP_PROXY")].value
+          value: "http://<proxy server>:<port>"
 
+  # Check no proxy settings are set when no proxy settings are provided
   - it: Only default environment variables are set when no proxy settings are provided
     template: deployment.yaml
     asserts:
--- a/deploy/charts/venafi-kubernetes-agent/tests/deployment_test.yaml
+++ b/deploy/charts/venafi-kubernetes-agent/tests/deployment_test.yaml
@@ -75,11 +75,11 @@ tests:
           path: spec.template.spec.containers[0].command
           content: notpreflight
 
-  # Check the volumes and volumeMounts works correctly
+  # Check the volumes and volumeMounts works correctly - FAILING
   - it: Volumes and VolumeMounts added correctly
     set:
-      config.organisation: test_org
-      config.cluster: test_cluster
+      # config.organisation: test_org -> Should be removed?
+      # config.cluster: test_cluster -> Should be config.clusterName?
     values:
       - ./values/custom-volumes.yaml
     asserts:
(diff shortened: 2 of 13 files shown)


deploy/charts/venafi-kubernetes-agent/templates/deployment.yaml
```diff
@@ -89,23 +105,6 @@ spec:
...
-              fieldRef:
-                fieldPath: metadata.name
-          - name: POD_UID
-            valueFrom:
-              fieldRef:
-                fieldPath: metadata.uid
-          - name: POD_NODE
-            valueFrom:
-              fieldRef:
-                fieldPath: spec.nodeName
```
wallrj, line 108: Apologies. I introduced this bug, in https://github.com/jetstack/jetstack-secure/pull/589/commits/5679dc52aeab5d1ac54462fe3fd800407bf3e394

hawksight:
If testing manually, then you have to escape the `,` values in the helm command:

```sh
helm template venafi-kubernetes-agent deploy/charts/venafi-kubernetes-agent \
  --version v1.5.0 -n venafi \
  --set http_proxy='http://<proxy server>:<port>' \
  --set no_proxy='127.0.0.1\,kubernetes.default.svc\,kubernetes.default.svc.cluster.local' \
  --set https_proxy='https://<proxy server>:<port>'
```

Don't believe this affects YAML values files as this is handled for you there.

wallrj:
I was curious to know why this bug wasn't picked up by the kubeconform check, and the answer is that the duplicate env stanza only appears if we supply the the proxy values. 
I verified this on the master branch by temporarily enabling the `no_proxy` value by default and re-running kubeconform, and this time it identified the problem:

```diff
--- a/deploy/charts/venafi-kubernetes-agent/values.yaml
+++ b/deploy/charts/venafi-kubernetes-agent/values.yaml
@@ -106,7 +106,7 @@ podSecurityContext: {}
 # Configures the NO_PROXY environment variable where a HTTP proxy is required,
 # but certain domains should be excluded.
 # +docs:property
-# no_proxy: 127.0.0.1,localhost
+no_proxy: 127.0.0.1,localhost

 # Add Container specific SecurityContext settings to the container. Takes
 # precedence over `podSecurityContext` when set. See
```

```
$ make verify-helm-kubeconform
rm -rf _bin/scratch/helm/venafi-kubernetes-agent-v1.5.0-47-g076f7341070089-dirty.tgz.tmp
mkdir -p _bin/scratch/helm/
cp -a deploy/charts/venafi-kubernetes-agent _bin/scratch/helm/venafi-kubernetes-agent-v1.5.0-47-g076f7341070089-dirty.tgz.tmp
/home/richard/projects/venafi/jetstack-secure/_bin/tools/yq '( .image.repository = "quay.io/jetstack/venafi-agent" ) | ( .image.tag = "v1.5.0-47-g076f7341070089-dirty" )' _bin/scratch/helm/venafi-kubernetes-agent-v1.5.0-47-g076f7341070089-dirty.tgz.tmp/values.yaml --inplace
/home/richard/projects/venafi/jetstack-secure/_bin/tools/yq '.annotations."artifacthub.io/prerelease" = "false"' \
        --inplace _bin/scratch/helm/venafi-kubernetes-agent-v1.5.0-47-g076f7341070089-dirty.tgz.tmp/Chart.yaml
mkdir -p _bin/scratch/helm/
/home/richard/projects/venafi/jetstack-secure/_bin/tools/helm package _bin/scratch/helm/venafi-kubernetes-agent-v1.5.0-47-g076f7341070089-dirty.tgz.tmp \
        --app-version v1.5.0-47-g076f7341070089-dirty \
        --version v1.5.0-47-g076f7341070089-dirty \
        --destination _bin/scratch/helm/
Successfully packaged chart and saved it to: _bin/scratch/helm/venafi-kubernetes-agent-v1.5.0-47-g076f7341070089-dirty.tgz
stdin - Deployment venafi-kubernetes-agent-release-name failed validation: error unmarshalling resource: error converting YAML to JSON: yaml: unmarshal errors:
  line 68: key "env" already set in map
make: *** [make/_shared/helm///helm.mk:186: verify-helm-kubeconform] Error 1
```

An idea for the future is that we could (somehow) extract all the example values (commented out) and the kube-conform check with those values enabled.

wallrj (approved):
Thanks for fixing this @hawksight 

I gave up trying to get the agent running in a cluster with a proxy available, but the changes to the Helm templates make perfect sense to me and the tests demonstrate that the correct proxy environment variables are added to the right part of the Deployment.

deploy/charts/venafi-kubernetes-agent/tests/deployment_test.yaml
```diff
@@ -75,11 +75,11 @@ tests:
...
           content: notpreflight
 
-  # Check the volumes and volumeMounts works correctly
+  # Check the volumes and volumeMounts works correctly - FAILING
   - it: Volumes and VolumeMounts added correctly
     set:
-      config.organisation: test_org
-      config.cluster: test_cluster
+      # config.organisation: test_org -> Should be removed?
+      # config.cluster: test_cluster -> Should be config.clusterName?
```
wallrj: @maelvls Will know more about this. It was introduced in https://github.com/jetstack/jetstack-secure/pull/543 but the tests were originally written for the jetstack-secure chart, so perhaps  there's been some mixup when running the tests for each chart, which went unnoticed because the helm unittests weren't being run in CI.

https://github.com/jetstack/jetstack-secure/blob/076f734107008945b8da90c8808824f4cac13105/deploy/charts/jetstack-agent/values.yaml#L90-L99

Better resolve this before merging this PR.

hawksight: Looking at the test, its validating mounting custom volumes / volumeMount for a CA cert, which still works. I think the config option as you say is for the previous revision of the agent / chart. 

Given it works without those set, let's remove the values and leave the test as it. Sorry I probably indeed added this based on some customer options.

deploy/charts/venafi-kubernetes-agent/tests/deployment_test.yaml
```diff
@@ -75,11 +75,8 @@ tests:
           path: spec.template.spec.containers[0].command
           content: notpreflight
 
-  # Check the volumes and volumeMounts works correctly
+  # Check the volumes and volumeMounts works correctly - FAILING
```
wallrj:
```suggestion
  # Check the volumes and volumeMounts works correctly
```

deploy/charts/venafi-kubernetes-agent/templates/deployment.yaml
```diff
@@ -32,8 +32,24 @@ spec:
...
+                fieldPath: metadata.name
+          - name: POD_UID
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.uid
+          - name: POD_NODE
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+          {{- if or .Values.http_proxy .Values.https_proxy .Values.no_proxy }}
```
wallrj: I think this if statement can be removed, because its original purpose was to omit the `env` stanza if no proxy env vars were needed. Now the `env` stanza is always included. I'll remove it.

make/02_mod.mk
```diff
@@ -52,3 +52,15 @@ shared_generate_targets += generate-crds-venconn
 ## @category Testing
 test-e2e-gke:
 	./hack/e2e/test.sh
+
+.PHONY: test-helm
+## Run `helm unittest`.
+## @category Testing
+test-helm: | $(NEEDS_HELM-UNITTEST)
+	$(HELM-UNITTEST) ./deploy/charts/venafi-kubernetes-agent/
```
wallrj, line 60: Not including the jetstack-secure chart here because it is soon to be removed: #672

wallrj:
I've done some further testing. 
I ran ` make test-e2e-gke` and observed the test pass. .

I then modified the values for the test to include the proxy settings.

```diff 
diff --git a/hack/e2e/values.venafi-kubernetes-agent.yaml b/hack/e2e/values.venafi-kubernetes-agent.yaml
index 630d76e..99d1757 100644
--- a/hack/e2e/values.venafi-kubernetes-agent.yaml
+++ b/hack/e2e/values.venafi-kubernetes-agent.yaml
@@ -11,3 +11,7 @@ authentication:
 extraArgs:
 - --logging-format=json
 - --log-level=6
+
+https_proxy: "https://proxy:8080"
+
+no_proxy: 127.0.0.1,localhost,kubernetes.default.svc,kubernetes.default.svc.cluster.local
```

The test fail, because I hadn't deployed a proxy.
But I could see the new expected proxy environment variables in addition to the standard env vars

```
$ kubectl describe deploy -n venafi venafi-kubernetes-agent
...
    Environment:
      POD_NAMESPACE:   (v1:metadata.namespace)
      POD_NAME:        (v1:metadata.name)
      POD_UID:         (v1:metadata.uid)
      POD_NODE:        (v1:spec.nodeName)
      HTTPS_PROXY:    https://proxy:8080
      NO_PROXY:       127.0.0.1,localhost,kubernetes.default.svc,kubernetes.default.svc.cluster.local

```

wallrj (approved):
/approve
/lgtm

wallrj:
@maelvls, @hawksight  I stumbled on this comment in the Kubernetes code today:
> // http.ProxyFromEnvironment doesn't respect CIDRs and that makes it impossible to exclude things like pod and service IPs from proxy settings
> // ProxierWithNoProxyCIDR allows CIDR rules in NO_PROXY
> --https://github.com/kubernetes/apimachinery/blob/d6651abdfec8bb9c001b60653bbda1168bb74a0b/pkg/util/net/http.go#L110-L114

https://github.com/kubernetes/kubernetes/blob/dcefe0ef41113324a4e4e43606f868a0858c55a6/staging/src/k8s.io/apimachinery/pkg/util/net/http.go#L396-L432

Perhaps we should use that wrapper too, to support CIDR in NO_PROXY?

PR #672, opened 2025-06-18
Remove jetstack-agent chart + docs

SgtCoDFish:
I think most of the jetstack-agent stuff has been removed and is no longer relevant to what we do day-to-day. This is a first step towards cleaning up the code base so we can refactor further.

There are several references in README to Jetstack secure, to the old chart, and to stuff that's no longer relevant, so I've cleaned it up a lot.

I also moved the release guide to RELEASE.md to match what we do in other projects, and cleaned up the release info to be a bit more readable and focus on details that are important to us.

I deleted the "Tiers" section, because I think that belongs in enterprise docs and not here. It'll just get out of date here.

## New Tag

To preserve the old Jetstack Secure functionality in the event we require it, I've already created + pushed a tag called
"before-jetstack-secure-removal".

If we need to do something Jetstack Secure related (such as an agent change or a chart tweak) we can create a branch from that tag and maintain that separately.


inteon (approved):
Thanks for this cleanup.

PR #682, opened 2025-08-07
fix: upgrade klone, vendor govulncheck workflow

SgtCoDFish:
I'd mistakenly thought that govulncheck respected GOPRIVATE from local testing, but I'd actually set up an environment where govulncheck had access to the private dependency I was trying to avoid and as such it had passed locally and started to fail in CI.

This PR:

1. Incorporates the contents of https://github.com/cert-manager/makefile-modules/pull/317
2. Disables upstream govulncheck
3. Makes some changes to the newly vendored govulncheck job so it runs


SgtCoDFish:
> Looks like you ran govulncheck workflow on this branch and it passed

I did, but the klone.yaml is still pointing at my fork - I'll repoint at the upstream makefile-modules!

PR #692, opened 2025-08-20
[VC-43403] Add Local File output path mode

wallrj-cyberark:
Changes:
- Refactored TLSPKMode to a more general OutputMode.
- Added a new Local File output mode.

Next I'd like to add a MachineHub mode:
 * #696 

Later I'll try and figure how we can allow multiple output modes.

## Testing

```
$ go run . agent    --agent-config-file examples/one-shot-secret.yaml    --one-shot    --output-path output.json
I0822 12:06:42.354685  284845 run.go:58] "Starting" logger="Run" version="development" commit=""
I0822 12:06:42.355219  284845 run.go:116] "Healthz endpoints enabled" logger="Run.APIServer" addr=":8081" path="/healthz"
I0822 12:06:42.355273  284845 run.go:120] "Readyz endpoints enabled" logger="Run.APIServer" addr=":8081" path="/readyz"
I0822 12:06:42.355300  284845 run.go:269] "Pod event recorder disabled" logger="Run" reason="The agent does not appear to be running in a Kubernetes cluster." detail="When running in a Kubernetes cluster the following environment variables must be set: POD_NAME, POD_NODE, POD_UID, POD_NAMESPACE"
I0822 12:06:42.558945  284845 client_file.go:35] "Data saved to local file" logger="Run.gatherAndOutputData.postData" outputPath="output.json"
I0822 12:06:42.558992  284845 run.go:417] "Data sent successfully" logger="Run.gatherAndOutputData.postData"

$ jq  -r '.[] | .data.items[] | .resource.metadata | "\(.namespace)/\(.name)"' < output.json
gmp-system/collection
gmp-system/rules
gmp-system/webhook-tls
team-1/app-0
venafi/cert-manager-approver-policy-tls
venafi/cert-manager-webhook-ca
default/venafi-kubernetes-agent-e2e.a2074131-f85a-48c0-88be-fb1a1d871a3b
gmp-system/alertmanager
```

PR #653, opened 2025-05-14
VC-41203: Allow users to select the Machine Hub mode

maelvls:
Ref: [VC-41203](https://venafi.atlassian.net/browse/VC-41203)

## Context

Up to now, Venafi Kubernetes Agent (this project) was already able to push to two different backends (Jetstack Secure, and Venafi Cloud). We want Machine Hub to act as another backend ("control plane").

Unlike Jetstack Secure and Venafi Cloud that were exclusive to each other, it was decided that the Venafi Cloud backend could be used alongside the Machine Hub backend. The intention is (probably) to make the transition painless.

## What am I adding?

This PR originates from Ashley's prototype (#652); I've added a new flag (`--machine-hub`) to turn on pushing to the Machine Hub.

Just to clarify: the Machine Hub integration **won't be working yet after this PR**. It's just an empty shell for now. This PR is just about adding a way to turn on the feature, but doesn't implement the feature.

## Things to consider

**Why `--machine-hub` when everything must be configured in the config file?**

You might have noticed that I could have just used the presence of the `machineHub` field in the configuration to turn on the feature, e.g.:

```yaml
machineHub:
  subdomain: foo
  credentialsSecretName: secret-1
period: 1h
```

I guess we can remove `--machine-hub` and just rely on the presence of the `machineHub` field; I've added the flag to make it "feel" similar to how you would turn on the Venafi Cloud Key Pair mode: `--venafi-cloud` alongside `--client-id`. But unlike `--venafi-cloud` that allowed us to disabiguate between the Jetstack Secure and Venafi Cloud modes, `--machine-hub` doesn't actually serve any real purpose.

## Testing

I've added two new unit tests to check that the configuration logic works.

Here is the manual testing I've done:

### Case 1: Machine Hub only

```console
$ go run . agent --machine-hub --install-namespace=default -c /dev/stdin <<EOF
period: 30s
machineHub:
  subdomain: "machinehub"
  credentialsSecretName: creds
data-gatherers: []
EOF
I0514 11:11:52.880764   10543 run.go:59] "Starting" logger="Run" version="development" commit=""
I0514 11:11:52.881631   10543 config.go:432] "Will push to CyberArk MachineHub using a username and password loaded from a Kubernetes Secret" logger="Run" credentialsSecretName="creds"
I0514 11:11:52.881639   10543 config.go:620] "Using period from config" logger="Run" period="30s"
[...]

PR #694, opened 2025-08-21
[VC-43403]  Revert some of the MachineHub changes from #653

wallrj-cyberark:
I want to revert some of the code which @maelvls  introduced in https://github.com/jetstack/jetstack-secure/pull/653 because I'm finding it really difficult to work with.

Specifically, 
 * I want to load the MachineHub / DisCo config and credentials from a file mounted into the Pod. I don't want to have to watch and load them from a Secret, so the SecretName configuration and subdomain configuration is redundant. 
 * I want to refactor `TLSPKMode` to be called `OutputMode`, or `Outputs` and add a `LocalFile` mode. Then we can just build a list of output functions, depending on what command line flags have been supplied and call them all in a batch, periodically, or one-shot.
 * Alternatively, we could rename `client.Client` to `output.Publisher` and have a function to compose an outputter from one or more of: LocalFilePublisher, TLSPKPublisher, MachineHubPublisher. 

Here's what I want to do next, if this PR merges:
 * #692 
 * #696 

I explained the problems to @inteon in the standup meeting this morning, so I'll ask him to review this.

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -1,7 +1,6 @@
 package agent
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"net/url"
@@ -62,9 +61,6 @@ type Config struct {
 	ExcludeAnnotationKeysRegex []string `yaml:"exclude-annotation-keys-regex"`
 	// Skips label keys that match the given set of regular expressions.
 	ExcludeLabelKeysRegex []string `yaml:"exclude-label-keys-regex"`
-
-	// MachineHub holds config specific to MachineHub mode.
-	MachineHub MachineHubConfig `yaml:"machineHub"`
 }
 
 type Endpoint struct {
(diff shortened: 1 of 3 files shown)


inteon (approved):

pkg/agent/config.go
```diff
@@ -92,33 +88,6 @@ type VenafiCloudConfig struct {
...
-type MachineHubConfig struct {
-	// Subdomain is the subdomain indicating where data should be pushed. Used
-	// for querying the Service Discovery Service to discover the Identity API
-	// URL.
-	Subdomain string `yaml:"subdomain"`
-
-	// CredentialsSecretName is the name of a Kubernetes Secret in the same
-	// namespace as the agent, which will be watched for a username and password
-	// to send to CyberArk Identity for authentication.
-	CredentialsSecretName string `yaml:"credentialsSecretName"`
```
maelvls, line 105: I agree, not sure why we went with loading the credentials from a Kubernetes secret.

PR #695, opened 2025-08-21
[VC-43403] Disable events when not in cluster and only require installNamespace when using venafi-connection mode

wallrj-cyberark:
It should be possible to run the agent outside a cluster, reading data from an input file and publishing to Jetstack or TLSPK, without having to supply any Kubernetes related configuration.

`--install-namespace` or `POD_NAMESPACE` is only required when using venafi-connection mode, because that client needs to know where to find the agent service account which it uses to impersonate the venafi-connection service account (I think).

So if the necessary POD_ environment variables are missing, the event recorder is now disabled. 
And specifying an install namespace (by `--install-namespace` or by `POD_NAMESPACE`) are now optional, unless venafi-connection mode is used.


cmd/agent_test.go
wallrj-cyberark, line 1:
```bash
$ go test ./cmd/...  -v -count 1
=== RUN   TestAgentRunOneShot
    agent_test.go:32: Running child process
    agent_test.go:50: STDOUT
        I0821 16:50:33.617894  135769 run.go:58] "Starting" logger="Run" version="development" commit=""
        I0821 16:50:33.618071  135769 config.go:491] "Configured to push to Venafi" logger="Run" mode="Jetstack Secure API Token" reason="--api-token was specified"
        I0821 16:50:33.618111  135769 run.go:116] "Healthz endpoints enabled" logger="Run.APIServer" addr=":8081" path="/healthz"
        I0821 16:50:33.618123  135769 run.go:120] "Readyz endpoints enabled" logger="Run.APIServer" addr=":8081" path="/readyz"
        I0821 16:50:33.618140  135769 run.go:269] "Pod event recorder disabled" logger="Run" reason="The agent does not appear to be running in a Kubernetes cluster." detail="When running in a Kubernetes cluster the following environment variables must be set: POD_NAME, POD_NODE, POD_UID, POD_NAMESPACE"
        I0821 16:50:33.618149  135769 run.go:308] "Reading data from local file" logger="Run.gatherAndOutputData" inputPath="testdata/agent/one-shot/success/input.json"
        I0821 16:50:33.618224  135769 run.go:334] "Data saved to local file" logger="Run.gatherAndOutputData" outputPath="/dev/null"
        I0821 16:50:33.618246  135769 run.go:476] "Starting" logger="Run.APIServer.ListenAndServe" addr=":8081"
        I0821 16:50:33.618250  135769 run.go:490] "Shutting down" logger="Run.APIServer.ListenAndServe" addr=":8081"
        I0821 16:50:33.618295  135769 run.go:505] "Shutdown complete" logger="Run.APIServer.ListenAndServe" addr=":8081"

    agent_test.go:51: STDERR

--- PASS: TestAgentRunOneShot (0.03s)
PASS
ok      github.com/jetstack/preflight/cmd       0.064s
````

File: cmd/agent_test.go [hunk 1]
```diff
@@ -15,21 +15,12 @@ import (
 // after the first data gathering iteration.
 func TestAgentRunOneShot(t *testing.T) {
 	if _, found := os.LookupEnv("GO_CHILD"); found {
-		// Silence the warning about missing pod name for event generation
-		// TODO(wallrj): This should not be required when an `--input-file` has been supplied.
-		t.Setenv("POD_NAME", "venafi-kubernetes-e2e")
-		// Silence the error about missing kubeconfig.
-		// TODO(wallrj): This should not be required when an `--input-file` has been supplied.
-		t.Setenv("KUBECONFIG", "testdata/agent/one-shot/success/kubeconfig.yaml")
```
wallrj-cyberark, line 23: These deleted  TODO comments were slightly inaccurate. What I meant was...if the agent is not running in a cluster, there's no Pod to attach events to, so the POD_NAME is not necessary.
And if the data is being read from a file, there are no K8S datagatherers, so KUBECONFIG should not be necessary. Nor is there a venafi-connection-lib to initialize, which is another reason KUBECONFIG is unnecessary.

cmd/agent_test.go
```diff
@@ -15,21 +15,12 @@ import (
...
-
 		os.Args = []string{
 			"preflight",
 			"agent",
 			"--one-shot",
 			// TODO(wallrj): This should not be required when an `--input-file` has been supplied.
 			"--api-token=should-not-be-required",
-			// TODO(wallrj): This should not be required when an `--input-file` has been supplied.
-			"--install-namespace=default",
```
wallrj-cyberark, line 32: And this deleted TODO comment should have said...`--install-namespace` is only necessary for initializing the venafi-connection-lib client, so it should not be necessary in this test.

pkg/agent/config_test.go
```diff
@@ -274,7 +273,6 @@ func Test_ValidateAndCombineConfig(t *testing.T) {
 			DataGatherers: []DataGatherer{
 				{Name: "d1", Kind: "dummy", Config: &dummyConfig{AlwaysFail: false}},
 			},
-			InputPath:      "/home",
```
wallrj-cyberark: I'll revert these unnecessary changes.

maelvls (approved):
You are correct, `--install-namespace` was originally only meant to let venafi-connection-lib where to find the service account used to impersonate.

Seems like I hijacked `--install-namespace` in https://github.com/jetstack/jetstack-secure/pull/589, but that was somewhat stupid; I should have used `POD_NAMESPACE` instead of using `--install-namespace`. My mistake!!!

Thanks for the example log output, it looks good.

PR #713, opened 2025-09-11
[VC-43403] Add filter for excluding TLS secrets without client certificates

wallrj-cyberark:
This PR introduces a configurable filter for the k8s-dynamic data gatherer that excludes some Kubernetes TLS secrets; 
those which do not contain a client certificate. 
This is in support of the CyberArk Discovery and Context service which wants various Secret types, but in the case of X.509 certificates, they only need certificates which allow Client Authentication. So in the interests of least privilege we try to only send data that is strictly needed.
 
- Introduce ExcludeTLSSecretsWithoutClientCert filter for k8s-dynamic gatherer
- Update config, docs, and examples to support and demonstrate filter usage
- Implement filter logic to exclude TLS secrets lacking client certificates
- Extend tests to cover filtering of TLS secrets based on certificate EKU
- Add logging for filter decisions and error cases

Part of: https://venafi.atlassian.net/browse/VC-43403

## Testing

Run the agent and watch it log trace messages as you create TLS secrets
```
$ go run ./cmd/ark agent --log-level 2 -c examples/machinehub.yaml --output-path data.json -p 1h
```

```
$ step certificate create foo foo.crt foo.key --profile self-signed --subtle --no-password --insecure
Your certificate has been saved in foo.crt.
Your private key has been saved in foo.key.

$ kubectl create secret  tls foo --key=foo.key --cert foo.crt
secret/foo created
```

```
I0911 22:34:03.464756  162389 cache.go:115] "TLS Secret contains a client certificate" namespace="default" name="foo"
I0911 22:35:08.707640  162389 cache.go:115] "TLS Secret contains a client certificate" namespace="cert-manager" name="foo"
```

Run the agent in --machine-hub mode and use mitmproxy to examine the snapshot that it uploads:

```
mitmproxy
```

```
$ export HTTPS_PROXY=localhost:8080
$ go run ./cmd/ark agent --log-level 2 -c examples/machinehub.yaml --machine-hub --one-shot
```

Export the request (via mitmproxy)

```
$ tail -1 request.txt | jq '.secrets'
```

```json

  {
    "apiVersion": "v1",
    "data": {
      "tls.crt": "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUJlakNDQVNDZ0F3SUJBZ0lRSWl2YnBPSzZLL1JOWWs5WVRjR0t5akFLQmdncWhrak9QUVFEQWpBT01Rd3cKQ2dZRFZRUURFd05tYjI4d0hoY05NalV3T1RFeE1qRXpNek01V2hjTk1qVXdPVEV5TWpFek16TTVXakFPTVF3dwpDZ1lEVlFRREV3Tm1iMjh3V1RBVEJnY3Foa2pPUFFJQkJnZ3Foa2pPUFFNQkJ3TkNBQVNFQnpyN3ozSWphMnJnCis0ODJrbUtCZzhWQUY5MElOQzBqYVM5bFZKbElCS2Zja2NhcDEvbHNKRTdCTm9UMDZVS3hTWng2NUwzWTZGTzYKdkRFeGE4cU5vMkF3WGpBT0JnTlZIUThCQWY4RUJBTUNCNEF3SFFZRFZSMGxCQll3RkFZSUt3WUJCUVVIQXdFRwpDQ3NHQVFVRkJ3TUNNQjBHQTFVZERnUVdCQlRSamxjRFdpaTBNNTQvLzRSVmQ1SERFa3RkTFRBT0JnTlZIUkVFCkJ6QUZnZ05tYjI4d0NnWUlLb1pJemowRUF3SURTQUF3UlFJaEFOazFOckNabk1FMHJlSlROU3NHT3N4aGk2ZnUKZ2paOFd1OG02QURJek9YMkFpQkU0V2gvUFhDdngyR3liZ2ltWVNYWXkxY1RNaFhjbXZ2a2JCbDhObnBEeGc9PQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg=="
    },
    "kind": "Secret",
    "metadata": {
      "creationTimestamp": "2025-09-11T21:34:03Z",
      "name": "foo",
      "namespace": "default",
      "resourceVersion": "4618",
      "uid": "11bd6a5e-b79b-4117-ac38-c16cb7f911d9"
    },
    "type": "kubernetes.io/tls"
  }
```

### DIFF
--- a/docs/datagatherers/k8s-dynamic.md
+++ b/docs/datagatherers/k8s-dynamic.md
@@ -106,3 +106,26 @@ when listing Secrets.
     - type!=bootstrap.kubernetes.io/token
     - type!=helm.sh/release.v1
 ```
+
+## Filters
+
+You can use filters to drop certain resources based on custom logic.
+For example, you can drop TLS secrets that do not contain any client certificates using the `ExcludeTLSSecretsWithoutClientCert` filter, as shown below:
+
+```yaml
+- kind: "k8s-dynamic"
+  name: "k8s/secrets"
+  config:
+    resource-type:
+      version: v1
+      resource: secrets
+    filters:
+    - ExcludeTLSSecretsWithoutClientCert
+```
+
+The available filters are:
+* `ExcludeTLSSecretsWithoutClientCert`: Drops TLS secrets that do not contain any client certificates.
+
+If you find that the filters are not having the desired effect, you can enable
+debug logging by setting the log-level to `debug`. This will log info about why
+certain resources are not being gathered.
--- a/pkg/datagatherer/k8s/cache.go
+++ b/pkg/datagatherer/k8s/cache.go
@@ -1,11 +1,16 @@
 package k8s
 
 import (
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/pem"
 	"fmt"
 	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/pmylund/go-cache"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/jetstack/preflight/api"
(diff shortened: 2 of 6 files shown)


pkg/datagatherer/k8s/dynamic.go
```diff
@@ -218,7 +245,7 @@ func (c *ConfigDynamic) newDataGathererWithClient(ctx context.Context, cl dynami
 
 	registration, err := newDataGatherer.informer.AddEventHandlerWithOptions(k8scache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			onAdd(log, obj, dgCache)
+			onAdd(log, obj, dgCache, c.Filters...)
 		},
 		UpdateFunc: func(oldObj, newObj interface{}) {
 			onUpdate(log, oldObj, newObj, dgCache)
```
maelvls, line 251: Shouldn't the filter also apply the the update func? I'm not sure, since I don't know how these AddFunc, UpdateFunc, DeleteFunc relate to the underlying client-go informer.
If it's not required, would be nice to have a comment explaining why the filtering only applies to AddFunc.

Also a test for that?

pkg/datagatherer/k8s/cache.go
```diff
@@ -39,9 +44,141 @@ func logCacheUpdateFailure(log logr.Logger, obj interface{}, operation string) {
...
+
+// excludeTLSSecretsWithoutClientCert filters out all TLS secrets that do not
+// contain a client certificate in the `tls.crt` key.
+// Secrets are obtained by a DynamicClient, so they have type
+// *unstructured.Unstructured.
+func excludeTLSSecretsWithoutClientCert(log logr.Logger, obj interface{}) bool {
+	// Fast path: type assertion and kind/type checks
+	unstructuredObj, ok := obj.(*unstructured.Unstructured)
+	if !ok {
+		log.V(4).Info("Object is not a Unstructured", "type", fmt.Sprintf("%T", obj))
```
maelvls, line 60: Shouldn't this be at level 1 or 2? The Helm chart says:

```yaml
# The log levels are: 0=Info, 1=Debug, 2=Trace.
```

pkg/datagatherer/k8s/dynamic.go
```diff
@@ -218,7 +245,7 @@ func (c *ConfigDynamic) newDataGathererWithClient(ctx context.Context, cl dynami
 
 	registration, err := newDataGatherer.informer.AddEventHandlerWithOptions(k8scache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			onAdd(log, obj, dgCache)
+			onAdd(log, obj, dgCache, c.Filters...)
```
maelvls, line 248: Filters are applied on add events but not on updates. If a TLS Secret’s EKU changes (e.g., client → server-only), the cache will retain an entry that should now be excluded.

Proposal:
- Pass filters to `onUpdate` as well. If any filter excludes `newObj`, remove it from the cache; if it was previously excluded but now passes, add/replace it.
- Add tests for both transitions (client→non-client and non-client→client) to ensure cache consistency.

pkg/datagatherer/k8s/dynamic.go
```diff
@@ -43,6 +43,11 @@ type ConfigDynamic struct {
 	IncludeNamespaces []string `yaml:"include-namespaces"`
 	// FieldSelectors is a list of field selectors to use when listing this resource
 	FieldSelectors []string `yaml:"field-selectors"`
+	// Filters is a list of filter functions to apply to the resources before adding them to the cache.
+	// Each filter function should return true if the resource should be excluded, false otherwise.
+	// Available filter functions:
+	// - ExcludeTLSSecretsWithoutClientCert: ignores all TLS secrets that do not contain client certificates
+	Filters []cacheFilterFunction `yaml:"filters"`
```
maelvls, line 50: Config shape nit: `Filters` on `ConfigDynamic` stores function pointers. That works after unmarshalling, but it makes the config awkward to marshal/log and couples config with runtime state. Consider storing `[]string` on the config (the names), and mapping them to functions during construction (e.g., in `NewDataGatherer`).

docs/datagatherers/k8s-dynamic.md
```diff
@@ -106,3 +106,26 @@ when listing Secrets.
...
+      resource: secrets
+    filters:
+    - ExcludeTLSSecretsWithoutClientCert
+```
+
+The available filters are:
+* `ExcludeTLSSecretsWithoutClientCert`: Drops TLS secrets that do not contain any client certificates.
+
+If you find that the filters are not having the desired effect, you can enable
+debug logging by setting the log-level to `debug`. This will log info about why
```
maelvls, line 130: Minor doc fix: the project uses numeric `--log-level` (0=Info, 1=Debug, 2=Trace; 6–9 HTTP trace). Suggest rewording to something like: "set `--log-level=1` (Debug)" so users don’t try `--log-level=debug`.

pkg/datagatherer/k8s/cache.go
```diff
@@ -39,9 +44,141 @@ func logCacheUpdateFailure(log logr.Logger, obj interface{}, operation string) {
...
+		tlsCrtBytes = v
+	default:
+		log.V(4).Info("tls.crt is not a string or byte slice", "type", fmt.Sprintf("%T", v))
+		return true
+	}
+
+	// Parse PEM certificate chain
+	certs, err := parsePEMCertificateChain(tlsCrtBytes)
+	if err != nil || len(certs) == 0 {
+		log.V(4).Info("Failed to parse tls.crt as PEM encoded X.509 certificate chain", "error", err.Error())
```
maelvls, line 107: possible nil dereference here: when `parsePEMCertificateChain` returns `(nil, nil)` (or `len(certs) == 0`), `err` is nil and `err.Error()` will panic

PR #765, opened 2026-01-27
Lift service discovery client out of CyberArkClient

SgtCoDFish:
First commit; Service discovery will also be required for getting encryption keys for envelope encryption. This change makes the service discovery client available earlier, so we can fetch keys before we sanitise / encrypt secret data.

Second commit: Bit more visibility into potentially flaky tests.

### DIFF
--- a/internal/cyberark/client.go
+++ b/internal/cyberark/client.go
@@ -49,24 +49,23 @@ func LoadClientConfigFromEnvironment() (ClientConfig, error) {
 // NewDatauploadClient initializes and returns a new CyberArk Data Upload client.
 // It performs service discovery to find the necessary API endpoints and authenticates
 // using the provided client configuration.
-func NewDatauploadClient(ctx context.Context, httpClient *http.Client, cfg ClientConfig) (*dataupload.CyberArkClient, error) {
-	discoveryClient := servicediscovery.New(httpClient)
-	serviceMap, err := discoveryClient.DiscoverServices(ctx, cfg.Subdomain)
-	if err != nil {
-		return nil, err
-	}
+func NewDatauploadClient(ctx context.Context, httpClient *http.Client, serviceMap *servicediscovery.Services, cfg ClientConfig) (*dataupload.CyberArkClient, error) {
 	identityAPI := serviceMap.Identity.API
 	if identityAPI == "" {
 		return nil, errors.New("service discovery returned an empty identity API")
 	}
-	identityClient := identity.New(httpClient, identityAPI, cfg.Subdomain)
-	err = identityClient.LoginUsernamePassword(ctx, cfg.Username, []byte(cfg.Secret))
-	if err != nil {
-		return nil, err
-	}
+
 	discoveryAPI := serviceMap.DiscoveryContext.API
 	if discoveryAPI == "" {
 		return nil, errors.New("service discovery returned an empty discovery API")
 	}
+
+	identityClient := identity.New(httpClient, identityAPI, cfg.Subdomain)
+
+	err := identityClient.LoginUsernamePassword(ctx, cfg.Username, []byte(cfg.Secret))
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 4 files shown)


internal/cyberark/client.go
```diff
@@ -49,24 +49,23 @@ func LoadClientConfigFromEnvironment() (ClientConfig, error) {
...
 	if discoveryAPI == "" {
 		return nil, errors.New("service discovery returned an empty discovery API")
 	}
+
+	identityClient := identity.New(httpClient, identityAPI, cfg.Subdomain)
+
+	err := identityClient.LoginUsernamePassword(ctx, cfg.Username, []byte(cfg.Secret))
+	if err != nil {
+		return nil, err
+	}
```
SgtCoDFish, line 68: note: Moving the login down isn't strictly required in this PR, but I did it so that we check both the identity and discovery API availability first, before we try the login. This means if we didn't get a disco API endpoint from service discovery, we won't do a pointless login attempt.

internal/cyberark/client.go
```diff
@@ -49,24 +49,23 @@ func LoadClientConfigFromEnvironment() (ClientConfig, error) {
...
-func NewDatauploadClient(ctx context.Context, httpClient *http.Client, cfg ClientConfig) (*dataupload.CyberArkClient, error) {
-	discoveryClient := servicediscovery.New(httpClient)
-	serviceMap, err := discoveryClient.DiscoverServices(ctx, cfg.Subdomain)
-	if err != nil {
-		return nil, err
-	}
+func NewDatauploadClient(ctx context.Context, httpClient *http.Client, serviceMap *servicediscovery.Services, cfg ClientConfig) (*dataupload.CyberArkClient, error) {
 	identityAPI := serviceMap.Identity.API
 	if identityAPI == "" {
 		return nil, errors.New("service discovery returned an empty identity API")
```
wallrj-cyberark, line 55: Optional: This error still talks about "service discovery returned..."
```suggestion
		return nil, errors.New("missing or empty serviceMap.Identity.API")
```

internal/cyberark/client.go
```diff
@@ -49,24 +49,23 @@ func LoadClientConfigFromEnvironment() (ClientConfig, error) {
...
 	}
-	identityClient := identity.New(httpClient, identityAPI, cfg.Subdomain)
-	err = identityClient.LoginUsernamePassword(ctx, cfg.Username, []byte(cfg.Secret))
-	if err != nil {
-		return nil, err
-	}
+
 	discoveryAPI := serviceMap.DiscoveryContext.API
 	if discoveryAPI == "" {
 		return nil, errors.New("service discovery returned an empty discovery API")
```
wallrj-cyberark, line 60: Optional: This error still talks about "service discovery returned..."
```suggestion
		return nil, errors.New("missing or empty serviceMap.DiscoveryContext.API")
```

internal/cyberark/client.go
```diff
@@ -49,24 +49,23 @@ func LoadClientConfigFromEnvironment() (ClientConfig, error) {
 // NewDatauploadClient initializes and returns a new CyberArk Data Upload client.
 // It performs service discovery to find the necessary API endpoints and authenticates
 // using the provided client configuration.
-func NewDatauploadClient(ctx context.Context, httpClient *http.Client, cfg ClientConfig) (*dataupload.CyberArkClient, error) {
-	discoveryClient := servicediscovery.New(httpClient)
-	serviceMap, err := discoveryClient.DiscoverServices(ctx, cfg.Subdomain)
-	if err != nil {
-		return nil, err
-	}
+func NewDatauploadClient(ctx context.Context, httpClient *http.Client, serviceMap *servicediscovery.Services, cfg ClientConfig) (*dataupload.CyberArkClient, error) {
```
wallrj-cyberark, line 52: Maybe check for `nil` serviceMap here, and return an error, to avoid a panic, or make the argument a value rather than a pointer.

wallrj-cyberark (approved):
A couple of suggestions, for this PR or a followup. 
Looks fine to merge as-is if you want to avoid another round of CI and review.

PR #787, opened 2026-03-24
disco: re-enable secret sending by default

SgtCoDFish:
We got confirmation from the DisCo team that the encryption we're doing works as expected. As such, the next release of the DisCo agent should enable the secret sending by default, as we'd planned.


inteon (approved):
Based on what @SgtCoDFish said: this was approved by management & security.
I only reviewed the code, not the solution as a whole.

PR #790, opened 2026-04-16
NGTS: Various further updates and fixes

SgtCoDFish:
This PR mostly fixes up comments, errors and test setup. Each individual commit should be self-descriptive.

This also adds the discovery-agent to the release process.

IMPORTANT: This fixes the ngts-e2e test and adds the possibility to run it in CI but restricts it via a label - it seems the dev API endpoints are blocked from github actions runners currently and the test just hangs and dies for that reason.

I _have_ run the e2e test locally and had success - but I don't think we can enable the test on every PR until we have a stable test tenant which is publicly accessible.


deploy/charts/discovery-agent/values.yaml
```diff
@@ -1,6 +1,63 @@
-# Default values for discovery-agent.
-# This is a YAML-formatted file.
-# Declare variables to be passed into your templates.
+# Configuration for the Discovery Agent
+config:
```
SgtCoDFish, line 2: note: this config section is just moved from below; it makes sense to have it at the top of the file since these are the most important options. The only other changes are to comments and to the type of tsgID

deploy/charts/discovery-agent/templates/configmap.yaml
```diff
@@ -20,9 +20,9 @@ data:
     {{- end }}
     data-gatherers:
     - kind: k8s-discovery
-      name: ngts/discovery
+      name: k8s/discovery
```
SgtCoDFish, line 23: note: this fixes something claude introduced. The disco-agent uses the ark prefix because it collects fundamentally different data to the venafi-kubernetes-agent.

Claude copied the idea of a different prefix, but it doesn't really gain us anything so I reverted here. This file should be very similar to the venafi-kubernetes-agent configmap

deploy/charts/discovery-agent/templates/_helpers.tpl
```diff
@@ -62,10 +62,8 @@ Create the name of the service account to use
 {{- end }}
 
 {{/*
-Util function for generating the image URL based on the provided options.
-IMPORTANT: This function is standardized across all charts in the cert-manager GH organization.
-Any changes to this function should also be made in cert-manager, trust-manager, approver-policy, ...
-See https://github.com/cert-manager/cert-manager/issues/6329 for a list of linked PRs.
+Util function for generating an image reference based on the provided options.
+This function is derived from similar functions used in the cert-manager GitHub organization
```
SgtCoDFish, line 66: note: replacing a copy-paste comment with one for this repo specifically!

deploy/charts/discovery-agent/templates/_helpers.tpl
```diff
@@ -85,12 +83,6 @@ usage through tuple/variable indirection.
 {{- $repository := "" -}}
 {{- if $image.repository -}}
 	{{- $repository = $image.repository -}}
-	{{- /*
-		Backwards compatibility: if image.registry is set, additionally prefix the repository with this registry.
-	*/ -}}
-	{{- if $image.registry -}}
-		{{- $repository = printf "%s/%s" $image.registry $repository -}}
-	{{- end -}}
```
SgtCoDFish, line 93: note: I see no reason to have this backwards compat shim in a brand new chart, so I removed it for simplicity.

.github/workflows/tests.yaml
```diff
@@ -109,6 +109,41 @@ jobs:
           ARK_USERNAME: ${{ secrets.ARK_USERNAME }}
           ARK_SECRET: ${{ secrets.ARK_SECRET }}
 
+  ngts-test-e2e:
```
SgtCoDFish, line 112: note: it makes sense to me to add this here gated behind a label.

I added the required variables to this repo already (NGTS_CLIENT_ID, etc) - but this test doesn't pass. I think it's because the relevant dev environment has IP address restrictions which github actions runners are falling foul of.

I don't think we should block this PR on the test passing - I've confirmed locally that the test can pass, I just don't think for now it will pass.

Ideally in the future when we have a stable prod environment for testing, we'll be able to point this test there.

.github/workflows/tests.yaml
```diff
@@ -149,7 +184,7 @@ jobs:
         id: timestamp # Give the step an ID to reference its output
         run: |
           # Generate a timestamp in the format YYMMDD-HHMMSS.
-          # Extracting from PR name would require sanitization due to GKE cluster naming constraints        
+          # Extracting from PR name would require sanitization due to GKE cluster naming constraints
```
SgtCoDFish, line 187: note: simply removing trailing whitespace

hack/ngts/test-e2e.sh
SgtCoDFish, line 1: note: this script was never run before it originally merged, so the state it was in before this PR was filled with guesses

SgtCoDFish (commented):
self-review comments!

hack/ngts/test-e2e.sh
```diff
@@ -56,41 +56,75 @@ fi
...
-     --set-json "podLabels={\"discovery-agent.ngts/test-id\": \"${RANDOM}\"}"
+     --values "${tmp_dir}/values.yaml"
 
 kubectl rollout status deployments/discovery-agent --namespace "${NAMESPACE}"
 
-# Wait 60s for log message indicating success.
+# Wait for log message indicating success.
 # Parse logs as JSON using jq to ensure logs are all JSON formatted.
-timeout 60 jq -n \
+timeout 120 jq -n \
```
George-Yanev, line 127: **issue:** On line 133 and line 141 `--namespace ngts` ngts is hard-coded but the namespace value is used as variable in other parts of the script. This should be fix as if the namespace is not ngts and the script will fail in that scenario

.github/workflows/tests.yaml
```diff
@@ -109,6 +109,41 @@ jobs:
...
+      - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c # v6.1.0
+        with:
+          go-version: ${{ steps.go-version.outputs.result }}
+
+      - uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        with:
+          path: _bin/downloaded
+          key: downloaded-${{ runner.os }}-${{ hashFiles('klone.yaml') }}-test-unit
+
+      - run: make -j ngts-test-e2e
```
George-Yanev, line 140: **issue:** You are missing in the Makefile - `02_mod.mk` ${NEEDS_YQ}  which will be provided to the `test-e2e.sh` script which starts using it from this PR

It is only for MacOS environment so the CI will not catch it

hack/ngts/custom_ca.yaml
George-Yanev, line 1: nice

George-Yanev (approved):
👍

PR #792, opened 2026-04-21
Run make upgrade-klone and make generate

SgtCoDFish:
Mostly, this is to bump go, but it also brings in a new upstream helm unit-test target which requires a few updates and some cleanup of this repository's old targets.

Important: I've also gated the ark-test-e2e behind a label because it's so consistently flaky (see the history of runs of #790 for motivation, and see the comment I added to the ark-test-e2e job)

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -12,3 +12,4 @@ aliases:
     - inteon
     - thatsmrtalbot
     - erikgb
+    - hjoshi123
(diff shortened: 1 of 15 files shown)


OWNERS_ALIASES
SgtCoDFish, line 1: self-review: this file is definitely not needed in this repo (it's a cert-manager thing we pull in).

But, it's not doing any harm and the _actual_ OWNERS file is massively out of date too. I'm leaving cleaning that up as a separate thing for the future and ignoring it here.

inteon (approved):

PR #793, opened 2026-04-22
Force TSG ID to be string, add helm unit-tests

SgtCoDFish:
This follows a similar change by @inteon in the Firefly Helm chart.

Allowing numbers can cause issues with scientific notation because of YAML's weirdness.

This PR also adds a plethora of helm unit tests - mostly, I just wanted to add a test for TSG IDs but I let claude generate a bunch more. I think the tests add some value, and it's better to have them than not.

### DIFF
--- a/deploy/charts/discovery-agent/tests/poddisruptionbudget_test.yaml
+++ b/deploy/charts/discovery-agent/tests/poddisruptionbudget_test.yaml
@@ -1,0 +1,102 @@
+suite: test poddisruptionbudget
+templates:
+  - poddisruptionbudget.yaml
+
+tests:
+  # Test PodDisruptionBudget is not created by default
+  - it: should not create PodDisruptionBudget when disabled
+    set:
+      config.clusterName: test-cluster
+      config.tsgID: "123456"
+      podDisruptionBudget.enabled: false
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  # Test PodDisruptionBudget is created when enabled
+  - it: should create PodDisruptionBudget when enabled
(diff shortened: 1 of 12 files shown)


deploy/charts/discovery-agent/tests/poddisruptionbudget_test.yaml
```diff
@@ -0,0 +1,102 @@
...
+          of: PodDisruptionBudget
+      - equal:
+          path: metadata.name
+          value: RELEASE-NAME-discovery-agent
+
+  # Test default minAvailable when neither minAvailable nor maxUnavailable is set
+  - it: should set default minAvailable when no disruption values are set
+    set:
+      config.clusterName: test-cluster
+      config.tsgID: "123456"
```
inteon, line 33: Hmmmm, for enterprise-issuer and distributed-issuer we were using "tsgid" and "clientId" intead of "tsgID" and "clientID". We might have to switch ....

inteon (approved):

deploy/charts/discovery-agent/tests/poddisruptionbudget_test.yaml
inteon: (unrelated to this PR)

SgtCoDFish: Thanks for pointing that out - definitely worth thinking about. Since the old venafi-kubernetes-agent chart supported `clientId` I've added support for that form in the latest commit I've pushed.

PR #794, opened 2026-04-23
VC-52159: add certOwnership Helm flag to discovery-agent chart

George-Yanev:
Add a certOwnership configuration field that flows from Helm values to the upload URL query parameter (?certOwnership=unassigned). When set, the TLSPK backend stores cert_ownership_mode='unassigned' on the cluster and leaves discovered certificates with NULL sub_tsg_id (claimable by any sub-TSG) rather than stamping them with the cluster's sub-TSG ID.

Changes:
- pkg/client/client.go: CertOwnership field on Options struct
- pkg/client/client_ngts.go: include ?certOwnership= in upload URL when set
- pkg/agent/config.go: CertOwnership in Config and CombinedConfig
- pkg/agent/run.go: pass CertOwnership through to PostDataReadingsWithOptions
- deploy/charts/discovery-agent/values.yaml: certOwnership optional field
- deploy/charts/discovery-agent/templates/configmap.yaml: render cert_ownership
- deploy/charts/discovery-agent/values.schema.json: schema entry for new field
- Tests: client_ngts_test and config_test verify the field flows end-to-end

Local replace directive added for venafi-connection-lib (private GitHub repo requires SAML SSO — points to local clone at development time).

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -55,9 +55,13 @@ type Config struct {
 	ClusterName string `yaml:"cluster_name"`
 	// ClusterDescription is a short description of the Kubernetes cluster where the
 	// agent is running.
-	ClusterDescription string             `yaml:"cluster_description"`
-	DataGatherers      []DataGatherer     `yaml:"data-gatherers"`
-	VenafiCloud        *VenafiCloudConfig `yaml:"venafi-cloud,omitempty"`
+	ClusterDescription string `yaml:"cluster_description"`
+	// ClaimableCerts controls whether discovered certs can be claimed by other tenants.
+	// true = certs are left unassigned, available for any tenant to claim.
+	// false (default) = certs are owned by this cluster's tenant.
+	ClaimableCerts bool               `yaml:"claimable_certs"`
+	DataGatherers  []DataGatherer     `yaml:"data-gatherers"`
+	VenafiCloud    *VenafiCloudConfig `yaml:"venafi-cloud,omitempty"`
 
 	// For testing purposes.
 	InputPath string `yaml:"input-path"`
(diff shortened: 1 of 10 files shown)


pkg/client/client_ngts_test.go
```diff
@@ -339,12 +339,25 @@ func TestNGTSClient_PostDataReadingsWithOptions(t *testing.T) {
...
 	require.NoError(t, err)
 	assert.Equal(t, 1, len(payload.DataReadings))
+
+	// Verify certOwnership=unassigned is included when set
+	t.Run("certOwnership=unassigned included in query when set", func(t *testing.T) {
+		optsUnassigned := Options{
+			ClusterName:   "test-cluster",
+			CertOwnership: "unassigned",
+		}
+		err = client.PostDataReadingsWithOptions(context.Background(), readings, optsUnassigned)
```
SgtCoDFish: nitpick: `t.Context()` over `context.Background()`

SgtCoDFish (changes requested):

pkg/agent/config.go
```diff
@@ -419,6 +423,11 @@ type CombinedConfig struct {
 	// the agent is running.
 	ClusterDescription string
 
+	// CertOwnership controls how discovered certs are scoped in NGTS.
+	// "unassigned" = certs stored with NULL sub_tsg_id, claimable by any sub-TSG.
+	// Empty (default) = TSG_OWNED: certs inherit the cluster's sub-TSG ID.
```
SgtCoDFish: suggestion: are we expecting this to gain more values in the future? seems like it could be simpler as a boolean?

pkg/client/client_ngts_test.go
George-Yanev: Done

pkg/agent/config.go
George-Yanev: Moved to boolean, change the name to reflect more cleanly what it does. fix the comments

SgtCoDFish (approved):
/lgtm
/approve

I'm not 100% sure about the name "claimableCerts" - I think there might be a more descriptive name for that, but I can't think of a better one. But I don't think it's worth blocking over, and if we ship that name it's not the end of the world.

Thanks for this!

PR #797, opened 2026-05-01
fix: per-gatherer excludeAnnotationKeysRegex now excludes resources from upload

FelixPhipps:
Fixes a bug where annotation/label-based exclusion rules had no effect on which resources were uploaded to TLSPK.                                                                                                      
                                                                                                                                                                                                                         
Two issues were present:                                                                                                                                                                                               

1.   The per-gatherer config fields excludeAnnotationKeysRegex and excludeLabelKeysRegex were ignored — ConfigDynamic didn't have those fields, so the YAML was discarded at parse time.                        
2.  Even via the top-level exclude-annotation-keys-regex, matching resources were still uploaded. The existing logic only stripped the matching keys from resource metadata (a privacy feature) — it never dropped the resource itself.                                                                                                                                                                                                       
                                                                  
Fix: Added excludeAnnotationKeysRegex and excludeLabelKeysRegex to ConfigDynamic so per-gatherer config is correctly parsed. Added excludeResources() called in Fetch() before redactList(), which drops any resource whose annotation or label keys match the configured patterns. Both the per-gatherer and top-level config formats now work as customers expect.

Automated testing added to PR.

### DIFF
--- a/pkg/datagatherer/k8sdynamic/dynamic.go
+++ b/pkg/datagatherer/k8sdynamic/dynamic.go
@@ -82,6 +82,10 @@ type ConfigDynamic struct {
 	FieldSelectors []string `yaml:"field-selectors"`
 	// LabelSelectors is a list of label selectors to use when listing this resource
 	LabelSelectors []string `yaml:"label-selectors"`
+	// ExcludeAnnotationKeysRegex is a list of regular expressions to exclude.
+	ExcludeAnnotationKeysRegex []string `yaml:"excludeAnnotationKeysRegex"`
+	// ExcludeLabelKeysRegex is a list of regular expressions to exclude.
+	ExcludeLabelKeysRegex []string `yaml:"excludeLabelKeysRegex"`
 }
 
 // UnmarshalYAML unmarshals the ConfigDynamic resolving GroupVersionResource.
@@ -93,10 +97,12 @@ func (c *ConfigDynamic) UnmarshalYAML(unmarshal func(any) error) error {
 			Version  string `yaml:"version"`
 			Resource string `yaml:"resource"`
 		} `yaml:"resource-type"`
-		ExcludeNamespaces []string `yaml:"exclude-namespaces"`
-		IncludeNamespaces []string `yaml:"include-namespaces"`
-		FieldSelectors    []string `yaml:"field-selectors"`
-		LabelSelectors    []string `yaml:"label-selectors"`
+		ExcludeNamespaces          []string `yaml:"exclude-namespaces"`
+		IncludeNamespaces          []string `yaml:"include-namespaces"`
+		FieldSelectors             []string `yaml:"field-selectors"`
+		LabelSelectors             []string `yaml:"label-selectors"`
+		ExcludeAnnotationKeysRegex []string `yaml:"excludeAnnotationKeysRegex"`
+		ExcludeLabelKeysRegex      []string `yaml:"excludeLabelKeysRegex"`
 	}{}
 	err := unmarshal(&aux)
 	if err != nil {
@@ -111,6 +117,8 @@ func (c *ConfigDynamic) UnmarshalYAML(unmarshal func(any) error) error {
 	c.IncludeNamespaces = aux.IncludeNamespaces
 	c.FieldSelectors = aux.FieldSelectors
 	c.LabelSelectors = aux.LabelSelectors
+	c.ExcludeAnnotationKeysRegex = aux.ExcludeAnnotationKeysRegex
+	c.ExcludeLabelKeysRegex = aux.ExcludeLabelKeysRegex
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/datagatherer/k8sdynamic/dynamic.go
```diff
@@ -434,6 +463,41 @@ func (g *DataGathererDynamic) Fetch(ctx context.Context) (any, int, error) {
...
 }
 
+// excludeResources drops any resource whose annotation or label keys match the
+// configured exclusion patterns. This is distinct from redactList, which strips
+// matching keys from kept resources.
+func (g *DataGathererDynamic) excludeResources(list []*api.GatheredResource) []*api.GatheredResource {
+	if len(g.ExcludeAnnotKeys) == 0 && len(g.ExcludeLabelKeys) == 0 {
+		return list
+	}
+	result := list[:0]
```
SgtCoDFish: suggestion: I think this is a little bit cleverer than we need to be here since we're reusing the memory that's backing `list`.

Just having a dumb slice var `var result []*api.GatheredResource` and appending to that is going to be a little less efficient but a lot less confusing to reason about with less potential for bugs

Alternatively, maybe using [slices.DeleteFunc](https://pkg.go.dev/slices#DeleteFunc) would be clearer than both?

pkg/datagatherer/k8sdynamic/dynamic.go
```diff
@@ -434,6 +463,41 @@ func (g *DataGathererDynamic) Fetch(ctx context.Context) (any, int, error) {
 	}, len(items), nil
 }
 
+// excludeResources drops any resource whose annotation or label keys match the
+// configured exclusion patterns. This is distinct from redactList, which strips
+// matching keys from kept resources.
+func (g *DataGathererDynamic) excludeResources(list []*api.GatheredResource) []*api.GatheredResource {
```
SgtCoDFish, line 482: note: There's risk here; this could be very, very slow and CPU intensive. There are a lot of ways for the amount of work needed by this function to increase:

- More labels
- More annotations
- More resources
- More exclusion rules
- More complex exclusion rules

I don't think we need to do anything about that right now (ultimately, the user could configure a bunch of complicated regexes on a huge cluster and that'll just _be_ slow) but it might be worth capturing in a (doc) comment that there's risk to this.

pkg/datagatherer/k8sdynamic/dynamic.go
```diff
@@ -434,6 +463,41 @@ func (g *DataGathererDynamic) Fetch(ctx context.Context) (any, int, error) {
...
+	return result
+}
+
+func (g *DataGathererDynamic) resourceMatchesExclusionKeys(item *api.GatheredResource) bool {
+	if res, ok := item.Resource.(*unstructured.Unstructured); ok {
+		return anyKeyMatches(res.GetAnnotations(), g.ExcludeAnnotKeys) ||
+			anyKeyMatches(res.GetLabels(), g.ExcludeLabelKeys)
+	}
+	return false
+}
```
SgtCoDFish, line 497: nitpick (non-blocking): I think this function feels like it's the wrong way around; normally in Go I think we'd check for the error (in this case, "Resource isn't castable to (*unstructured.Unstructured)") and return if it fails

It's not wrong and it doesn't really matter in this instance, but I'm writing it out anyway!


```suggestion
func (g *DataGathererDynamic) resourceMatchesExclusionKeys(item *api.GatheredResource) bool {
	res, ok := item.Resource.(*unstructured.Unstructured)
	if !ok {
		return false
	}

	return anyKeyMatches(res.GetAnnotations(), g.ExcludeAnnotKeys) ||
			anyKeyMatches(res.GetLabels(), g.ExcludeLabelKeys)
}
```

pkg/datagatherer/k8sdynamic/dynamic.go
```diff
@@ -309,6 +329,13 @@ func (c *ConfigDynamic) newDataGathererWithClient(ctx context.Context, cl dynami
 	}
 	newDataGatherer.registration = registration
 
+	for _, r := range c.ExcludeAnnotationKeysRegex {
+		newDataGatherer.ExcludeAnnotKeys = append(newDataGatherer.ExcludeAnnotKeys, regexp.MustCompile(r))
+	}
+	for _, r := range c.ExcludeLabelKeysRegex {
+		newDataGatherer.ExcludeLabelKeys = append(newDataGatherer.ExcludeLabelKeys, regexp.MustCompile(r))
+	}
```
SgtCoDFish, line 345: suggestion: It's kinda fair enough to use `MustCompile` here since we checked it in `validate()` - but this is a pretty complicated codebase which has evolved organically over time and it scares me a bit to use `Must` here just in case this function ever gets called without validation. I haven't checked if that's possible, and it probably isn't - but I still don't trust it, and it doesn't really hurt us to be extra defensive here.

I'd vote to also use `Compile` here, and return an error if it fails. That's a lot less catastrophic than the panic that would happen if `MustCompile` failed.

SgtCoDFish (changes requested):
Got a few bits and bobs although I think really this is close!

SgtCoDFish (approved):
/lgtm
/approve

Thank you!

PR #798, opened 2026-05-02
[VC-52458] Support both number and string for tsgID

inteon:
Based on feedback, this change updates tsgID to accept both numbers and strings.
Previously, the Helm chart’s values.schema.json enforced tsgID as a string. However, this constraint introduced unnecessary friction for users during initial setup.
To improve the user experience, we now allow both numeric and string inputs. As a side effect, this adds some complexity to the chart implementation, requiring a workaround for a Helm issue related to scientific notation handling.

context: in https://github.com/jetstack/jetstack-secure/pull/793 we enforced tsgID to be a string

```console
$ helm template ./deploy/charts/discovery-agent \
    --set config.tsgID=111111111111111111111111 \
    --set config.clusterName=aaa | grep 1111
            - "111111111111111111111111"
```

### DIFF
--- a/deploy/charts/discovery-agent/README.md
+++ b/deploy/charts/discovery-agent/README.md
@@ -6,13 +6,13 @@ The Discovery Agent connects your Kubernetes or OpenShift cluster to Palo Alto N
 
 <!-- AUTO-GENERATED -->
 
-#### **config.tsgID** ~ `string`
+#### **config.tsgID** ~ `number,string`
 > Default value:
 > ```yaml
 > ""
 > ```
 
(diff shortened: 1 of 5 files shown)


maelvls:
Works as expected, nice, thanks for fixing this.

For whaat it's worth, the string-based methods still work as expected:

```console
$ helm template ./deploy/charts/discovery-agent \
    --set config.tsgID=\\"111\\" \
    --set config.clusterName=aaa | grep 111
            - "111"
```

```console
$ helm template ./deploy/charts/discovery-agent \
    --set-string config.tsgID=111 \
    --set config.clusterName=aaa | grep 111
            - "111"
```

maelvls (approved):

PR #801, opened 2026-05-05
Prepare for v1.10.0-alpha.2 release; bump dependencies

FelixPhipps:
Adds support for the per-gatherer excludeAnnotationKeysRegex and excludeLabelKeysRegex config fields, which previously had no effect. Resources with annotation or label keys matching the configured patterns are now excluded entirely from upload, rather than just having those keys stripped from their metadata.


go.mod
```diff
@@ -1,120 +1,116 @@
...
 	github.com/spf13/pflag v1.0.10
 	github.com/stretchr/testify v1.11.1
-	golang.org/x/sync v0.19.0
+	golang.org/x/sync v0.20.0
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.34.3
-	k8s.io/apimachinery v0.34.3
-	k8s.io/client-go v0.34.3
-	k8s.io/component-base v0.34.3
-	sigs.k8s.io/controller-runtime v0.22.4
```
SgtCoDFish, line 26: note: I'm not fully confident that bumping these k8s deps will be totally safe given we'll want to also support older versions of k8s. We'll have to be careful to test this after we merge this PR, and check that the agent still works with the oldest k8s version we can find.

(I can't give a specific concern, but I think it's a focus point!)

internal/envelope/keyfetch/client_test.go
```diff
@@ -305,37 +305,8 @@ func TestClient_FetchKey(t *testing.T) {
...
-
-		client, _ := testClientSetup(t, server.URL)
-		key, err := client.FetchKey(t.Context())
-
-		require.NoError(t, err)
-		// Should skip the small key and return the valid one
-		assert.Equal(t, "valid-key", key.KeyID)
-	})
+	// Note: We used to test with smaller key sizes but the library started to reject such small keys.
+	// Therefore, we have removed the test for the time being.
```
SgtCoDFish, line 309: nb: I suggested replacing this test with a comment. No point in testing that _we_ reject small keys if the library does it for us.

SgtCoDFish (approved):
/lgtm
/approve

Thank you!

PR #803, opened 2026-05-07
Add helm unit tests for claimable certs behaviour

SgtCoDFish:
This just lets us be a little more confident about the behaviour here


George-Yanev (approved):
👍

PR #809, opened 2026-05-26
fix: Add default exclusions for GitOps tool annotations

kiril-cyberark:
Some GitOps tools (kapp, Rancher Fleet, Banzai Cloud) store a full copy of the original manifest in object annotations, which can cause sensitive data (such as Secret values) to be inadvertently included in data pushed to the platform. This change adds those annotation keys to the default exclusion list so they are stripped from all Kubernetes objects before being sent upstream.

### DIFF
--- a/deploy/charts/disco-agent/values.yaml
+++ b/deploy/charts/disco-agent/values.yaml
@@ -177,7 +177,10 @@ config:
   # to avoid YAML parsing issues with `\.`.
   #
   # Example: excludeAnnotationKeysRegex: ['^kapp\.k14s\.io/original.*']
-  excludeAnnotationKeysRegex: []
+  excludeAnnotationKeysRegex:
+    - '^kapp\.k14s\.io/original$'
+    - '^objectset\.rio\.cattle\.io/applied$'
+    - '^banzaicloud\.com/last-applied$'
   excludeLabelKeysRegex: []
 
   # A human readable name for the cluster where the agent is deployed (optional).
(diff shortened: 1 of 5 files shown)


deploy/charts/disco-agent/values.yaml
```diff
@@ -177,7 +177,10 @@ config:
   # to avoid YAML parsing issues with `\.`.
   #
   # Example: excludeAnnotationKeysRegex: ['^kapp\.k14s\.io/original.*']
-  excludeAnnotationKeysRegex: []
+  excludeAnnotationKeysRegex:
+    - '^kapp\.k14s\.io/original$'
+    - '^objectset\.rio\.cattle\.io/applied$'
+    - '^banzaicloud\.com/last-applied$'
```
SgtCoDFish: suggestion: I don't think adding these to helm values will do anything? If we add these keys to RedactFields they should always be excluded in any case

kiril-cyberark: removed the lines

SgtCoDFish (changes requested):
This is what I'd expect the fix to look like 👍 

I think we're definitely missing a description for this PR though - doesn't have to be super detailed (and it shouldn't contain any internal information - this repo is public!) but it's worth just adding a note about why this is being done.

kiril-cyberark:
> This is what I'd expect the fix to look like 👍
> 
> I think we're definitely missing a description for this PR though - doesn't have to be super detailed (and it shouldn't contain any internal information - this repo is public!) but it's worth just adding a note about why this is being done.

Description and title added

SgtCoDFish (approved):
/lgtm
/approve

Perfect, thank you!

PR #810, opened 2026-05-26
add _lastModifiedTime field to secret snapshots

EldarShalev:
Extract the most recent time from metadata.managedFields and include it as a synthetic _lastModifiedTime field on Secret objects sent to the backend. This enables detection of stale or unrotated secrets.

### DIFF
--- a/pkg/datagatherer/k8sdynamic/dynamic.go
+++ b/pkg/datagatherer/k8sdynamic/dynamic.go
@@ -381,6 +381,10 @@ type DataGathererDynamic struct {
 	// Encryptor, if non-nil, will be used to envelope encrypt Secret data.
 	// If nil, Secret data will be redacted.
 	Encryptor envelope.Encryptor
+
+	// IncludeLastModifiedTime, if true, extracts the most recent time from
+	// metadata.managedFields and includes it as _lastModifiedTime on Secrets.
+	IncludeLastModifiedTime bool
 }
 
 func (g *DataGathererDynamic) GVR() schema.GroupVersionResource {
@@ -555,7 +559,9 @@ func (g *DataGathererDynamic) redactList(ctx context.Context, list []*api.Gather
 						}
 					}
 
-					setLastModifiedTime(resource)
+					if g.IncludeLastModifiedTime {
+						setLastModifiedTime(resource)
+					}
 
 					// Redact to only selected fields
 					if err := Select(secretSelectedFields, resource); err != nil {
@@ -630,7 +636,8 @@ func setLastModifiedTime(resource *unstructured.Unstructured) {
 		return
 	}
 
-	var latestTime string
+	var latestTime time.Time
+	var latestTimeStr string
 	for _, entry := range managedFieldsRaw {
 		entryMap, ok := entry.(map[string]any)
 		if !ok {
@@ -640,16 +647,21 @@ func setLastModifiedTime(resource *unstructured.Unstructured) {
 		if !ok || timeVal == "" {
 			continue
 		}
-		if timeVal > latestTime {
-			latestTime = timeVal
+		parsed, err := time.Parse(time.RFC3339, timeVal)
+		if err != nil {
+			continue
+		}
+		if parsed.After(latestTime) {
+			latestTime = parsed
+			latestTimeStr = timeVal
 		}
 	}
 
-	if latestTime == "" {
+	if latestTimeStr == "" {
 		return
 	}
 
-	_ = unstructured.SetNestedField(resource.Object, latestTime, lastModifiedTimeFieldName)
+	_ = unstructured.SetNestedField(resource.Object, latestTimeStr, lastModifiedTimeFieldName)
 }
 
 // encryptDataField encrypts the `data` field of the given secret and stores the encrypted data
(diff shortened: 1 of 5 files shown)


pkg/datagatherer/k8sdynamic/dynamic.go
```diff
@@ -555,6 +555,8 @@ func (g *DataGathererDynamic) redactList(ctx context.Context, list []*api.Gather
 						}
 					}
 
+					setLastModifiedTime(resource)
```
SgtCoDFish: suggestion: this unilaterally sets the last modified time in all cases. This repo contains 3 agents which we release, and this change is only for the disco-agent and not the others. I think we'd need to change this PR so it only affects the disco agent.

pkg/datagatherer/k8sdynamic/dynamic.go
```diff
@@ -617,6 +619,39 @@ const encryptedDataFieldName = "_encryptedData"
...
+		if !ok {
+			continue
+		}
+		timeVal, ok := entryMap["time"].(string)
+		if !ok || timeVal == "" {
+			continue
+		}
+		if timeVal > latestTime {
+			latestTime = timeVal
+		}
```
SgtCoDFish: suggestion: treating the values as strings means that Go will compare lexicographically, not by the times represented.

It should be possible to write a test case which detects the wrong latestTime based on this comparison. You'll have to parse the timestamps into time.Time values to compare them safely + properly.

pkg/datagatherer/k8sdynamic/dynamic.go
maelvls: Looks like this was fixed. Closing.

pkg/datagatherer/k8sdynamic/dynamic.go
maelvls: > Go will compare lexicographically

I guess it would be OK to compare lexicographically if the RFC3339-formatted times were all using the same timezone... but we don't know that, so better not compare without parsing.

I see that `timeVal` and `latestTime` are now parsed and compared using `After`. Closing.

maelvls:
Ash's feedback has been addressed, I think we are good to go.

To test this manually, I went to https://tlskp-test.integration-cyberark.cloud/, logged in using mael.valais@cyberark.cloud.420375, then used these creds to do:

```
export ARK_USERNAME=mael.valais@cyberark.cloud.420375
export ARK_SECRET=REDACTED
export ARK_PLATFORM_DOMAIN=integration-cyberark.cloud
export ARK_SUBDOMAIN=tlskp-test
export ARK_DISCOVERY_API=https://platform-discovery.integration-cyberark.cloud
go run ./cmd/ark agent --machine-hub -c config.yaml --one-shot
```

I used mitmproxy to see what payload is being sent, and I was able to confirm that the new `_lastModifiedTime` was being sent:

```yaml
{
  "agent_version": "development",
  "cluster_id": "db43366f-d7db-4c27-b384-985db34aefa9",
  "cluster_name": "mael.valais@cyberark.cloud.420375",
  "k8s_version": "v1.35.0",
  "openid_configuration": {
    "id_token_signing_alg_values_supported": ["RS256"],
    "issuer": "https://kubernetes.default.svc.cluster.local",
    "jwks_uri": "https://172.18.0.2:6443/openid/v1/jwks",
    "response_types_supported": ["id_token"],
    "subject_types_supported": ["public"]
  },
  "jwks": {
    "keys": [
      {
        "alg": "RS256",
        "e": "AQAB",
        "kid": "CHm2bo7FnsdchfAyB-2RV3LLlYDsab8AW8pZfxrmqI0",
        "kty": "RSA",
        "n": "n-xg_DuZuQtQDoseuyal00Q6Hoi_tZZ0NrfKL0oaaUP3hPtldZn5g2u9ak1L9j5sgxp34FhTs9b98B9NgHIuvrwiRiEkIaYhcJT8LnF3qMiM-FQgFuv2REtHXwY-fL8_HZBhHuLkYlg3zG1mPPUss0cBV3iK8y3M0GZ7o_WQTUHXo_NoR56qE_oGAwF4TZfTN17TN3CkjrVCOfOJQS6OGFL6nKa8g5qZgUszPcU4z0FrAiPqKRKuuFkFsypxidh_6--CpSDrjzmg0CIp5onf8UESqhel_3rb8P7RURoBODb6Cl4I1in4FEDJUezVDDz_X1I-bMy0GP5NgddPU22sVw",
        "use": "sig"
      }
    ]
  },
  "secrets": [
    {
      "_lastModifiedTime": "2026-05-28T13:24:36Z",
      "apiVersion": "v1",
      "kind": "Secret",
      "metadata": {
        "creationTimestamp": "2026-05-28T13:24:36Z",
        "name": "agent-credentials",
        "namespace": "ark",
        "resourceVersion": "5833",
        "uid": "0bb44bd2-4120-482f-b26c-19d71249b54e"
      },
      "type": "Opaque"
    },
    {
      "_lastModifiedTime": "2026-05-28T13:24:26Z",
      "apiVersion": "v1",
      "kind": "Secret",
      "metadata": {
        "creationTimestamp": "2026-05-28T13:24:26Z",
        "name": "agent-credentials",
        "namespace": "default",
        "resourceVersion": "5818",
        "uid": "c0781cce-3a55-4e96-b0e7-c1df32e181d5"
      },
      "type": "Opaque"
    }
  ],
  # ...
```

maelvls (approved):

PR #811, opened 2026-06-02
Add VenafiConnection support for NGTS

inteon:
Adds support for authenticating using a VenafiConnection resource for NGTS.

Also unlocks OIDC authentication for NGTS.


deploy/charts/venafi-kubernetes-agent/README.md
```diff
@@ -390,7 +390,10 @@ Configure VenafiConnection authentication
 > false
 > ```
 
-When set to true, the Discovery Agent will authenticate to CyberArk Certificate Manager using the configuration in a VenafiConnection resource. Use `venafiConnection.enabled=true` for [secretless authentication](https://docs.cyberark.com/mis-saas/vaas/k8s-components/t-install-tlspk-agent/). When set to true, the `authentication.secret` values will be ignored and the. Secret with `authentication.secretName` will _not_ be mounted into the  
+When set to true, the Discovery Agent will authenticate to its upload backend using the configuration in a VenafiConnection resource. The backend is determined by the VenafiConnection's spec: use `spec.vcp` for. CyberArk Certificate Manager (CMSaaS), or `spec.ngts` (with `tsgID` or
```
wallrj-cyberark: Nit: there are two stray periods in this description — `...for. CyberArk Certificate Manager` and `...and the. Secret with`. The first was introduced in this PR; the second was pre-existing but now the whole paragraph is being touched.

Suggested fix:
```
...use `spec.vcp` for CyberArk Certificate Manager (CMSaaS), or...
...will be ignored and the Secret with...
```

deploy/charts/venafi-kubernetes-agent/values.schema.json
```diff
@@ -137,7 +137,7 @@
...
+      "description": "When set to true, the Discovery Agent will authenticate to its upload backend using the configuration in a VenafiConnection resource. The backend is determined by the VenafiConnection's spec: use `spec.vcp` for. CyberArk Certificate Manager (CMSaaS), or `spec.ngts` (with `tsgID` or\n`url`, and a `jwt` source) for NGTS / Palo Alto Networks. `spec.tpp` and\n`spec.vcp.apiKey` are rejected by the agent.\nUse `venafiConnection.enabled=true` for [secretless authentication](https://docs.cyberark.com/mis-saas/vaas/k8s-components/t-install-tlspk-agent/). When set to true, the `authentication.secret` values will be ignored and the. Secret with `authentication.secretName` will _not_ be mounted into the\nDiscovery Agent Pod.",
```
wallrj-cyberark: Same stray-period issue here (`...for. CyberArk Certificate Manager` and `...and the. Secret with`). This is the JSON schema description — if it is auto-generated from `values.yaml`, fixing the source should be sufficient.

deploy/charts/discovery-agent/README.md
```diff
@@ -84,8 +84,30 @@ The name of the Secret containing the NGTS built-in service account credentials.
...
+#### **config.venafiConnection.enabled** ~ `bool`
+> Default value:
+> ```yaml
+> false
+> ```
+
+When set to true, config.tsgID, config.serverURL, config.clientID and config.clientId must not be set (the chart will fail to render otherwise), and the Secret named by config.secretName will _not_ be mounted into the Discovery Agent Pod.
+#### **config.venafiConnection.name** ~ `string`
+> Default value:
+> ```yaml
```
wallrj-cyberark, line 99: Nit: there is no blank line between the `config.venafiConnection.enabled` and `config.venafiConnection.name` headings (line 37-38). The other fields in this README have blank lines separating them. Is the README auto-generated from `values.yaml`? If so this might be a generator quirk.

deploy/charts/discovery-agent/templates/deployment.yaml
```diff
@@ -71,6 +71,24 @@ spec:
             - "agent"
             - "-c"
             - "/etc/discovery-agent/config.yaml"
+            {{- if .Values.config.venafiConnection.enabled }}
+            {{- if .Values.config.tsgID }}
+              {{- fail "config.tsgID must not be set when config.venafiConnection.enabled is true; the TSG ID is read from the VenafiConnection's spec.ngts" }}
```
wallrj-cyberark, line 76: The `config.secretName` field is silently ignored in VenafiConnection mode (the Secret volume mount is skipped) but is not rejected by `{{- fail }}` like `tsgID`, `serverURL`, `clientID`, and `clientId` are.

Is that intentional? The documentation says "Ignored when config.venafiConnection.enabled is true" (softer than the "Must not be set" wording on the other fields), so it seems deliberate. But a user who sets `secretName` alongside `venafiConnection.enabled=true` might not realise their credentials Secret is not being used.

Worth considering either:
- Adding a `{{- fail }}` if `secretName` differs from the default when VenafiConnection mode is on, or
- Leaving it as-is (current behaviour) if you want to allow a smooth migration where both are configured

pkg/client/client_venconn.go
```diff
@@ -137,17 +139,25 @@ func (c *VenConnClient) PostDataReadingsWithOptions(ctx context.Context, reading
...
-		return fmt.Errorf(`VenafiConnection %s/%s: the agent cannot be used with TPP`, c.venConnNS, c.venConnName)
-	}
-	if details.VCP != nil && details.VCP.APIKey != "" {
+	server := details.ServerDetails()
+	switch server.BackendAuth {
+	case connection_details.VCPAccessToken, connection_details.NGTSAccessToken:
+		// Supported.
+	case connection_details.TPPAccessToken:
+		return fmt.Errorf(`VenafiConnection %s/%s: the agent does not support the TPP backend`, c.venConnNS, c.venConnName)
+	case connection_details.DistributedIssuerAccessToken:
```
wallrj-cyberark, line 148: Good: the explicit `switch` on `server.BackendAuth` with a `default` case is a nice improvement over the old nil-check pattern — it will fail closed if `venafi-connection-lib` adds a new backend type.

pkg/client/client_venconn.go
```diff
@@ -171,21 +182,32 @@ func (c *VenConnClient) PostDataReadingsWithOptions(ctx context.Context, reading
...
 
+	req.Header.Set("Accept", "application/json")
 	req.Header.Set("Content-Type", "application/json")
-	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", details.VCP.AccessToken))
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
 	version.SetUserAgent(req)
 
-	q := req.URL.Query()
-	q.Set("name", opts.ClusterName)
+	isNGTS := server.BackendAuth == connection_details.NGTSAccessToken
```
wallrj-cyberark: The `bluemonday.StrictPolicy()` sanitiser is created on every call to `PostDataReadingsWithOptions`. Since `StrictPolicy()` is documented as safe to reuse, it could be a package-level `var` or a field on the struct. Minor efficiency point — not a correctness issue.

pkg/agent/config.go
```diff
@@ -386,13 +397,13 @@ func InitAgentCmdFlags(c *cobra.Command, cfg *AgentCmdFlags) {
...
 
 const (
-	JetstackSecureOAuth         OutputMode = "Jetstack Secure OAuth"
-	JetstackSecureAPIToken      OutputMode = "Jetstack Secure API Token"
-	VenafiCloudKeypair          OutputMode = "Venafi Cloud Key Pair Service Account"
-	VenafiCloudVenafiConnection OutputMode = "Venafi Cloud VenafiConnection"
-	LocalFile                   OutputMode = "Local File"
-	MachineHub                  OutputMode = "MachineHub"
-	NGTS                        OutputMode = "NGTS"
+	JetstackSecureOAuth    OutputMode = "Jetstack Secure OAuth"
```
wallrj-cyberark, line 400: The rename from `VenafiCloudVenafiConnection` to `VenafiConnection` changes the string value from `"Venafi Cloud VenafiConnection"` to `"VenafiConnection"`. This string appears in user-facing log messages and error text.

Is there any downstream tooling (monitoring alerts, log parsers, documentation) that matches on the old string `"Venafi Cloud VenafiConnection"`? If so, a heads-up in the PR description would be helpful.

pkg/agent/config_test.go
```diff
@@ -642,7 +642,20 @@ func Test_ValidateAndCombineConfig(t *testing.T) {
 			`)),
 			withCmdLineFlags("--venafi-connection", "venafi-components"))
 		require.NoError(t, err)
-		assert.Equal(t, VenafiCloudVenafiConnection, got.OutputMode)
+		assert.Equal(t, VenafiConnection, got.OutputMode)
+	})
+
+	t.Run("venafi-cloud-workload-identity-auth: --venafi-cloud is tolerated alongside --venafi-connection for backwards compatibility with older rendered charts", func(t *testing.T) {
```
wallrj-cyberark, line 648: Nice — this backwards-compatibility test ensures that existing Helm-rendered manifests with `--venafi-cloud --venafi-connection` continue to work after the `--venafi-cloud` flag is moved inside the keypair-mode block. Good forward-thinking.

wallrj-cyberark (commented):
## Review Summary

This PR adds VenafiConnection support for NGTS to both the `discovery-agent` and `venafi-kubernetes-agent` Helm charts, and updates the Go code to support the new `NGTSAccessToken` backend type from `venafi-connection-lib`. The scope is clear and the implementation is well-structured.

**What this PR does:**
- Adds `config.venafiConnection` values to the discovery-agent chart, with mutual exclusivity guards (`{{- fail ...}}`) for legacy keypair fields
- Renames `VenafiCloudVenafiConnection` → `VenafiConnection` throughout Go code, reflecting that VenafiConnection mode is no longer VCP-only
- Updates `client_venconn.go` to use the `connection_details` package from `venafi-connection-lib` (replacing the old `details.VCP` / `details.TPP` struct pattern), supporting both `VCPAccessToken` and `NGTSAccessToken`
- Moves `--venafi-cloud` inside the `{{- else }}` block in the venafi-kubernetes-agent template so it is only rendered in keypair mode
- Adds Helm unit tests for both charts covering VenafiConnection mode and mutual exclusivity rejections
- Bumps `venafi-connection-lib` to a pre-release with NGTS support

**Findings:**
- Two stray periods in the `venafi-kubernetes-agent` README description (carried over from the original, but now compounded)
- The `config.secretName` field is not rejected by `{{- fail }}` in VenafiConnection mode — it is silently ignored. Worth confirming this is intentional
- Minor cosmetic nit in the discovery-agent README

**Jira issue:** I searched Jira extensively but could not find an exact issue for this work. The closest candidates are:
- **VC-52460** — "NGTS cert-manager team component improvements (Fast Follows)" (broad epic, In Progress)
- **VC-35747** — "Make it possible to share a single VenafiConnection across the three components" (closed)
- **VC-31964** — "AGENT: Secretless authentication (OIDC) to VCP" (closed)

None of these are a direct match. You may want to create a new issue or ask Tim which one this is tracked under.

**Prior art:** PR #644 (`Make venafiConnection a root-level Helm option`) is related but tackles the venafi-kubernetes-agent chart structure only and is still open. This PR appears complementary rather than overlapping.

**Overall:** The changes are well-tested, the mutual exclusivity guards are thorough with clear error messages, and the backwards-compatibility test (`--venafi-cloud` tolerated alongside `--venafi-connection`) is a nice touch. A few documentation nits inline.

*Review conducted with assistance from Claude (Opus 4.6)*

pkg/agent/config.go
inteon: I changed this because `VenafiConnection` can be used for `Venafi Cloud` and for `NGTS`.
There should be no downstream dependencies on this.

pkg/client/client_venconn.go
inteon: This pattern was already used in the venafi_cloud and ngts clients.
We can fix this in a follow-up.

deploy/charts/discovery-agent/templates/deployment.yaml
inteon: I added a fail

deploy/charts/discovery-agent/README.md
inteon: Yes, this is auto-generated and probably a quirk.

wallrj-cyberark (approved):
LGTM — review comments addressed.

*Review conducted with assistance from Claude (Opus 4.6)*

PR #812, opened 2026-06-02
Make --tsg-id and --ngts-server-url mutually exclusive

inteon:
When providing the NGTS server URL, the TSGID is not required.
Requiring and even allowing both to be provided is confusing.

Additionally, I updated the GH action secrets to connect to a working QA tenant (this was not working before - https://github.com/jetstack/jetstack-secure/pull/790).

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -146,9 +146,9 @@ jobs:
       - run: make -j ngts-test-e2e
         env:
           OCI_BASE: ${{ secrets.NGTS_OCI_BASE }}
-          NGTS_CLIENT_ID: ${{ secrets.NGTS_CLIENT_ID }}
+          NGTS_CLIENT_ID: e3c8bde7-5f13-11f1-99f4-5e067e231041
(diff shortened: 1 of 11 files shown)


mladen-rusev-cyberark (approved):

.github/workflows/tests.yaml
```diff
@@ -146,9 +146,9 @@ jobs:
       - run: make -j ngts-test-e2e
         env:
           OCI_BASE: ${{ secrets.NGTS_OCI_BASE }}
-          NGTS_CLIENT_ID: ${{ secrets.NGTS_CLIENT_ID }}
+          NGTS_CLIENT_ID: e3c8bde7-5f13-11f1-99f4-5e067e231041
           NGTS_PRIVATE_KEY: ${{ secrets.NGTS_PRIVATE_KEY }}
-          NGTS_TSG_ID: ${{ secrets.NGTS_TSG_ID }}
+          NGTS_TSG_URL: https://1806660206.ngts.qa.venafi.io
```
inteon, line 151: Putting this in a secret causes the value to be redacted from the test output.
For debugging, I find it much easier when this value is not redacted.

PR #816, opened 2026-06-18
add CSI information to discovery-agent

YaaraSirkis:


maelvls:
Looks straightforward. I haven't tested with with an actual Kubernetes cluster with a Secrets Store CSI Driver installed, though.

pkg/datagatherer/k8sdynamic/dynamic.go
```diff
@@ -567,11 +567,18 @@ func (g *DataGathererDynamic) redactList(ctx context.Context, list []*api.Gather
...
-				} else if gvk.Kind == "Route" && gvk.Group == "route.openshift.io" {
-					// route object
+				case gvk.Kind == "Route" && gvk.Group == "route.openshift.io":
 					if err := Select(RouteSelectedFields, resource); err != nil {
 						return err
 					}
+				case gvk.Kind == "SecretProviderClass" && gvk.Group == "secrets-store.csi.x-k8s.io":
+					if err := Select(SecretProviderClassSelectedFields, resource); err != nil {
+						return err
+					}
```
maelvls: Is there really a point in redacting fields from SecretProviderClass resources (beyond the regular redacted metadata/labels)?

The reason we have `Select` for Routes and Secrets is because they do contain confidential information (`data.tls\.key` in secrets, `spec.tls.key` in routes). Are you sure SecretProviderClass or SecretProviderClassPodStatus have some secrets?

We use a "default allow" approach across all other resources (pods, etc). If we need to switch to "default deny", let's talk about it, because it might be worth doing that with all resources.

maelvls:
LGTM!

maelvls (approved):

PR #817, opened 2026-07-01
feat(agent): add SMS/Conjur JWT authentication alongside username/password

roeezis:
## Summary

Add a **workload-identity JWT** auth path to the agent, **alongside** the existing CyberArk Identity **username/password** login — not replacing it. Both methods coexist and the agent **auto-selects by config**. The agent reads a projected ServiceAccount token, exchanges it at Secrets Manager (Conjur Cloud) `authn-jwt` for a short-lived Conjur access token, and presents it as a `Bearer` credential. Only the token source changes, behind the existing `identity.RequestAuthenticator` seam; `dataupload` / service discovery / upload pipeline untouched.

## Auth-method selection (backward compatibility)

```
config.cyberark.serviceId set?           -> Conjur JWT exchange      [preferred]
else ARK_USERNAME + ARK_SECRET present?  -> legacy username/password [GA installs]
else                                     -> fail closed at startup
both set                                 -> serviceId wins (warn, ignore UP)
```

serviceId-wins lets a migrating install add the service-id before removing old creds — no rollout window with no method. Existing GA installs (UP creds, no serviceId) keep working with zero config change.

## What changed

- **`internal/cyberark/conjur/`** — authn-jwt exchange, short-lived token cache, sets `Bearer`. Implements `RequestAuthenticator`. Requests the base64-encoded access token explicitly (Conjur's canonical wire form); it is forwarded as an opaque credential.
- **`internal/cyberark/jwtsource/`** — `file` source (projected SA token, default `/var/run/secrets/tokens/jwt`); `spiffe` deferred.
- **`internal/cyberark/identity/username_password.go`** — legacy UP-login retained, behaviour unchanged.
- **`internal/cyberark/client.go`** — `selectAuthenticator` picks the method by config. `account="conjur"` confirmed correct (was a POC placeholder — verified against a live tenant).
- **`internal/cyberark/servicediscovery/`** — parse the `secrets_manager` service from service discovery into a new `Services.SecretsManager` field, and use it as the base URL for the authn-jwt exchange. See the note below.
- **`internal/envelope/keyfetch/`** — JWKS fetch uses the same authenticator selector.
- **`pkg/agent/config.go`** — MachineHub validation no longer hard-requires `service_id`; runtime fails closed if no method configured.
- **helm** — optional `ARK_USERNAME`/`ARK_SECRET` env alongside the projected `aud=conjur` token volume; README + values document selection. `config.cyberark.serviceId` must be the bare authenticator service-id segment, not the full policy path.

## The authn-jwt exchange targets `secrets_manager`, not `identity_administration`

Worth a reviewer's attention, since it is the one non-obvious wiring detail here. `authn-jwt/<service-id>/<account>/authenticate` is served by **Secrets Manager**, which service discovery reports as a different host from `identity_administration`:

```
identity_administration  https://<identity-id>.id.<env>-cyberark.cloud
secrets_manager          https://<subdomain>.secretsmgr.<env>-cyberark.cloud/api
```

An earlier revision of this branch passed `Identity.API` as the Conjur base URL, so every live exchange hit a host that does not serve the endpoint. Fixed by resolving `secrets_manager` explicitly and failing closed when it is absent. The username/password path still uses `Identity.API`, unchanged.

This class of bug was invisible to the unit tests: the mocks wired the Conjur exchange server *into* `Identity.API`, so the mock followed whichever field the code happened to read and both the bug and the fix passed. The mocks now point `Identity` and `SecretsManager` at deliberately distinct hosts, and `discovery_test.go` asserts the two differ, so reading the wrong one fails.

## Onboarding

Tenant onboarding — the `authn-jwt` authenticator scoped to the cluster's OIDC issuer and JWKS, a registered workload for the agent's ServiceAccount, and the grants that let it authenticate and upload — is performed through the CyberArk web console and is out of scope for this repo. The chart README states what onboarding must produce and points at the product documentation.

Onboarding needs only the tenant administrator's own credentials. The agent holds no Conjur identity beyond its projected ServiceAccount token, and nothing in this chart requires a Conjur admin credential at deploy time.

## Tests

Full suite green locally (`go build ./...`, `go test ./...` with envtest assets, `gofmt` clean, `helm lint` clean), including the 4 auth-selection cases (serviceID→conjur, UP→identity, both→conjur wins, neither→error), the retained UP-login tests, and the new service-discovery assertion described above.

Validated end-to-end against a live Secrets Manager tenant: the authn-jwt exchange returns a token and the agent completes gather → minimize → upload.


roeezis:
Splitting this into a stacked series of smaller, single-concern PRs for reviewability:

1. #818 — jwtsource package
2. #819 — split legacy username/password login into its own file
3. #820 — resolve secrets_manager alongside identity_administration
4. #821 — Conjur JWT authentication client
5. #822 — select Conjur JWT or legacy username/password authenticator
6. #823 — wire Conjur JWT config into the top-level CyberArk client
7. #824 — Helm chart support for Conjur JWT authentication

Each is stacked on master and carries the cumulative diff of the PRs before it in the series — that diff will shrink to just its own change as earlier PRs merge. Closing this one in favor of the series.

PR #818, opened 2026-08-23
feat(agent): add jwtsource package for reading projected SA tokens

roeezis:
## Summary
Part 1 of the SMS/Conjur JWT authentication series (split out of #817 for reviewability).

Introduces a small, isolated interface for reading a JWT from a file path — the first piece of the upcoming Conjur JWT authentication path. Nothing else depends on it in this PR; the Conjur client that uses it lands later in the stack.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/cyberark/jwtsource/...`

### DIFF
--- a/internal/cyberark/jwtsource/jwtsource_test.go
+++ b/internal/cyberark/jwtsource/jwtsource_test.go
@@ -1,0 +1,32 @@
+// internal/cyberark/jwtsource/jwtsource_test.go
+package jwtsource
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFileSource_ReadsToken(t *testing.T) {
+	dir := t.TempDir()
+	p := filepath.Join(dir, "jwt")
+	require.NoError(t, os.WriteFile(p, []byte("the-jwt\n"), 0o600))
+	got, err := NewFileSource(p).Read(t.Context())
+	require.NoError(t, err)
+	require.Equal(t, "the-jwt", got) // trimmed
+}
+
+func TestFileSource_MissingFile(t *testing.T) {
+	_, err := NewFileSource("/no/such/file").Read(t.Context())
+	require.Error(t, err)
+}
+
+func TestFileSource_EmptyFile(t *testing.T) {
+	dir := t.TempDir()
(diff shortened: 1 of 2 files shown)


mladen-rusev-cyberark (approved):

internal/cyberark/jwtsource/jwtsource_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFileSource_ReadsToken(t *testing.T) {
+	dir := t.TempDir()
+	p := filepath.Join(dir, "jwt")
+	require.NoError(t, os.WriteFile(p, []byte("the-jwt\n"), 0o600))
+	got, err := NewFileSource(p).Read(context.Background())
```
SgtCoDFish: suggestion: no reason to use context.Background in tests, t.Context() is the way to go!

roeezis:
Fixed — switched both tests to `t.Context()`.

roeezis:
@mladen-rusev-cyberark 
Re: stacking the base branches so each PR shows only its own diff — looked into it, and there's a real constraint: GitHub requires a PR's base ref to live in that PR's own base repository. #819-824 are already opened against jetstack/jetstack-secure, and the intermediate branches (split/01...split/06) only exist on my fork — so I can't retarget them after the fact.

The only way to get true stacking would be to close #819-824 and reopen them on the fork itself (roeezis/jetstack-secure), each based on the previous branch — fully public and reviewable without needing write access to my fork. That loses the existing review-thread history on those 6 PRs though (several rounds of your feedback already resolved on them), so I didn't want to do that without flagging the tradeoff first.

In the meantime: each PR's diff already shrinks to just its own delta once its predecessors merge (base is master, so the diff is against master's current state) — noted in each PR's description. If a maintainer merges these in order (818 → 819 → ... → 824), the stacking problem resolves itself without needing the close/reopen. Let me know if you'd rather I do the recreate anyway, or if merging in order as they're approved works for you.

internal/cyberark/jwtsource/jwtsource_test.go
roeezis: Fixed — switched both tests to `t.Context()`.

PR #819, opened 2026-08-23
feat(agent): split legacy username/password login into its own file

roeezis:
## Summary
Part 2 of the SMS/Conjur JWT authentication series (split out of #817 for reviewability). Stacked on #818 — includes #818's commit until that merges, so the diff here will shrink to just this PR's own change once #818 lands.

`identity.go` mixed shared client/token-cache plumbing with the CyberArk Identity username/password (UP) login flow. Moves the UP-specific code into `username_password.go` so the shared plumbing stays easy to find once a second login mechanism (Conjur JWT) is added alongside it.

No behavior change — pure extraction, plus unexporting `ActionAnswer` and exporting the mock's success credentials for other packages' tests.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/cyberark/identity/...`


internal/cyberark/identity/mock.go
```diff
@@ -32,6 +32,18 @@ const (
 	// mock server in response to a successful AdvanceAuthentication request
 	// Must match what's in testdata/advance_authentication_success.json
 	mockSuccessfulStartAuthenticationToken = "success-token"
+
+	// actionAnswer is the string sent to an AdvanceAuthentication request to indicate we're
+	// providing credentials as plain text.
+	actionAnswer = "Answer"
```
mladen-rusev-cyberark: The commit message says `ActionAnswer` was unexported, but it's still exported at `username_password.go:35` and still referenced there (`:328`) and by `advance_authentication_test.go` / `start_authentication_test.go`. This adds a *second* constant with the same value rather than unexporting the first — two literals for one protocol string in one package will drift.

Simplest fix for a "pure extraction" PR: delete this const and have `mock.go:228` keep using `ActionAnswer`. Either way the commit message needs correcting — as written it describes a change that didn't happen.

mladen-rusev-cyberark (commented):
Verified the "no behaviour change" claim mechanically — sorting all lines of the pre-image `identity.go` against the sorted union of the post-image `identity.go` + `username_password.go` yields only comment/import differences, no code changes. The split is a good idea and lands cleanly.

One correction needed (inline), plus: the Python SDK reference link dropped from the `Client` doc comment is the only pointer in the repo to the reference implementation this client mirrors — worth moving to `username_password.go` rather than deleting.

Reviewed as this PR's own commit (`2483110`) rather than the cumulative diff against master.

roeezis:
Fixed — deleted the duplicate `actionAnswer` const, `mock.go` now uses the single exported `ActionAnswer`. Also corrected the commit message.

mladen-rusev-cyberark (approved):

internal/cyberark/identity/mock.go
roeezis: Fixed — deleted the duplicate `actionAnswer` const, `mock.go` now uses the single exported `ActionAnswer`. Also corrected the commit message.

PR #820, opened 2026-08-23
fix(agent): resolve secrets_manager alongside identity_administration

roeezis:
## Summary
Part 3 of the SMS/Conjur JWT authentication series (split out of #817). Stacked on #818, #819 — diff will shrink once those merge.

The Service Discovery API returns several independently-hosted services; the authn-jwt exchange this series adds is served by `secrets_manager`, a different host from `identity_administration`. Adds a `SecretsManager` field to `Services` and parses it, so callers have it available — nothing reads it yet, that lands in a later PR alongside the client that needs it.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/cyberark/servicediscovery/... ./internal/envelope/...`


internal/cyberark/servicediscovery/discovery.go
```diff
@@ -180,6 +189,13 @@ func (c *Client) DiscoverServices(ctx context.Context) (*Services, string, error
 					break
 				}
 			}
+		case SecretsManagerServiceName:
```
mladen-rusev-cyberark, line 193: Following on from this new case: `secretsManagerAPI` is deliberately *not* validated in the block below (`discovery.go:202`, where `identityAPI == ""` is rejected). I think that's right — requiring it would break every existing username/password install on a tenant not yet onboarded to Conjur, and #822 validates it at point of use (`internal/cyberark/client.go:111`).

Could you make that explicit with a comment there? Something like *"`secrets_manager` is intentionally not required here — only the Conjur JWT path needs it, and it validates at use in `cyberark.selectAuthenticator`"*. Otherwise a future reader "fixes" the asymmetry and breaks legacy tenants.

While you're in that block, the `//TODO: Should add a check for discoveryContextAPI too?` at `discovery.go:206` now sits adjacent to a third unchecked field — worth answering or dropping.

Minor, separate: these three `case` bodies are now identical six-line copies. A small `mainActiveAPI(eps []ServiceEndpoint) string` helper would collapse all three. Not worth blocking on, but the next service added makes it four.

mladen-rusev-cyberark (commented):
Correct and minimal. The `assert.NotEqual(t, services.Identity.API, services.SecretsManager.API)` assertion is the right one — it defends against the exact failure mode where a downstream mock points at whichever field the code reads, so tests pass and production 404s.

Approving in spirit; the one comment below is a docs request, not a change request.

roeezis:
All three addressed:
- Added a comment above the identityAPI check making explicit that secretsManagerAPI/discoveryContextAPI are intentionally not required there, with a note that not every caller validates discoveryContextAPI yet (keyfetch doesn't).
- Answered the discoveryContextAPI TODO in that same comment rather than leaving it as an open question.
- Factored the three identical endpoint-selection loops into a `mainActiveAPI` helper.

internal/cyberark/servicediscovery/discovery.go
```diff
@@ -163,35 +183,37 @@ func (c *Client) DiscoverServices(ctx context.Context) (*Services, string, error
...
-					break
-				}
-			}
+			discoveryContextAPI = mainActiveAPI(svc.Endpoints)
+		case SecretsManagerServiceName:
+			secretsManagerAPI = mainActiveAPI(svc.Endpoints)
 		}
 	}
 
 	if identityAPI == "" {
```
mladen-rusev-cyberark, line 202: Small nit here which will be mentioned in #822 as I think it belongs there and not here

mladen-rusev-cyberark (approved):

roeezis:
Added the note above `identityAPI == ""` per your first optional follow-on, and updated `pkg/testutil/envtest.go:291`'s comment to cite it instead of reading like a workaround.

internal/cyberark/servicediscovery/discovery.go
roeezis: All three addressed:
- Added a comment above the identityAPI check making explicit that secretsManagerAPI/discoveryContextAPI are intentionally not required there, with a note that not every caller validates discoveryContextAPI yet (keyfetch doesn't).
- Answered the discoveryContextAPI TODO in that same comment rather than leaving it as an open question.
- Factored the three identical endpoint-selection loops into a `mainActiveAPI` helper.

internal/cyberark/servicediscovery/discovery.go
roeezis: Added the note above `identityAPI == ""` per your first optional follow-on, and updated `pkg/testutil/envtest.go:291`'s comment to cite it instead of reading like a workaround.

PR #821, opened 2026-08-23
feat(agent): add Conjur JWT authentication client

roeezis:
## Summary
Part 4 of the SMS/Conjur JWT authentication series (split out of #817). Stacked on #818, #819, and the service-discovery PR — diff will shrink once those merge.

Exchanges a projected ServiceAccount JWT (via the jwtsource package from #818) for a Conjur access token through the authn-jwt endpoint, and authenticates requests with it as `identity.RequestAuthenticator`. Nothing wires this in yet — that's the next PR, once both this and the legacy identity client exist side by side.

The identity returned for audit tagging is the token's own `sub` claim when it can be extracted, falling back to the configured service ID otherwise, so an unusual token shape degrades rather than fails the request.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/cyberark/conjur/...`
- [x] Live end-to-end verified against a real CyberArk tenant: authn-jwt exchange, snapshot-links, and S3 upload all succeed.


internal/cyberark/conjur/conjur.go
```diff
@@ -0,0 +1,161 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"k8s.io/klog/v2"
+
+	"github.com/jetstack/preflight/internal/cyberark/jwtsource"
+)
+
+const tokenTTL = 8 * time.Minute
```
mladen-rusev-cyberark: Conjur access tokens default to an 8-minute lifetime and this is set to exactly that, with no margin. `c.tokenTime` is stamped *after* the exchange returns (later than real issue time), so the client will serve a token at 7m59s — network latency, a slow upload, or minor clock skew then puts the request past expiry. Result is a sporadic 401 with no retry.

You're already decoding the payload for `sub` in `identityFromToken` — reading `exp` in the same pass and driving the cache off it would remove the guess entirely, keeping this constant as the fallback for the undecodable case.

Please also make the TTL a `Client` field rather than a package const, so expiry is testable. There's currently no test that the cache is *invalidated*, only that it's used.

mladen-rusev-cyberark (commented):
Shape is right — plugs into the existing `RequestAuthenticator` seam, fails closed on non-200, and unlike the legacy identity client it refreshes its own token. Main concern is the token lifetime handling (inline).

Two smaller things: `New` and `MockConjurExchangeServerStatus` are missing doc comments; and `buildJWSToken` only emits `base64.StdEncoding`, so three of the four encodings `conjurTokenObject` claims to tolerate (raw JSON, URL-safe base64, unpadded) are never exercised — a table test over all four would be cheap.

internal/cyberark/conjur/conjur.go
```diff
@@ -0,0 +1,161 @@
...
+// AuthenticateRequest implements identity.RequestAuthenticator.
+// It exchanges the JWT for a Conjur access token, sets the Authorization
+// header, and returns an identity string for audit tagging. The identity is
+// the token's own `sub` claim when it can be extracted; otherwise it falls
+// back to the configured service ID so a token in an unexpected shape never
+// fails the request.
+func (c *Client) AuthenticateRequest(req *http.Request) (string, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.token == "" || time.Since(c.tokenTime) >= tokenTTL {
```
mladen-rusev-cyberark: No path to re-authenticate on a 401 from the resource server. If the cached token is rejected early (Conjur restart, authenticator toggled, clock skew), this only ever looks at wall-clock age, so callers get 401s until the TTL rolls. Worth exposing an `Invalidate()` the callers can use on 401, or retrying the exchange once.

Separately: the mutex is held across the network round-trip. That looks deliberate (no thundering herd) and the current call sites are effectively serial, so no change needed — but a one-line comment would stop it reading as an oversight. Note whichever caller wins the race also controls the exchange deadline via `req.Context()`, so an unrelated cancellation can fail a waiting caller.

internal/cyberark/conjur/conjur.go
```diff
@@ -0,0 +1,161 @@
...
+	// Request the base64-encoded access token — Conjur's canonical wire form
+	// for the token, and the encoding this client's own decoding below
+	// expects.
+	req.Header.Set("Accept-Encoding", "base64")
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("authn-jwt exchange transport error: %w", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
```
mladen-rusev-cyberark, line 87: Conjur returns a JSON error body with the actual reason; discarding it means every misconfiguration produces the same generic sentence and the operator has to go read Conjur's audit log — which is what #824's README ends up telling them to do. Suggest including a bounded prefix of the body.

Also, the body is never drained before `Close()`, so the connection can't be reused. Minor, but this is the path a retry loop hits.

roeezis:
All three addressed:
- TTL is now driven by the token's own `exp` claim when it can be read (renamed `identityFromToken` -> `claimsFromToken` to also extract it), falling back to a guessed TTL — now a `Client` field, not a package const — only when the token doesn't parse. Added `Invalidate()` for callers to force a fresh exchange on a 401, and a comment on the mutex-held-across-round-trip behavior noting it's deliberate. Also added tests: expired `exp` forces a re-exchange, a future `exp` overrides a short fallback TTL, and `Invalidate` forces a re-exchange.
- Exchange errors on a non-200 now include a bounded prefix of Conjur's own response body, and drain the remainder so the connection can be reused. Added a test asserting the body's message surfaces in the error.

internal/cyberark/conjur/conjur.go
```diff
@@ -0,0 +1,203 @@
...
+
+	"github.com/jetstack/preflight/internal/cyberark/jwtsource"
+)
+
+// defaultTokenTTL is the fallback cache lifetime used when a token's own
+// `exp` claim can't be read — Conjur access tokens default to an 8-minute
+// lifetime. It's a Client field, not a const, so tests can shrink it.
+const defaultTokenTTL = 8 * time.Minute
+
+// Client exchanges a JWT for a Conjur access token and authenticates requests with it.
```
mladen-rusev-cyberark, line 33: Suggested comment, not a blocker
```suggestion
// Client exchanges a JWT for a Conjur access token and authenticates requests with it.
//
// The token cache is currently never hit in production: selectAuthenticator
// builds a fresh Client for every upload, and keyfetch's long-lived Client is
// guarded by a 15-minute key cache that outlives the 8-minute token. That's a
// property of today's call sites, not a guarantee of this package. If a Client
// ever survives more than one authenticated request, three things need
// revisiting together: expiry needs a skew/latency margin, an already-past exp
// needs a floor so it doesn't re-exchange on every call, and Invalidate needs
// wiring wherever a 401 is observed.
```

mladen-rusev-cyberark (approved):

roeezis:
Took the suggested comment as-is — accurately scopes the current guarantee without overclaiming.

internal/cyberark/conjur/conjur.go
roeezis: TTL is now driven by the token's own `exp` claim when it can be read (renamed `identityFromToken` -> `claimsFromToken` to also extract it), falling back to a guessed TTL — now a `Client` field, not a package const — only when the token doesn't parse. Added `Invalidate()` for callers to force a fresh exchange on a 401, and a comment on the mutex-held-across-round-trip behavior noting it's deliberate. Also added tests: expired `exp` forces a re-exchange, a future `exp` overrides a short fallback TTL, and `Invalidate` forces a re-exchange.

internal/cyberark/conjur/conjur.go
roeezis: Covered by the same fix as the sibling comment — Invalidate() + exp-driven TTL address the re-authentication gap directly.

internal/cyberark/conjur/conjur.go
roeezis: Exchange errors on a non-200 now include a bounded prefix of Conjur's own response body, and drain the remainder so the connection can be reused. Added a test asserting the body's message surfaces in the error.

internal/cyberark/conjur/conjur.go
roeezis: Took the suggested comment as-is — accurately scopes the current guarantee without overclaiming.

PR #822, opened 2026-08-23
feat(agent): select Conjur JWT or legacy username/password authenticator

roeezis:
## Summary
Part 5 of the SMS/Conjur JWT authentication series (split out of #817). Stacked on the prior 4 PRs in this series — diff will shrink as they merge.

Adds `NewRequestAuthenticator`, choosing between the new Conjur JWT exchange and the legacy CyberArk Identity username/password login based on which config is present — Conjur JWT takes priority when both are set. Conjur JWT requires `service_id` and resolves its base URL from the `secrets_manager` service discovered in #820, not `identity_administration`; the two are different hosts.

Switches keyfetch's client over to the new authenticator selection instead of constructing a username/password identity client directly, so JWKS fetching supports both authentication methods too.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/cyberark/... ./internal/envelope/...`


mladen-rusev-cyberark (commented):
The selection logic is clean and `auth_select_test.go` pins the four cases well. But the keyfetch refactor introduces a regression that breaks the GA username/password path — details inline on `internal/envelope/keyfetch/client.go`.

Two smaller things:

- `ErrMissingEnvironmentVariables` now covers exactly one variable, so the name misleads. `ErrMissingSubdomain`?
- `auth_select_test.go`'s header comment says pointing a mock at one endpoint "proves which endpoint the code actually called". True for the username/password case, but the Conjur exchange is lazy, so `NewDatauploadClient` succeeds without ever dialling `SecretsManager`. It proves *selection*, which is the valuable part; to prove the endpoint you'd need to call the returned authenticator.

internal/envelope/keyfetch/client.go
```diff
@@ -117,7 +116,7 @@ func (c *Client) FetchKey(ctx context.Context) (PublicKey, error) {
 		return PublicKey{}, fmt.Errorf("failed to create request: %w", err)
 	}
 
-	_, err = c.identityClient.AuthenticateRequest(req)
+	_, err = c.authenticate(req)
```
mladen-rusev-cyberark, line 131: **Blocker.** This drops the per-`FetchKey` `LoginUsernamePassword` call, but `identity.Client.AuthenticateRequest` never refreshes — it serves whatever `LoginUsernamePassword` last cached, forever (`internal/cyberark/identity/authenticated_http_client.go`). The 15-minute re-login window lives *inside* the call that was deleted.

`keyfetch.NewClient` is called once per process at startup (`pkg/agent/run.go:301`) and the `Client` lives for the pod's lifetime, while `FetchKey` re-fetches every 15 min. So: log in once, then every JWKS fetch reuses that one bearer token. Once the identity token expires (~1h), every `FetchKey` 401s for the rest of the pod's life with no self-recovery — a restart is required. This is the GA path, on the default `sendSecretValues: true`.

The Conjur path is fine because `conjur.Client` re-exchanges on TTL — which is really the underlying problem: one `RequestAuthenticator` implementation refreshes and one doesn't, and the interface documents neither. Preferred fix is giving `identity.Client` the same self-refreshing `AuthenticateRequest`, so the contract holds for every future caller of this seam.

internal/envelope/keyfetch/client_test.go
```diff
@@ -10,30 +10,39 @@ import (
...
+	"github.com/jetstack/preflight/internal/cyberark/conjur"
 	"github.com/jetstack/preflight/internal/cyberark/servicediscovery"
 )
 
-// testClientSetup sets up a complete test environment with mock identity and discovery servers
-// and returns a configured client along with the test ClientConfig
+// testClientSetup sets up a complete test environment with mock conjur and discovery servers
+// and returns a configured client along with the test ClientConfig.
+// NOTE: this file imports venafi-connection-lib (via the keyfetch package itself) and therefore
+// cannot be compiled or run locally — the mock wiring is correct per the conjur mock pattern
```
mladen-rusev-cyberark: This isn't accurate — the package compiles and its tests run in a plain checkout:

```
$ go test ./internal/envelope/keyfetch/... -count=1
ok  	github.com/jetstack/preflight/internal/envelope/keyfetch	0.573s
```

The comment asserts the wiring is correct on the basis of an untrue premise, and invites the next reader to skip verifying this file. Please delete it.

Related, and more important: this helper was *converted* from the identity mock to the Conjur mock rather than extended, so keyfetch now has zero coverage of the username/password path — the one the blocker above breaks. That's why CI is green. Please parameterise `testClientSetup` over both auth methods, and add a test that advances past the token lifetime and asserts a second `FetchKey` still succeeds.

internal/cyberark/client.go
```diff
@@ -3,69 +3,151 @@ package cyberark
...
+//
+// Selection (backward compatible — the product is GA):
+//   - ServiceID set                  → Conjur JWT exchange (preferred).
+//   - else Username+Secret present   → legacy CyberArk Identity UP login.
+//   - neither                        → ErrNoAuthMethod.
+//
+// When both are configured, ServiceID wins (a migrating install can set the
+// service-id without first removing its old credentials) and a warning is logged.
+func selectAuthenticator(ctx context.Context, httpClient *http.Client, serviceMap *servicediscovery.Services, cfg ClientConfig) (identity.RequestAuthenticator, error) {
 	identityAPI := serviceMap.Identity.API
```
mladen-rusev-cyberark: This runs before the switch, so a tenant with `secrets_manager` but no `identity_administration` is rejected even on the pure-JWT path, which never uses that host. Unreachable today (`DiscoverServices` already errors on an empty identity API), but it's a coupling that works against the point of the PR. Suggest moving it into the `case hasUP:` arm next to `identity.New(...)`.

internal/cyberark/client.go
```diff
@@ -3,69 +3,151 @@ package cyberark
...
+	hasConjur := cfg.ServiceID != ""
+	hasUP := cfg.Username != "" && len(cfg.Secret) > 0
+
+	switch {
+	case hasConjur:
+		if hasUP {
+			klog.FromContext(ctx).Info("both Conjur service_id and ARK_USERNAME/ARK_SECRET are set; using the Conjur JWT exchange and ignoring the username/password credentials")
+		}
+		if cfg.JWTSource != "" && cfg.JWTSource != "file" {
+			return nil, fmt.Errorf("jwt_source %q not supported in POC (only 'file')", cfg.JWTSource)
```
mladen-rusev-cyberark: "POC" shouldn't reach an operator reading agent logs on a GA product. Suggest `jwt_source %q is not supported; supported values are "" and "file"`. The same label appears on the `ClientConfig` fields above.

This rule is also validated again with different wording in `pkg/agent/config.go` (#823). Worth hoisting `"file"` and `"conjur"` to exported consts here and having the config layer call a shared validator, so the two can't drift.

roeezis:
**Blocker fixed.** identity.Client.AuthenticateRequest is now self-refreshing, matching conjur.Client — it re-logs-in internally once the cached token passes tokenTTL (now a Client field, not a const), using a durable copy of the credentials captured at the last LoginUsernamePassword call. The caller's own password slice is still zeroed exactly as before; only the client's own internal copy persists, and that's not a materially different exposure than today's status quo (the config already holds it in memory for the process's lifetime). A refresh that fails falls back to the cached token rather than failing the request outright.

Also: moved the `identityAPI == ""` check into the hasUP branch — Conjur JWT never uses it.

Fixed in the same commit as the second finding below.

roeezis:
Fixed on both counts:
- Deleted the false comment.
- Added `testKeyfetchClientWithIdentityAuth` and `TestClient_FetchKey_UsernamePasswordAuth`, covering the legacy path end-to-end through `FetchKey` — including a case that forces the JWKS cache stale to prove a second fetch re-authenticates and still succeeds (this is exactly the seam the blocker above broke). Made keyfetch's own JWKS-cache TTL a Client field too, matching the pattern already used for conjur/identity, since the test needs to force it stale without a real 15-minute wait.

roeezis:
Fixed — hoisted `JWTSourceFile`/`DefaultAccount` consts and a shared `ValidateJWTSource` helper into `internal/cyberark`, called from both this file and `pkg/agent/config.go` so the two can't drift on wording again. Dropped the "POC" wording from the operator-facing message: now `jwt_source %q is not supported; supported values are "" and "file"`.

mladen-rusev-cyberark (commented):
Round 2 — **request changes.** Every round-1 finding is addressed, and the self-refresh implementation itself is correct: I checked the obvious hazard (`doAdvanceAuthentication` takes `*[]byte`) and confirmed it only reads the password, so the client's internal copy survives repeated refreshes. The new legacy-path tests are the right shape.

But the fix introduced a new blocker that's worse than the one it replaces — details inline on `internal/cyberark/client.go`. Legacy username/password auth now breaks on the **first** upload rather than an hour in.

Smaller point: the self-refresh is largely moot on the dataupload path, since `NewDatauploadClient` builds a brand-new `identity.Client` on every upload — the token cache never survives between cycles and the refresh never fires there. It only does real work for keyfetch's long-lived client. Worth knowing, since it means the retained-credential trade-off buys less than it appears, and it's another argument for hoisting authenticator construction out of the per-upload path.

internal/cyberark/client.go
```diff
@@ -3,69 +3,171 @@ package cyberark
...
+		conjurClient := conjur.New(httpClient, smsAPI, cfg.ServiceID, account, src)
+		return conjurClient.AuthenticateRequest, nil
+
+	case hasUP:
+		identityAPI := serviceMap.Identity.API
+		if identityAPI == "" {
+			return nil, errors.New("service discovery returned an empty identity API")
+		}
+		identityClient := identity.New(httpClient, identityAPI, cfg.Subdomain)
+		if err := identityClient.LoginUsernamePassword(ctx, cfg.Username, cfg.Secret); err != nil {
```
mladen-rusev-cyberark, line 138: **Blocker.** `LoginUsernamePassword` zeroes the caller's slice on return. `cfg` is passed by value, but `Secret` is a slice — the copy shares one backing array — so this wipes the caller's secret in place.

`NewCyberArk` builds one `cfg` and hands out a closure returning that same value forever (`pkg/client/client_cyberark.go:55`), and both consumers go through it:

- `loadEncryptor` → `keyfetch.NewClient` → `NewRequestAuthenticator` (`run.go:175`, startup)
- `PostDataReadingsWithOptions` → `NewDatauploadClient` (`run.go:272`, **every `config.period`**)

So on the chart default `sendSecretValues: true`, startup consumes the secret and the *first* upload already fails. With it `false`, the first upload works and every one after fails. Permanent either way — `ARK_SECRET` is only read at startup.

`hasUP` doesn't notice: `len(cfg.Secret) > 0` is still true for an all-zeros slice, so we take this branch and send zeros.

Reproduced through the real client path:

```
--- FAIL: TestProbe_UsernamePassword_SecondUpload
    Error: while initializing data upload client: CyberArk Identity username/password
           login failed: got a failure response from request to advance authentication:
           message="Authentication (login or challenge) has failed..."
    Messages: second upload
```

and at this level directly — `cfg.Secret after first use: "\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"`.

Verified fix, one line:

```go
if err := identityClient.LoginUsernamePassword(ctx, cfg.Username, append([]byte(nil), cfg.Secret...)); err != nil {
```

CI is green only because every test builds a fresh config while the agent reuses one — the closest test uploads exactly once. I've left the concrete regression-test request on #823, where that test lives.

internal/cyberark/client.go
```diff
@@ -3,69 +3,171 @@ package cyberark
...
+	// Conjur JWT exchange (preferred for new installs).
+	ServiceID   string // authn-jwt service id (POC: per-cluster, e.g. "dev-cluster")
+	Account     string // defaults to DefaultAccount
+	JWTSource   string // "" or JWTSourceFile (POC) | "spiffe" (deferred)
+	JWTFilePath string // default jwtsource.DefaultTokenPath
+
+	// Legacy CyberArk Identity username/password (backward compatibility).
+	// Sourced from ARK_USERNAME / ARK_SECRET. Used only when ServiceID is unset.
+	Username string
+	Secret   []byte
```
mladen-rusev-cyberark, line 58: This type change from `string` to `[]byte` is the root cause of the blocker below. Previously the call site was `[]byte(cfg.Secret)`, which allocated a fresh copy on every call — that's what made the zeroing safe. Now the same array is shared between the config and the callee that wipes it.

Beyond the one-line fix, I'd reconsider the contract: two owners of one array with an implicit "callee wipes it" rule is what caused this. Cleaner is for `LoginUsernamePassword` to copy internally and *stop* zeroing the caller's slice — the client now deliberately retains its own copy for the process lifetime anyway (`identity.go`'s `secret` field), so wiping the caller's buys nothing and only leaves this hazard for the next caller.

internal/cyberark/client.go
```diff
@@ -3,69 +3,171 @@ package cyberark
...
+		if smsAPI == "" {
+			return nil, errors.New("service discovery returned an empty secrets_manager API, which is required for the Conjur JWT exchange")
+		}
+		src := jwtsource.NewFileSource(cfg.JWTFilePath)
+		conjurClient := conjur.New(httpClient, smsAPI, cfg.ServiceID, account, src)
+		return conjurClient.AuthenticateRequest, nil
+
+	case hasUP:
+		identityAPI := serviceMap.Identity.API
+		if identityAPI == "" {
```
mladen-rusev-cyberark: The nit I flagged on #820 at `discovery.go:198`, raised here where it belongs.

This check can't fire. `DiscoverServices` returns an error when `identity_administration` is absent (`servicediscovery/discovery.go:198`), and every caller — `NewDatauploadClient`, `keyfetch.NewClient`, `keyfetch.FetchKey` — runs discovery before reaching here, so `serviceMap.Identity.API` is guaranteed non-empty by the time `selectAuthenticator` runs.

Moving it into `hasUP` was the right response to my round-1 point; the issue is what it now implies. It reads as "the Conjur path tolerates a missing Identity service," and that isn't a property the system has — a JWT-only tenant without `identity_administration` is rejected by discovery long before this branch. Suggest dropping it:

```go
case hasUP:
    identityClient := identity.New(httpClient, serviceMap.Identity.API, cfg.Subdomain)
```

The `smsAPI` check above stays — that one is genuinely reachable, and it's exactly the caller-validates pattern #820's new comment describes.

To be clear I'm not asking to relax discovery. Every service in `testdata/discovery_success.json.template` is present and active for the captured tenant, `identity_administration` included, so requiring it looks correct — and it may be the only suspended-tenant signal available, since `tenant_flags` carries nothing about suspension. The requirement is fine; the unreachable second check is what I'd remove.

Two optional follow-ons if you agree, both one-liners:
- a note above `discovery.go:198` saying `identityAPI` is required unconditionally *because* it's present for every healthy tenant, so callers may rely on it being non-empty — the block at `:202` explains why the other two aren't required but says nothing about this one
- `pkg/testutil/envtest.go:291` currently reads "Unused by the Conjur path, but service discovery requires it," which sounds like a workaround; with the contract written down it can just cite it

roeezis:
**Blocker confirmed and fixed** — reproduced your exact repro before fixing (same failure). Took the structural fix, not the one-liner: `LoginUsernamePassword` no longer zeroes the caller's slice at all — it already keeps its own durable copy (needed for the self-refresh from the earlier round), so wiping the caller's copy bought nothing and is exactly what caused this. Added a two-upload regression test on #823 (where the integration test for this path lives) — a single-upload test structurally can't catch this class of bug.

roeezis:
Agreed on the contract critique — fixed by removing the caller-side wipe entirely rather than just patching the call site (see the sibling comment on the blocker above). Two owners of one array with an implicit "callee wipes it" rule is exactly what broke; now there's one owner (the Client's internal copy) and the caller's slice is left alone.

roeezis:
Fixed — dropped the check, using serviceMap.Identity.API directly. Agreed on both optional follow-ons and added them: a note above `discovery.go:198` stating identityAPI is required unconditionally because it's present for every healthy tenant, and reworded envtest.go's comment to cite that contract instead of reading like a workaround.

internal/cyberark/identity/identity.go
```diff
@@ -1,188 +1,37 @@
...
 	tokenCachedMutex sync.Mutex
 	tokenCachedTime  time.Time
+
+	// username and secret are a durable copy of the credentials most recently
+	// passed to LoginUsernamePassword, retained so AuthenticateRequest can
+	// re-login on its own once the cached token ages out — the caller's own
+	// password slice is wiped after that call returns and can't be reused.
+	// This doesn't meaningfully change the secret's exposure: it already
+	// lives in memory for the process's lifetime in the config this client
+	// was constructed from.
```
mladen-rusev-cyberark:
```suggestion
    // username and secret are a durable copy of the credentials most recently
    // passed to LoginUsernamePassword, retained so AuthenticateRequest can
    // re-login on its own once the cached token ages out. Because this copy
    // exists, LoginUsernamePassword deliberately does NOT zero the caller's
    // slice — the caller may reuse it (NewCyberArk's ClientConfig is shared
    // across every upload). This doesn't meaningfully change the secret's
    // exposure: it already lives in memory for the process's lifetime in the
    // config this client was constructed from.
```

mladen-rusev-cyberark: I applied the suggestion by mistake, please revert or apply at your discretion but by using tabs instead of spaces to pass lint checks. @roeezis

roeezis: No worries — your suggestion was actually right (I had left that struct-field comment stale after removing the wipe, it still said the caller's slice gets zeroed). Took your wording as-is, fixed the tabs, and folded it into the existing commit rather than keeping a separate "apply suggestion" commit, to keep one commit per PR. Pushed.

internal/envelope/keyfetch/client.go
roeezis: **Blocker fixed.** identity.Client.AuthenticateRequest is now self-refreshing, matching conjur.Client — it re-logs-in internally once the cached token passes tokenTTL (now a Client field, not a const), using a durable copy of the credentials captured at the last LoginUsernamePassword call. The caller's password slice is still zeroed exactly as before; only the client's own internal copy persists, and that's not a materially different exposure than today's status quo (the config already holds it in memory for the process's lifetime). A refresh that fails falls back to the cached token rather than failing the request outright.

Also: moved the `identityAPI == ""` check into the hasUP branch — Conjur JWT never uses it.

internal/envelope/keyfetch/client_test.go
roeezis: Fixed on both counts:
- Deleted the false comment.
- Added `testKeyfetchClientWithIdentityAuth` and `TestClient_FetchKey_UsernamePasswordAuth`, covering the legacy path end-to-end through `FetchKey` — including a case that forces the JWKS cache stale to prove a second fetch re-authenticates and still succeeds. Made keyfetch's own JWKS-cache TTL a Client field too, matching the pattern already used for conjur/identity, since the test needs to force it stale without a real 15-minute wait.

internal/cyberark/client.go
roeezis: Fixed — hoisted `JWTSourceFile`/`DefaultAccount` consts and a shared `ValidateJWTSource` helper into `internal/cyberark`, called from both this file and `pkg/agent/config.go` so the two can't drift on wording again. Dropped the "POC" wording from the operator-facing message: now `jwt_source %q is not supported; supported values are "" and "file"`.

roeezis: Fixed — moved this into the `case hasUP:` arm, next to `identity.New(...)`. The Conjur path never reaches it now.

internal/cyberark/client.go
roeezis: Fixed — hoisted `JWTSourceFile`/`DefaultAccount` consts and a shared `ValidateJWTSource` helper into `internal/cyberark`, called from both this file and `pkg/agent/config.go` so the two can't drift on wording again. Dropped the "POC" wording from the operator-facing message: now `jwt_source %q is not supported; supported values are "" and "file"`.

internal/cyberark/client.go
roeezis: **Blocker confirmed and fixed** — reproduced your exact repro before fixing (same failure). Took the structural fix, not the one-liner: `LoginUsernamePassword` no longer zeroes the caller's slice at all — it already keeps its own durable copy (needed for the self-refresh from the earlier round), so wiping the caller's copy bought nothing and is exactly what caused this. Added a two-upload regression test on #823 (where the integration test for this path lives) — a single-upload test structurally can't catch this class of bug.

internal/cyberark/client.go
roeezis: Agreed on the contract critique — fixed by removing the caller-side wipe entirely rather than just patching the call site (see the blocker reply above). Two owners of one array with an implicit "callee wipes it" rule is exactly what broke; now there's one owner (the Client's internal copy) and the caller's slice is left alone.

internal/cyberark/client.go
roeezis: Fixed — dropped the check, using serviceMap.Identity.API directly. Agreed on both optional follow-ons and added them: a note above `discovery.go:198` stating identityAPI is required unconditionally because it's present for every healthy tenant, and reworded envtest.go's comment to cite that contract instead of reading like a workaround.

mladen-rusev-cyberark (approved):

PR #823, opened 2026-08-23
feat(agent): wire Conjur JWT config into the top-level CyberArk client

roeezis:
## Summary
Part 6 of the SMS/Conjur JWT authentication series (split out of #817). Stacked on the prior 5 PRs in this series — diff will shrink as they merge.

`NewCyberArk` and the agent config schema were still legacy-username/password-only at the top level, even though the underlying authenticator selection (#822) already supports Conjur JWT. Threads `service_id`/`account`/`jwt_source`/`jwt_file_path` through `cyberark.service_id` in the agent config down to `NewCyberArk`, so the config file is the single place an operator chooses which authentication method to use.

`config.go`'s call site and `client_cyberark.go`'s signature change together — splitting them across two PRs would leave one non-building at every commit in between, so they're one PR.

## Test plan
- [x] `go build ./...`
- [x] `go test ./pkg/agent/... ./pkg/client/... ./pkg/testutil/...`


pkg/agent/config.go
```diff
@@ -753,17 +772,15 @@ func ValidateAndCombineConfig(log logr.Logger, cfg Config, flags AgentCmdFlags)
...
 		case MachineHub:
 			clusterName = cfg.ClusterName
-			if clusterName == "" {
-				if arkUsername, found := os.LookupEnv("ARK_USERNAME"); found {
-					log.Info("Using ARK_USERNAME environment variable as cluster name", "clusterName", arkUsername)
-					clusterName = arkUsername
-				}
+			if clusterName == "" && cfg.ClusterID != "" {
+				log.Info("Using cluster_id as cluster_name for backwards compatibility", "clusterID", cfg.ClusterID)
+				clusterName = cfg.ClusterID
```
mladen-rusev-cyberark, line 778: This changes the MachineHub cluster-name fallback from `ARK_USERNAME` to `cluster_id`, and promotes `cluster_id` from explicitly-ignored to load-bearing. The change is motivated — `ARK_USERNAME` doesn't exist on the JWT path — but an existing install relying on the old fallback that also sets `cluster_id` will start reporting under a **different cluster name** after upgrade, possibly fragmenting its history in Discovery & Context. Installs with neither now report an empty cluster name.

Three asks:

1. Mention it in the commit message and PR description — as written both are entirely about Conjur config threading, so a reader wouldn't know cluster naming moved.
2. Confirm with whoever owns the MachineHub data model that a cluster-name change on upgrade is acceptable, and whether a migration note is needed.
3. The log line says "for backwards compatibility", but using `cluster_id` is the *new* behaviour — worth rewording.

Given it's independently reviewable and independently riskier than the rest of this PR, it may deserve its own.

pkg/testutil/envtest.go
```diff
@@ -262,31 +262,42 @@ func FakeTPP(t testing.TB) (*httptest.Server, *x509.Certificate) {
...
-	identityAPI, _ := identity.MockIdentityServer(t)
+	// Write a temp JWT file so jwtsource.NewFileSource can read a token.
+	jwtFile, err := os.CreateTemp(t.TempDir(), "jwt-*")
+	require.NoError(t, err)
+	_, err = jwtFile.WriteString("fake-service-account-jwt")
+	require.NoError(t, err)
+	require.NoError(t, jwtFile.Close())
+
+	// "success-token" matches the dataupload mock's expected bearer token.
+	conjurSrv, _ := conjur.MockConjurExchangeServer(t, "success-token")
```
mladen-rusev-cyberark, line 285: Same pattern as #822's keyfetch tests: the shared helper is converted to Conjur rather than extended to cover both. Every integration test built on `FakeCyberArk` now exercises only the new path, and combined with the keyfetch change the GA username/password path has no end-to-end coverage left anywhere in the tree. Between the two, that's why the #822 regression passes CI.

Please make this parameterisable over auth method and keep at least one caller on the legacy path.

Smaller note on the same function: the signature changed to return `(httpClient *http.Client, jwtFilePath string)`, where the second value is only meaningful for one auth method. That's a hint an options struct or two constructors would age better.

pkg/agent/config.go
```diff
@@ -773,6 +790,24 @@ func ValidateAndCombineConfig(log logr.Logger, cfg Config, flags AgentCmdFlags)
 		res.ClaimableCerts = cfg.ClaimableCerts
 	}
 
+	// Validation of `cyberark.*` (MachineHub mode only).
+	if res.OutputMode == MachineHub {
+		ark := CyberArkConfig{}
+		if cfg.CyberArk != nil {
+			ark = *cfg.CyberArk
```
mladen-rusev-cyberark, line 814: The reasoning here is sound as far as it goes, but "fails closed" happens at first upload, not at startup — up to a full `config.period` later. It *does* fail at startup today, but only because `sendSecretValues: true` is the chart default and `loadEncryptor` builds an authenticator eagerly; set it to `false` and a fully unconfigured agent reports healthy for an hour.

This function can see the env — the old code read `ARK_USERNAME` via `os.LookupEnv` right here — so the check is cheap:

```go
if ark.ServiceID == "" && os.Getenv("ARK_USERNAME") == "" {
	errs = multierror.Append(errs, fmt.Errorf("MachineHub mode requires either cyberark.service_id or ARK_USERNAME/ARK_SECRET"))
}
```

Misconfiguration should surface at `helm upgrade`, not an hour into the pod's life.

mladen-rusev-cyberark (commented):
Config plumbing is straightforward, and the reasoning for combining the call-site and signature change in one commit is right. Two things ride along that I'd separate, or at least call out — inline.

Minor points:

- `NewCyberArk` now takes four adjacent same-typed string params (`pkg/client/client_cyberark.go:45`) — classic transposition hazard, and the compiler can't help. `CyberArkConfig` already exists and is already being destructured at the call site; passing the struct would be one self-documenting parameter, and adding a fifth field later wouldn't be a signature change.
- `pkg/agent/config.go:783`: after the block above it, `clusterName == ""` already implies `cfg.ClusterID == ""`, so the second conjunct is dead.
- `res.CyberArk = ark` only runs in MachineHub mode, so a `cyberark:` block in a Venafi Cloud config is silently ignored. Everywhere else in this function ignored fields get a log line (see the `organization_id` handling just above).

roeezis:
Fixed — added `FakeCyberArkUsernamePassword` and a matching integration test (`TestCyberArkClient_PostDataReadingsWithOptions_UsernamePasswordMockAPI`) that leaves serviceID empty, so the legacy path is what's actually under test rather than silently exercising Conjur.

On the signature note: kept `FakeCyberArk`'s existing shape as-is (minimal blast radius) and gave the new function its own return shape (`httpClient, username, password`) instead — two constructors, each specific to its auth method, rather than forcing one shared shape.

roeezis:
Fixed — added the check you suggested (same code, `os.Getenv("ARK_USERNAME")`), and reworded the comment to describe what's actually checked and why (deferring to selectAuthenticator alone would leave a misconfigured agent reporting healthy for up to a full config.period). Updated the two existing tests that asserted "valid at config time" to set ARK_USERNAME (preserving what they were actually testing — the username/password fallback still works without service_id), and added a negative test for neither being set.

roeezis:
Confirmed accurate, and I'm not resolving this one — it's a product decision, not something I should silently pick:

- Old behavior: `cluster_id` was explicitly ignored in MachineHub mode; the fallback was `ARK_USERNAME`.
- New behavior (this PR): falls back to `cluster_id`, which is now load-bearing.

An existing install with `ARK_USERNAME` set AND a `cluster_id` in config (previously just ignored) will report under a different cluster name post-upgrade, and one with neither set will report an empty name where it previously had `ARK_USERNAME`'s value.

I've reworded the log line (point 3 — it said "for backwards compatibility" describing what is actually new behavior). Points 1 and 2 need a human call: @rzisholz — is a cluster-name change on upgrade acceptable here, and does whoever owns the MachineHub data model need to weigh in or get a migration note? I don't have the context to decide this myself, and I don't want to guess and silently ship a behavior change to production cluster identity.

roeezis:
Resolved with a compromise from @rzisholz: cluster_id still takes priority over ARK_USERNAME when both are set (needed since ARK_USERNAME doesn't exist on the Conjur JWT path), but ARK_USERNAME is restored as a further fallback rather than dropped — order is now cluster_name > cluster_id > ARK_USERNAME > empty.

This means:
- An existing install with ARK_USERNAME set and cluster_id also set (previously ignored) still gets a different cluster_name than before — that part of the original concern stands, cluster_id genuinely needs priority for the new auth path.
- An existing install with ARK_USERNAME set and neither cluster_name nor cluster_id configured is no longer affected — it keeps reporting under its ARK_USERNAME-derived name instead of going empty.

Added three tests pinning this exact order (`cluster_name falls back to cluster_id when set, even if ARK_USERNAME is also set` / `... to ARK_USERNAME when cluster_id is unset` / `... is empty when all three are unset`).

roeezis:
The fallback now depends on which auth method is actually active, mirroring selectAuthenticator's own precedence, instead of one universal order:

- Conjur JWT (service_id set): falls back to cluster_id — ARK_USERNAME doesn't exist on this path.
- legacy username/password (service_id unset): the *original* fallback this mode has always had, completely unchanged — ARK_USERNAME only.

This fully closes the original concern: an existing username/password install now sees zero behavior change, in any configuration. cluster_id is only ever consulted on the new Conjur path, which no existing install could be on.

roeezis:
Follow-up on the "confirm with whoever owns the MachineHub data model" ask — checked directly against discoverycontext-file-ingestor-service rather than guessing.

`cluster_id` is the sole identity key end-to-end: the entity record's ID (`OriginStore.id`) is generated from `[data_source_type, provider_id, provider_type]` — `provider_id` traces to `cluster_id`. `cluster_name` is excluded from that generation entirely, and every one of its ~15 usages across the codebase (SNS payload attribute, DynamoDB stored column, K8s label mirror) is display-only — none participate in any DynamoDB key, generated ID, or lookup/equality check.

So there's no data/identity fragmentation risk from this fallback-order change, in either direction — worst case is a cosmetic display-label flicker for one upgrade, since `cluster_id` (unchanged) is what actually determines whether two snapshots correlate to the same cluster. Downgrading this from "needs data-model-owner sign-off" to "informational" — closing.

pkg/agent/config.go
```diff
@@ -753,7 +773,20 @@ func ValidateAndCombineConfig(log logr.Logger, cfg Config, flags AgentCmdFlags)
 			clusterID = cfg.ClusterID
 		case MachineHub:
 			clusterName = cfg.ClusterName
-			if clusterName == "" {
+			// The fallback order depends on which auth method is active,
+			// mirroring cyberark.selectAuthenticator's own precedence
```
mladen-rusev-cyberark: You escalated this rather than deciding it silently, which was right. But I think the decision can be avoided: the new requirement and the old behaviour aren't actually in conflict.

The direction here is good — `cluster_id` → `cluster_name` mirrors what `VenafiCloudKeypair`/`VenafiConnection` already do at `config.go:757-761`, so MachineHub becomes consistent rather than special. And I checked that reusing config's `cluster_id` as a *name* doesn't collide with the snapshot's own `cluster_id`: that's populated from the `ark/discovery` reading (`client_cyberark.go:155`), the real cluster UID, not from config.

What's lost is the `ARK_USERNAME` fallback. Restoring it *below* both config fields keeps every existing install's name while leaving the new behaviour intact:

```go
case MachineHub:
	clusterName = cfg.ClusterName
	if clusterName == "" && cfg.ClusterID != "" {
		log.Info("Using cluster_id as cluster_name", "clusterID", cfg.ClusterID)
		clusterName = cfg.ClusterID
	}
	if clusterName == "" {
		// Legacy fallback: pre-Conjur installs set neither cluster_name nor
		// cluster_id — the chart only ever emitted cluster_name — and were
		// named after ARK_USERNAME. Kept below both config fields so explicit
		// configuration always wins. Naming a cluster after a login identity
		// is an accident of the original design and is worth retiring once
		// installs set cluster_name; the variable itself stays, since it is
		// still the credential for username/password auth.
		if arkUsername := os.Getenv("ARK_USERNAME"); arkUsername != "" {
			log.Info("Using ARK_USERNAME as cluster name because neither cluster_name nor cluster_id is set; prefer setting cluster_name explicitly", "clusterName", arkUsername)
			clusterName = arkUsername
		}
	}
	if cfg.OrganizationID != "" {
		log.Info(fmt.Sprintf(`Ignoring the organization_id field in the config file. This field is not needed in %s mode.`, res.OutputMode))
	}
	if clusterName == "" {
		log.Info("cluster_name is not set in MachineHub mode; cluster name will be empty")
	}
```

Precedence becomes `cluster_name` → `cluster_id` → `ARK_USERNAME` → empty. Explicit config always beats the env var, so this doesn't introduce the inversion that would otherwise make the fallback objectionable — there's no other place in the agent where an env var overrides a config field, and this doesn't become one.

What makes it work is that the conflict is nearly unreachable: `deploy/charts/disco-agent/templates/configmap.yaml` emits only `cluster_name` (line 10) — the chart has no `clusterId` value at all. So a chart-installed MachineHub agent can't have `cluster_id` set, and `cluster_id` can't displace `ARK_USERNAME` for those installs.

| install | `cluster_name` | `cluster_id` | `ARK_USERNAME` | today | this PR | proposed |
|---|---|---|---|---|---|---|
| chart, named | set | — | set | *name* | *name* | *name* |
| **chart, bare** | — | — | set | ARK_USERNAME | **""** | ARK_USERNAME |
| **migrating to JWT** | — | — | set | ARK_USERNAME | **""** | ARK_USERNAME |
| JWT-only | — | set | — | "" | cluster_id | cluster_id |
| hand-written, both | — | set | set | ARK_USERNAME | cluster_id | cluster_id |

The migrating row matters: `selectAuthenticator` deliberately supports `service_id` and `ARK_USERNAME` together so an operator can adopt JWT without first deleting credentials. As written, that operator's cluster renames itself the moment they add `service_id`. Changing how you authenticate shouldn't rename your cluster — nothing validates `cluster_name` (empty included), so a renamed or blank cluster lands in MachineHub as a new entry with its history orphaned.

The last row is the one real change left, and I'd accept it: `cluster_id` is explicit config and should win, and today that config already logs `Ignoring the cluster_id field`. If you want zero silent renames, add a line in the `cluster_id` branch noting it superseded an `ARK_USERNAME` that was also set.

This also collapses the dead conjunct at line 783 — `clusterName == "" && cfg.ClusterID == ""` can just be `clusterName == ""`, since the block above already assigns `clusterName` whenever `ClusterID` is non-empty.

If you take this, the only thing left for a product call is whether to retire the naming fallback on a timeline — which needn't gate this merge.

pkg/agent/config.go
```diff
@@ -773,6 +806,27 @@ func ValidateAndCombineConfig(log logr.Logger, cfg Config, flags AgentCmdFlags)
 		res.ClaimableCerts = cfg.ClaimableCerts
```
mladen-rusev-cyberark, line 807: This misses `ARK_SECRET`. `selectAuthenticator` requires both (`hasUP := cfg.Username != "" && len(cfg.Secret) > 0`, `internal/cyberark/client.go:107`), so `ARK_USERNAME` set with `ARK_SECRET` unset passes validation here, the agent starts healthy, and then fails at first upload with `ErrNoAuthMethod` — exactly the deferred failure this check exists to prevent.

```go
if ark.ServiceID == "" && (os.Getenv("ARK_USERNAME") == "" || os.Getenv("ARK_SECRET") == "") {
```

To be clear where this came from: you added the check exactly as I wrote it in my round-1 review, so the bug is mine. I caught it in round 2 but that review was never posted — this is the first you're seeing of it.

For contrast, `ARK_SUBDOMAIN` is already enforced properly at startup: `LoadClientConfigFromEnvironment` rejects it empty and `NewCyberArk` is called from `ValidateAndCombineConfig` (`config.go:1029`) with the error appended to `errs`. The fail-closed story is one `||` away from complete — worth landing before #824, whose README keeps a "fails closed at startup" sentence that is false for this one case until then.

mladen-rusev-cyberark (commented):
Round 3 — **approve**, with one code gap and one behaviour proposal inline.

The two-upload regression test is the right test and it genuinely works: I reintroduced the `defer` wipe in `username_password.go` and it fails on the second upload with the original Identity error; reverting makes it pass. Its doc comment explains the `[]byte`-via-closure sharing clearly and says outright that a single upload structurally can't catch it. The `envtest.go` reword cites the discovery contract properly now.

Two things inline:

1. **`config.go:806` still misses `ARK_SECRET`.** This one is mine — you added the check exactly as I wrote it in round 1, and I got it wrong. My round-2 review flagging it was never posted, so you've had no chance to fix it. Details inline.
2. **The MachineHub cluster-name change** — I think this can preserve existing behaviour *and* meet the new requirement, without the product decision you escalated. Concrete proposal inline.

Minor, all fine to leave:

- `res.CyberArk = ark` only runs in MachineHub mode, so a `cyberark:` block in a Venafi Cloud config is silently ignored — unlike `organization_id` just above, which logs.
- `NewCyberArk`'s four adjacent string params: passing a struct would mean `pkg/client` importing `pkg/agent`, so leaving it is defensible — worth a comment saying so.

roeezis:
Took this exactly as proposed — verified both supporting claims first (`client_cyberark.go:155` confirms `ClusterID` on the snapshot comes from the `ark/discovery` reading, not config; `configmap.yaml` confirms the chart never emits `cluster_id`, so the conflict really is unreachable for chart installs). Precedence is now `cluster_name > cluster_id > ARK_USERNAME > empty`, the dead conjunct is gone, and added the migrating-install case as its own test (`service_id` set, no `cluster_id`, `ARK_USERNAME` still present → name unchanged) since that's the exact regression you caught.

roeezis:
Confirmed and fixed — added the `|| os.Getenv("ARK_SECRET") == ""` you gave, plus a test for ARK_USERNAME-without-ARK_SECRET, and had to fix two existing tests that were (accidentally) relying on the old gap.

pkg/agent/config.go
roeezis: Confirmed accurate, and I'm not resolving this one myself — it's a product decision: old behavior fell back to `ARK_USERNAME`, new behavior falls back to `cluster_id` (previously explicitly ignored). Tagged @rzisholz directly on this thread rather than picking a side silently. Superseded by the redesign below once they weighed in — see the later reply in this thread.

pkg/testutil/envtest.go
roeezis: Fixed — added `FakeCyberArkUsernamePassword` and a matching integration test (`TestCyberArkClient_PostDataReadingsWithOptions_UsernamePasswordMockAPI`) that leaves serviceID empty, so the legacy path is what's actually under test rather than silently exercising Conjur.

On the signature note: kept `FakeCyberArk`'s existing shape as-is (minimal blast radius) and gave the new function its own return shape (`httpClient, username, password`) instead — two constructors, each specific to its auth method, rather than forcing one shared shape.

pkg/agent/config.go
roeezis: Fixed — added the check you suggested (same code, `os.Getenv("ARK_USERNAME")`), and reworded the comment to describe what's actually checked and why (deferring to selectAuthenticator alone would leave a misconfigured agent reporting healthy for up to a full config.period). Updated the two existing tests that asserted "valid at config time" to set ARK_USERNAME (preserving what they were actually testing — the username/password fallback still works without service_id), and added a negative test for neither being set.

(Turned out this was missing ARK_SECRET too — see your later comment on this thread, fixed there.)

pkg/agent/config.go
roeezis: Took this exactly as proposed — verified both supporting claims first (`client_cyberark.go:155` confirms `ClusterID` on the snapshot comes from the `ark/discovery` reading, not config; `configmap.yaml` confirms the chart never emits `cluster_id`, so the conflict really is unreachable for chart installs). Precedence is now `cluster_name > cluster_id > ARK_USERNAME > empty`, the dead conjunct is gone, and added the migrating-install case as its own test (`service_id` set, no `cluster_id`, `ARK_USERNAME` still present → name unchanged) since that's the exact regression you caught.

Follow-up verification against discoverycontext-file-ingestor-service: `cluster_id` (not `cluster_name`) is the sole identity key end-to-end — the entity ID is generated from `[data_source_type, provider_id, provider_type]`, with `cluster_name` excluded from that hash and every one of its usages purely display/pass-through. So this whole fallback-order question was lower-stakes than either of us initially framed it — no data/identity fragmentation risk either way, worst case is a cosmetic label for one upgrade.

pkg/agent/config.go
roeezis: Confirmed and fixed — added the `|| os.Getenv("ARK_SECRET") == ""` you gave, plus a test for ARK_USERNAME-without-ARK_SECRET, and had to fix two existing tests that were (accidentally) relying on the old gap.

mladen-rusev-cyberark (approved):

PR #824, opened 2026-08-23
feat(agent): add Helm chart support for Conjur JWT authentication

roeezis:
## Summary
Part 7 (final) of the SMS/Conjur JWT authentication series (split out of #817). Stacked on the prior 6 PRs in this series — diff will shrink to just this PR's own change once they merge.

Exposes `config.cyberark`'s new Conjur JWT fields (`serviceId`, `account`, `jwtSource`) in `values.yaml`/`values.schema.json`, and documents per-cluster Conjur onboarding as a prerequisite in the chart README — the tenant administrator's own Conjur Cloud credentials, no admin credential needed at deploy time.

## Test plan
- [x] `go build ./...`
- [x] `helm lint deploy/charts/disco-agent/`

### DIFF
--- a/deploy/charts/disco-agent/README.md
+++ b/deploy/charts/disco-agent/README.md
@@ -16,24 +16,38 @@ kubectl create ns "$NAMESPACE" || true
 
 ### Add credentials to a Secret
 
-You will require tenant details and credentials for the CyberArk Identity Security Platform.
-Put them in the following environment variables:
+The agent supports **two authentication methods**, selected automatically by
+config:
+
+| Set this | Method used |
+|---|---|
+| `config.cyberark.serviceId` (Conjur authn-jwt service-id) | **Conjur JWT exchange** — exchanges a projected ServiceAccount token for a short-lived Conjur access token. No stored password. Preferred for new installs. |
+| `ARK_USERNAME` + `ARK_SECRET` in the Secret (and no `serviceId`) | **Legacy CyberArk Identity username/password** — backward compatible with existing GA installs. |
+
+If **both** are set, the Conjur `serviceId` wins (so a migrating install can add
+the service-id before removing its old credentials) and a warning is logged. If
+**neither** is set, the agent fails closed at startup.
+
+The only credential always required in the Kubernetes Secret is the CyberArk
+tenant subdomain (`ARK_SUBDOMAIN`).
 
 ```sh
-export ARK_SUBDOMAIN=      # your CyberArk tenant subdomain e.g. tlskp-test
-export ARK_USERNAME=       # your CyberArk username
-export ARK_SECRET=         # your CyberArk password
-# OPTIONAL: the URL for the CyberArk Discovery API if not using the production environment
+export ARK_SUBDOMAIN=      # your CyberArk tenant subdomain, e.g. tlskp-test
+# OPTIONAL: Discovery API URL for non-production environments
 export ARK_DISCOVERY_API=https://platform-discovery.integration-cyberark.cloud/
 ```
 
-Create a Secret containing the tenant details and credentials:
+Create the Secret:
 
 ```sh
+# Production (no ARK_DISCOVERY_API override needed):
 kubectl create secret generic agent-credentials \
         --namespace "$NAMESPACE" \
-        --from-literal=ARK_USERNAME=$ARK_USERNAME \
-        --from-literal=ARK_SECRET=$ARK_SECRET \
+        --from-literal=ARK_SUBDOMAIN=$ARK_SUBDOMAIN
+
+# Non-production, targeting a non-default Discovery API:
+kubectl create secret generic agent-credentials \
+        --namespace "$NAMESPACE" \
         --from-literal=ARK_SUBDOMAIN=$ARK_SUBDOMAIN \
         --from-literal=ARK_DISCOVERY_API=$ARK_DISCOVERY_API
 ```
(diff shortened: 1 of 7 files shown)


deploy/charts/disco-agent/templates/deployment.yaml
```diff
@@ -116,6 +122,11 @@ spec:
             - name: config
               mountPath: "/etc/disco-agent"
               readOnly: true
+            {{- if or (not .Values.config.cyberark.jwtSource) (eq .Values.config.cyberark.jwtSource "file") }}
+            - name: conjur-token
+              mountPath: /var/run/secrets/tokens
```
mladen-rusev-cyberark, line 127: **Blocker.** This mount path is hardcoded while `configmap.yaml` passes the user's `jwtFilePath` straight through, so a non-default value produces a deployment that can't work. Confirmed by rendering:

```
$ helm template t deploy/charts/disco-agent --set acceptTerms=true \
    --set config.cyberark.serviceId=x --set config.cyberark.jwtFilePath=/custom/tok/jwt

      jwt_file_path: "/custom/tok/jwt"        # agent reads here
              mountPath: /var/run/secrets/tokens
                  path: jwt                   # token is actually here
```

The agent fails with `jwt source file "/custom/tok/jwt" not found or unreadable` and never authenticates. Both `values.yaml:230` and `values.schema.json` actively assert this works ("The chart auto-mounts a projected volume at this path").

Either make it true — `mountPath: {{ dir .Values.config.cyberark.jwtFilePath }}` and `path: {{ base .Values.config.cyberark.jwtFilePath }}` — or drop the knob entirely and let the agent's `jwtsource.DefaultTokenPath` apply. I'd take the second: the chart owns the mount, and it's one fewer way to misconfigure.

mladen-rusev-cyberark (commented):
Chart renders cleanly and the env-var changes are right — `ARK_SUBDOMAIN` stays required, legacy keys become optional, so JWT-only installs need no username/password. `expirationSeconds: 600` is the API-server minimum and works out fine given the kubelet rotates the file in place and `jwtsource` re-reads per exchange, so the 8-minute Conjur cache never outlives the projected token.

Findings are all in the volume wiring and the docs. One I couldn't leave inline:

**Generated docs are stale.** The `## Values` section of `README.md` and `values.schema.json` are both generated by `helm-tool` (`make/_shared/helm/helm.mk:98,106`), and the new `config.cyberark.*` values are absent from the `<!-- AUTO-GENERATED -->` block (lines 146–520). So either the verify target fails in CI, or the schema was hand-edited into a shape `helm-tool` wouldn't produce. Please run `make generate` and commit the result — the existing `values.yaml` entries carry `# +docs:property` markers and the new block has none, which may be why generation skipped it.

Also: `jwtSource` has no `enum` in the schema, so a typo (`fille`) renders fine, deploys fine, and CrashLoops. Worth `"enum": ["", "file"]`.

deploy/charts/disco-agent/templates/deployment.yaml
```diff
@@ -116,6 +122,11 @@ spec:
             - name: config
               mountPath: "/etc/disco-agent"
               readOnly: true
+            {{- if or (not .Values.config.cyberark.jwtSource) (eq .Values.config.cyberark.jwtSource "file") }}
```
mladen-rusev-cyberark: This gates on `jwtSource` only, which defaults to `file` — so **every** install gets an `audience: conjur` projected token volume, including pure `ARK_USERNAME`/`ARK_SECRET` ones with `serviceId: ""`. Harmless in most clusters, but it's unnecessary surface, and admission policies that constrain projected-token audiences will reject or flag it on upgrade for users who opted into nothing.

Suggest gating on the thing that actually selects the feature:

```
{{- if and .Values.config.cyberark.serviceId (or (not .Values.config.cyberark.jwtSource) (eq .Values.config.cyberark.jwtSource "file")) }}
```

Worth defining as a named template, since the condition appears twice and must stay in sync with the volume below.

Related: `audience: conjur` is hardcoded, but the new troubleshooting table tells operators to confirm it matches the authenticator's configured audience — with no way to change it from values. Either expose `config.cyberark.audience` (defaulting to `conjur`), or state in the README that it's fixed and the Conjur side must match.

deploy/charts/disco-agent/README.md
```diff
@@ -16,26 +16,36 @@ kubectl create ns "$NAMESPACE" || true
...
+config:
+
+| Set this | Method used |
+|---|---|
+| `config.cyberark.serviceId` (Conjur authn-jwt service-id) | **Conjur JWT exchange** — exchanges a projected ServiceAccount token for a short-lived Conjur access token. No stored password. Preferred for new installs. |
+| `ARK_USERNAME` + `ARK_SECRET` in the Secret (and no `serviceId`) | **Legacy CyberArk Identity username/password** — backward compatible with existing GA installs. |
+
+If **both** are set, the Conjur `serviceId` wins (so a migrating install can add
+the service-id before removing its old credentials) and a warning is logged. If
+**neither** is set, the agent fails closed at startup.
```
mladen-rusev-cyberark, line 29: This isn't guaranteed. Per #822/#823, `LoadClientConfigFromEnvironment` only requires `ARK_SUBDOMAIN` and there's no auth-method check in `ValidateAndCombineConfig`; `ErrNoAuthMethod` surfaces inside `selectAuthenticator` at first upload. It happens at startup today only because `sendSecretValues: true` is the chart default and `loadEncryptor` builds an authenticator eagerly — set it to `false` and the sentence becomes false.

Either fix the agent (see my comment on `pkg/agent/config.go` in #823) and keep this, or reword now. Docs asserting a safety property the code doesn't guarantee are worse than no sentence.

Separately, a few lines below: the optional `ARK_DISCOVERY_API` went from a simple `--from-literal` to a commented-out `kubectl patch --type=json` with nested shell quoting and a manual `base64`. That's fragile (unquoted `$ARK_DISCOVERY_API`, and `base64` line-wrapping differs between GNU and BSD) and harder to follow. Since the secret key is already `optional: true` in the deployment, the plain two-variant `kubectl create secret` form would read better.

roeezis:
**Blocker fixed** — took your recommended second option: dropped the `jwtFilePath` knob entirely (removed from values.yaml/values.schema.json/configmap.yaml) rather than trying to template mountPath from it. The agent's `jwtsource.DefaultTokenPath` already matches the chart's fixed mount path, so there's nothing left to misconfigure. Re-verified with your exact repro command — `jwt_file_path` no longer appears in the rendered config at all, and the mount is unconditionally the working path.

roeezis:
Fixed — gated both the volume and the mount on a new `disco-agent.conjurJWTEnabled` named template (`serviceId` set AND jwtSource unset-or-"file"), used in both places so they can't drift out of sync again. A pure username/password install now renders zero references to the conjur-token volume.

On audience: documented in the README's troubleshooting table that it's fixed at `conjur`, not a values.yaml setting, rather than adding a config knob nothing currently needs.

roeezis:
Both addressed:
- The startup-failure claim is now actually true independent of `sendSecretValues` — fixed the underlying agent-side gap in #823 (config-validation now requires service_id or ARK_USERNAME at `ValidateAndCombineConfig` time, not deferred to first upload), so kept the README sentence as-is rather than hedging it.
- Replaced the `kubectl patch --type=json` snippet with two plain `kubectl create secret` variants (with/without ARK_DISCOVERY_API) — simpler, no shell-quoting/base64-portability issues.

deploy/charts/disco-agent/values.schema.json
```diff
@@ -149,6 +152,37 @@
...
+        }
+      },
+      "type": "object"
+    },
+    "helm-values.config.cyberark.account": {
+      "default": "conjur",
+      "description": "The Conjur account name. For CyberArk-hosted tenants this is always \"conjur\".",
+      "type": "string"
+    },
+    "helm-values.config.cyberark.jwtSource": {
```
mladen-rusev-cyberark, line 175: Nit: no `enum` here. `"enum": ["", "file"]` would turn a typo into a `helm install`-time error instead of a CrashLoopBackOff.

Lower priority than it was in round 1, since `ValidateJWTSource` now rejects a bad value at startup with a clear message — so this is genuinely just about where the operator finds out. Fine to leave.

deploy/charts/disco-agent/README.md
```diff
@@ -49,23 +63,67 @@ metadata:
...
+— the agent builds the authenticate URL as
+`<base>/authn-jwt/<serviceId>/<account>/authenticate`, so a path here would
+double the `conjur/authn-jwt` prefix. The remaining defaults are correct for
+CyberArk-hosted tenants:
+
+| Value | Default | Description |
+|---|---|---|
+| `config.cyberark.serviceId` | `""` | Conjur authn-jwt service ID (required). Example: `disco-agent` |
+| `config.cyberark.account` | `conjur` | Conjur account name. Always `conjur` for CyberArk-hosted tenants. |
+| `config.cyberark.jwtSource` | `file` | Token source. `file` = projected SA-token volume (default). `spiffe` deferred. |
```
mladen-rusev-cyberark, line 91: Nit, about the block below rather than this line: README lines 151–525 are wrapped in `<!-- AUTO-GENERATED -->` / `<!-- /AUTO-GENERATED -->`, but nothing regenerates them — the generate targets only touch `venafi-kubernetes-agent`. Inside that block `config.period`, `config.clusterName` and `config.sendSecretValues` are documented; every `config.cyberark.*` value is absent. They *are* documented — by hand, here at lines 89–91.

Either add them inside the block or drop the markers, so nobody assumes a generator owns the file and regenerates it away.

Correction to my round-1 review: I claimed these docs were stale and asked for `make generate`. That was wrong — I ran the target and it doesn't touch this chart. This narrower point is what survives.

roeezis:
Added `"enum": ["", "file"]` — a typo now fails at `helm install` instead of a CrashLoopBackOff.

roeezis:
Dropped the markers — nothing regenerates this chart, and they implied otherwise. The cyberark.* fields stay hand-documented above the section, where they already were.

deploy/charts/disco-agent/templates/deployment.yaml
roeezis: **Blocker fixed** — took your recommended second option: dropped the `jwtFilePath` knob entirely (removed from values.yaml/values.schema.json/configmap.yaml) rather than trying to template mountPath from it. The agent's `jwtsource.DefaultTokenPath` already matches the chart's fixed mount path, so there's nothing left to misconfigure. Re-verified with your exact repro command — `jwt_file_path` no longer appears in the rendered config at all, and the mount is unconditionally the working path.

deploy/charts/disco-agent/templates/deployment.yaml
roeezis: Fixed — gated both the volume and the mount on a new `disco-agent.conjurJWTEnabled` named template (`serviceId` set AND jwtSource unset-or-"file"), used in both places so they can't drift out of sync again. A pure username/password install now renders zero references to the conjur-token volume.

On audience: documented in the README's troubleshooting table that it's fixed at `conjur`, not a values.yaml setting, rather than adding a config knob nothing currently needs.

deploy/charts/disco-agent/README.md
roeezis: Both addressed:
- The startup-failure claim is now actually true independent of `sendSecretValues` — fixed the underlying agent-side gap in #823 (config-validation now requires service_id or ARK_USERNAME at `ValidateAndCombineConfig` time, not deferred to first upload), so kept the README sentence as-is rather than hedging it.
- Replaced the `kubectl patch --type=json` snippet with two plain `kubectl create secret` variants (with/without ARK_DISCOVERY_API) — simpler, no shell-quoting/base64-portability issues.

deploy/charts/disco-agent/values.schema.json
roeezis: Added `"enum": ["", "file"]` — a typo now fails at `helm install` instead of a CrashLoopBackOff.

deploy/charts/disco-agent/README.md
roeezis: Dropped the markers — nothing regenerates this chart, and they implied otherwise. The cyberark.* fields stay hand-documented above the section, where they already were.

mladen-rusev-cyberark (approved):

PR #827, opened 2026-09-02
CP-25864: fall back cluster_name to cyberark.service_id under Conjur JWT

roeezis:
## Summary
- MachineHub `cluster_name` fallback chain (`cluster_name` -> `cluster_id` -> `ARK_USERNAME` -> empty) always resolves empty for Conjur JWT installs: the disco-agent chart never sets `cluster_id`, and `ARK_USERNAME` doesn't exist under JWT auth.
- Adds `cyberark.service_id` (the Conjur authn-jwt service ID, always set by the CyberArk onboarding wizard) as a final fallback before empty.
- Combined with the existing config-validation gate (requires `service_id` or `ARK_USERNAME`+`ARK_SECRET`), an empty `ClusterName` is now unreachable for any MachineHub config that passes validation.

## Test plan
- [x] `go test ./pkg/agent/...` (excluding the pre-existing envtest-gated `Test_ValidateAndCombineConfig_VenafiConnection`, unrelated to this change)
- [x] Updated 2 existing tests whose expected behavior changed, added 1 new fallback test, removed 1 now-unreachable "empty" case with explanatory comment

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -793,6 +793,14 @@ func ValidateAndCombineConfig(log logr.Logger, cfg Config, flags AgentCmdFlags) 
 					clusterName = arkUsername
 				}
 			}
+			if clusterName == "" && cfg.CyberArk != nil && cfg.CyberArk.ServiceID != "" {
+				// Conjur JWT installs have no ARK_USERNAME to fall back to.
+				// cyberark.service_id (the Conjur authn-jwt service ID,
+				// typically the cluster UUID) is always set for these
+				// installs, so it's a better last resort than an empty name.
+				log.Info("Using cyberark.service_id as cluster name because cluster_name, cluster_id, and ARK_USERNAME are all unset; prefer setting cluster_name explicitly", "clusterName", cfg.CyberArk.ServiceID)
+				clusterName = cfg.CyberArk.ServiceID
+			}
 			if cfg.OrganizationID != "" {
 				log.Info(fmt.Sprintf(`Ignoring the organization_id field in the config file. This field is not needed in %s mode.`, res.OutputMode))
 			}
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -796,9 +796,11 @@ func ValidateAndCombineConfig(log logr.Logger, cfg Config, flags AgentCmdFlags) 
 			if clusterName == "" && cfg.CyberArk != nil && cfg.CyberArk.ServiceID != "" {
 				// Conjur JWT installs have no ARK_USERNAME to fall back to.
 				// cyberark.service_id (the Conjur authn-jwt service ID,
-				// typically the cluster UUID) is always set for these
-				// installs, so it's a better last resort than an empty name.
-				log.Info("Using cyberark.service_id as cluster name because cluster_name, cluster_id, and ARK_USERNAME are all unset; prefer setting cluster_name explicitly", "clusterName", cfg.CyberArk.ServiceID)
+				// expected to be cluster-unique, but that uniqueness is an
+				// onboarding convention, not something this agent verifies)
+				// is always set for these installs, so it's a better last
+				// resort than an empty name.
+				log.Info("cluster_name and cluster_id are unset and ARK_USERNAME is not set; falling back to cyberark.service_id as the cluster name — set cluster_name explicitly for a human-readable name, and ensure service_id is unique per cluster if you rely on this fallback", "clusterName", cfg.CyberArk.ServiceID)
 				clusterName = cfg.CyberArk.ServiceID
 			}
 			if cfg.OrganizationID != "" {
--- a/pkg/agent/config_test.go
+++ b/pkg/agent/config_test.go
@@ -1388,14 +1388,17 @@ func TestConfig_CyberArk_Validation(t *testing.T) {
 	})
 
 	// cluster_name fallback order: cluster_name > cluster_id > ARK_USERNAME >
-	// empty. Explicit config (cluster_name, cluster_id) always wins over the
-	// env var; ARK_USERNAME is a legacy fallback kept only for pre-Conjur
-	// installs that set neither config field — the chart never emits
-	// cluster_id, so a chart-installed agent can't have set it, and
-	// ARK_USERNAME still applies. This also covers a migrating install
+	// cyberark.service_id > empty. Explicit config (cluster_name, cluster_id)
+	// always wins over the env var; ARK_USERNAME is a legacy fallback kept
+	// only for pre-Conjur installs that set neither config field — the chart
+	// never emits cluster_id, so a chart-installed agent can't have set it,
+	// and ARK_USERNAME still applies. This also covers a migrating install
 	// (service_id set to test Conjur alongside still-present ARK_USERNAME,
 	// no cluster_id yet): its reported name doesn't change just because
-	// service_id was added.
+	// service_id was added. service_id is checked last, below ARK_USERNAME,
+	// so it only kicks in for pure Conjur-JWT installs that have nothing
+	// else set — the onboarding wizard always sets it, so it's a real
+	// fallback even if the generated Helm command omits cluster_name.
 	t.Run("cluster_name falls back to cluster_id when set, even if ARK_USERNAME is also set", func(t *testing.T) {
 		setEnv(t)
 		t.Setenv("ARK_USERNAME", "svc-agent@tenant")
@@ -1434,7 +1437,7 @@ func TestConfig_CyberArk_Validation(t *testing.T) {
 		assert.Equal(t, "svc-agent@tenant", got.ClusterName)
 	})
 
-	t.Run("cluster_name is empty when cluster_name, cluster_id, and ARK_USERNAME are all unset", func(t *testing.T) {
+	t.Run("cluster_name falls back to cyberark.service_id when cluster_name, cluster_id, and ARK_USERNAME are all unset", func(t *testing.T) {
 		setEnv(t)
 		got, _, err := ValidateAndCombineConfig(discardLogs(),
 			withConfig(testutil.Undent(`
@@ -1443,8 +1446,16 @@ func TestConfig_CyberArk_Validation(t *testing.T) {
 			`)),
 			withCmdLineFlags("--period", "1m", "--machine-hub"))
 		require.NoError(t, err)
-		assert.Equal(t, "", got.ClusterName)
+		assert.Equal(t, "dev-cluster", got.ClusterName)
 	})
+
+	// The config-validation gate (either service_id or ARK_USERNAME+ARK_SECRET
+	// must be set) combined with the fallback chain means a valid MachineHub
+	// config can no longer produce an empty ClusterName: whichever of the two
+	// auth methods satisfies validation also satisfies a cluster_name
+	// fallback. There is deliberately no test asserting an empty ClusterName
+	// here — every combination that reaches this point has a non-empty
+	// fallback available.
 
 	t.Run("jwt_source spiffe is rejected", func(t *testing.T) {
 		setEnv(t)
--- a/pkg/agent/config_test.go
+++ b/pkg/agent/config_test.go
@@ -1449,13 +1449,23 @@ func TestConfig_CyberArk_Validation(t *testing.T) {
 		assert.Equal(t, "dev-cluster", got.ClusterName)
 	})
 
-	// The config-validation gate (either service_id or ARK_USERNAME+ARK_SECRET
-	// must be set) combined with the fallback chain means a valid MachineHub
-	// config can no longer produce an empty ClusterName: whichever of the two
-	// auth methods satisfies validation also satisfies a cluster_name
-	// fallback. There is deliberately no test asserting an empty ClusterName
-	// here — every combination that reaches this point has a non-empty
-	// fallback available.
+	// The only way ClusterName ends up empty is if config-validation already
+	// failed: whichever of the two auth methods (service_id, or
+	// ARK_USERNAME+ARK_SECRET) satisfies the gate at config.go's cyberark
+	// validation block also satisfies a cluster_name fallback above. Assert
+	// the gate directly, rather than only in prose, so a future change that
+	// makes either auth method optional without updating the fallback chain
+	// fails this test instead of silently reintroducing an empty
+	// cluster_name for a config that validates successfully.
+	t.Run("empty ClusterName only occurs when config validation also fails", func(t *testing.T) {
+		setEnv(t)
+		got, _, err := ValidateAndCombineConfig(discardLogs(),
+			withConfig(""),
+			withCmdLineFlags("--period", "1m", "--machine-hub"))
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "MachineHub mode requires either cyberark.service_id or ARK_USERNAME/ARK_SECRET")
+		assert.Equal(t, "", got.ClusterName)
+	})
 
 	t.Run("jwt_source spiffe is rejected", func(t *testing.T) {
 		setEnv(t)
--- a/deploy/charts/disco-agent/README.md
+++ b/deploy/charts/disco-agent/README.md
@@ -424,8 +424,7 @@ Enable sending of Secret values to CyberArk in addition to metadata. Metadata is
 > ""
 > ```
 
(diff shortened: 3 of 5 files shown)


pkg/agent/config.go
```diff
@@ -793,6 +793,14 @@ func ValidateAndCombineConfig(log logr.Logger, cfg Config, flags AgentCmdFlags)
 					clusterName = arkUsername
 				}
 			}
+			if clusterName == "" && cfg.CyberArk != nil && cfg.CyberArk.ServiceID != "" {
+				// Conjur JWT installs have no ARK_USERNAME to fall back to.
+				// cyberark.service_id (the Conjur authn-jwt service ID,
+				// typically the cluster UUID) is always set for these
```
wallrj-cyberark: `typically the cluster UUID` does not match what this repository tells users about `service_id`.

- `deploy/charts/disco-agent/values.yaml` describes `config.cyberark.serviceId` as "The Conjur authn-jwt authenticator service ID configured for **this tenant**".
- The README section explaining it is titled "Per-tenant Conjur onboarding".
- The worked `helm upgrade` example in that README passes the literal `--set config.cyberark.serviceId=disco-agent`.
- `internal/cyberark/client.go:50` hedges the other way: `// authn-jwt service id (POC: per-cluster, e.g. "dev-cluster")`.

The genuine per-cluster UUID is already collected elsewhere: `Snapshot.ClusterID`, sourced from the `kube-system` namespace UID in `pkg/datagatherer/k8sdiscovery/discovery.go`.

Concrete failure: a tenant onboards one authn-jwt authenticator and installs the agent on ten clusters with the wizard's Helm command, which omits `cluster_name`. All ten upload `cluster_name: "disco-agent"`. In Discovery and Context they become ten identically named clusters. That is arguably worse than the current empty name, because a name that looks deliberately chosen gives the operator no signal that a fallback fired, whereas an unnamed cluster does.

If the wizard genuinely mints a per-cluster service ID, please say so here and cite it, because the chart documentation currently says the opposite and one of the two needs correcting.

pkg/agent/config.go
```diff
@@ -793,6 +793,14 @@ func ValidateAndCombineConfig(log logr.Logger, cfg Config, flags AgentCmdFlags)
...
+				// installs, so it's a better last resort than an empty name.
+				log.Info("Using cyberark.service_id as cluster name because cluster_name, cluster_id, and ARK_USERNAME are all unset; prefer setting cluster_name explicitly", "clusterName", cfg.CyberArk.ServiceID)
+				clusterName = cfg.CyberArk.ServiceID
```
wallrj-cyberark, line 804: Worth confirming against the backend before merging: this changes what *existing* Conjur JWT installs send, not only what new ones send.

`dataupload.Snapshot.ClusterName` has no `omitempty` (`internal/cyberark/dataupload/dataupload.go:62`), so today every existing Conjur JWT install PUTs `"cluster_name": ""` once per period. After this change the same installs start PUTting `"cluster_name": "<service_id>"` on their first upload after upgrade. If the Discovery and Context API upserts the display name keyed on `cluster_id`, a name an operator set in the UI is silently overwritten, and the only way to opt out is to start setting `config.clusterName` in the chart.

The other clients deliberately avoid this: `pkg/client/client_venafi_cloud.go:195` and `pkg/client/client_ngts.go:241` both add the `name` query parameter only `if opts.ClusterName != ""`, so empty means "leave the name alone". The CyberArk snapshot path has no equivalent guard, which is what turns a config-time default into an unconditional rename.

pkg/agent/config_test.go
```diff
@@ -1443,9 +1446,17 @@ func TestConfig_CyberArk_Validation(t *testing.T) {
 			`)),
 			withCmdLineFlags("--period", "1m", "--machine-hub"))
 		require.NoError(t, err)
-		assert.Equal(t, "", got.ClusterName)
+		assert.Equal(t, "dev-cluster", got.ClusterName)
 	})
 
+	// The config-validation gate (either service_id or ARK_USERNAME+ARK_SECRET
```
wallrj-cyberark: Replacing the test with a comment removes the only guard on this invariant.

The "empty is now unreachable" property is not local to this fallback chain. It depends on the auth gate thirty lines away at `pkg/agent/config.go:830` (`service_id == "" && (ARK_USERNAME == "" || ARK_SECRET == "")` is an error). I agree it holds today. But if someone later adds a third auth method, or makes `service_id` optional for a new JWT source, an empty `ClusterName` becomes reachable again and nothing in the suite fails.

The PR also keeps the code that handles the case it declares impossible: `pkg/agent/config.go:807-809` still logs "cluster name will be empty". Please pick one. If it is reachable, keep a test for it. If it is not, delete the branch as part of this change. Dead code plus a prose assertion that it is dead is the worst of the two options.

The cheap fix is to keep a subtest but assert the thing that actually makes empty unreachable: a MachineHub config with neither `service_id` nor `ARK_USERNAME`/`ARK_SECRET` must fail validation.

pkg/agent/config.go
```diff
@@ -793,6 +793,14 @@ func ValidateAndCombineConfig(log logr.Logger, cfg Config, flags AgentCmdFlags)
 					clusterName = arkUsername
 				}
 			}
+			if clusterName == "" && cfg.CyberArk != nil && cfg.CyberArk.ServiceID != "" {
```
wallrj-cyberark, line 796: The chart documentation still describes the old behaviour and is not updated here.

`deploy/charts/disco-agent/values.yaml` says of `config.clusterName`: "If empty (the default), the service account name will be used instead." The same sentence is copied into the generated `deploy/charts/disco-agent/README.md:404` and `values.schema.json:152`.

For a Conjur JWT install there is no service account name. `ARK_USERNAME` does not exist, which is the premise of this PR, and the reported name will now be the `serviceId`. Since the motivation is that the wizard's generated Helm command omits `cluster_name`, the chart values are exactly where an operator will look to work out what their cluster ends up being called.

Please update the `clusterName` comment in `values.yaml` and re-run `make generate-helm-docs generate-helm-schema` so the README and the schema follow.

pkg/agent/config.go
roeezis: Checked this against the actual onboarding wizard implementation (internal tooling, not part of this repo) rather than relying on the chart's README example. Confirmed: each cluster registration mints its own value for this field — driven by a per-cluster identifier the operator supplies during onboarding, used to scope a dedicated authenticator and policy branch to that specific cluster. Two clusters onboarded under the same tenant get different values, not a shared one. So the wizard flow this fallback exists for is genuinely per-cluster; the chart README's single literal example is for the generic manual-deploy path and doesn't reflect what the wizard generates — a separate, pre-existing doc gap I can address in a follow-up PR if useful.

Caveat: that identifier is operator-typed free text, validated only for UUID shape, not auto-derived from the cluster itself — so per-cluster-uniqueness holds structurally but depends on the operator entering distinct values per cluster. Not a defect in this repo, just worth knowing.

pkg/agent/config.go
roeezis: Checked this against the actual backend persistence path (internal, not part of this repo). Three findings:

1. Confirmed: every periodic upload overwrites the stored name unconditionally — no first-write-wins, no separate display-name field today.
2. The per-cluster identity used to decide *which* record gets updated is derived only from the cluster ID, never from the name — the name is display-only and never participates in matching/deduplication. So the worst case from this overwrite behavior is a wrong label, never a merged, split, or duplicated inventory record.
3. There's no UI feature today that lets an operator set a name for this field to clobber — only read paths exist. So the specific failure (operator renames in UI, agent silently reverts it) is hypothetical right now, not a live regression this PR introduces.

What this PR changes for already-deployed Conjur JWT installs: they go from uploading an empty name every cycle (no fallback existed) to uploading a non-empty fallback value every cycle — a net improvement given there's nothing today to overwrite. Your point stands as a real architectural gap for whenever a rename UI ships, affecting every client, not just this one — worth a tracking item on the backend side, but I don't think it should block this PR. Let me know if you'd still like this held pending sign-off.

pkg/agent/config_test.go
roeezis: Fair — replaced the comment with a real test in 777a67c: `empty ClusterName only occurs when config validation also fails` asserts the validation error directly, not just prose. Left the "cluster name will be empty" log in place as a defensive branch for if a future auth method reopens this path — it's no longer dead-code-plus-unverified-claim, since there's now a test tied to the actual gate that would need updating (and would fail) if that gate ever changes.

pkg/agent/config.go
roeezis: Fixed in 777a67c: updated the `clusterName` comment in values.yaml to describe the actual fallback chain (ARK_USERNAME for legacy, cyberark.serviceId for Conjur JWT, no service-account-name fallback), and ran `make ark-generate` to regenerate README.md and values.schema.json from it.

pkg/agent/config.go
wallrj-cyberark: Thank you for checking the wizard implementation — that answers the question I could not answer from this repository, and I withdraw the main finding. On your evidence the values are minted per cluster, so the "ten clusters all report `disco-agent`" scenario I described does not arise from the wizard flow this fallback is written for.

Two residual points, neither blocking.

The first follows from your own caveat. If the identifier is operator-typed and validated only for UUID shape, uniqueness is a convention the operator follows rather than something the system guarantees. That is fine, but it does mean the failure mode survives in weaker form: an operator who reuses one value across clusters now gets several identically named clusters instead of several unnamed ones. A one-line note in the log message at `pkg/agent/config.go:801` would cost nothing and would tell an operator where the name came from when they see a surprising one.

The second is that a UUID-shaped value is a poor display name. `values.yaml` opens by calling `clusterName` "A human readable name for the cluster", and this PR makes the fallback for Conjur JWT installs a value validated for UUID shape. Whether a UUID beats an empty name in the Discovery and Context list is your call and I would not block on it, but it is worth being deliberate about, because it is the outcome for every wizard install that does not set `cluster_name`.

On the chart documentation being a pre-existing gap: agreed that it predates you, but I would rather it were not deferred. `values.schema.json:182` still describes `serviceId` as "configured for this tenant" and still gives `disco-agent` as the worked example, and this PR adds a sentence two entries above it telling the operator that `serviceId` is what their cluster gets named. A reader now meets both statements in the same generated file. Since you are already regenerating these three files in this PR, correcting that one sentence in `values.yaml` and re-running the generator seems cheaper than carrying a follow-up.

pkg/agent/config.go
wallrj-cyberark: Withdrawing this one as a blocker. Your point 3 is the part I could not see from this repository, and it settles it: if there is no write path for the name today, there is nothing for the agent to clobber, so this PR does not introduce a live regression. Your point 2 also rules out the outcome I was most concerned about — if matching is keyed on the cluster ID and never on the name, the worst case really is a wrong label rather than a split or merged inventory record.

I agree the remaining issue is a backend one and is not this PR's to fix. I would only note that it is latent rather than absent: the moment a rename path ships, every client that sends a non-empty name unconditionally starts overwriting it, and this PR moves the Conjur JWT installs from the harmless side of that line to the affected side. That is an argument for the tracking item you suggested, not for holding this.

No sign-off needed from me on this thread.

pkg/agent/config_test.go
wallrj-cyberark: Resolved, thank you. I checked out 777a67c and ran the suite: all fourteen subtests of `TestConfig_CyberArk_Validation` pass, including the new one. It asserts the validation error itself rather than the absence of a value, which is the property that actually makes an empty name unreachable, so a future auth method that loosens that gate will fail this test.

I accept the argument for keeping the log branch now that the gate is covered.

pkg/agent/config.go
wallrj-cyberark: Resolved, thank you. `values.yaml`, `README.md:404` and `values.schema.json:152` now carry the same corrected sentence, so the regeneration was run properly and the three files agree.

One observation, not a request. The new text describes a three-step chain, whereas `pkg/agent/config.go` implements four: `cluster_name`, then `cluster_id`, then `ARK_USERNAME`, then `serviceId`. The omitted `cluster_id` step sits ahead of `ARK_USERNAME`. I checked before raising it and the chart exposes no `clusterId` key at all, so that step is unreachable for anyone installing via Helm and the text is accurate for its audience. Mentioning it only so the omission is a known one rather than an oversight, in case the chart ever grows that key.

pkg/agent/config.go
roeezis: Both addressed in 62a26ad:

- Log note: the fallback log line now names which fields were unset and explicitly calls out that service_id uniqueness is a convention, not something the agent verifies, so a surprising name points an operator at the right cause.
- values.yaml's serviceId comment still said "configured for this tenant" two lines above where this PR's clusterName comment says it becomes the cluster's display name — same generated file asserting both. Fixed to per-cluster, matching README's existing hand-written prose, and dropped the non-unique worked example from the comment. Regenerated README.md/values.schema.json.

On the UUID-as-display-name point: agreed it's not a great display name, but that's true of any fallback value here (ARK_USERNAME wasn't chosen for readability either) — I'll leave it as your observation rather than build something to prettify it, since you said you wouldn't block on it.

pkg/agent/config.go
roeezis: Agreed on all points — latent rather than absent is the right framing, and moving Conjur JWT installs to the affected side of that line is a fair thing to flag even without blocking here. I'll raise the tracking item on the backend side separately.

roeezis:
All threads resolved, checks green. Let me know if there's anything else you'd like changed before approving — happy to address further.

pkg/agent/config.go
wallrj-cyberark: The reference section reads correctly now, but the contradiction has moved rather than gone: it is still in `README.md`, about 320 lines above the text you fixed.

`README.md:100` is still headed "Per-tenant Conjur onboarding" and still says "the target tenant must be onboarded", and the deploy snippet at `README.md:125` still passes `--set config.cyberark.serviceId=disco-agent`. That is the same shared literal you have just removed from the generated comment for being non-unique. So the file now tells an operator at line 125 to copy a fixed value, and at line 424 that the value must be cluster-unique or their cluster names will collide.

The concrete case is an operator onboarding three clusters from that snippet. They paste the same command three times, which is what a copy-paste block invites, and all three upload `cluster_name: disco-agent` — the exact outcome the new reference text warns about, reached by following the documented instructions.

This is the last thing outstanding from me. The heading, the "tenant must be onboarded" sentence and the literal in the snippet are the three spots; everything else in this PR looks right. Verified at 62a26ad: `go build ./...` is clean and `TestConfig_CyberArk_Validation` passes all fourteen subtests, with only the pre-existing `KUBEBUILDER_ASSETS` envtests failing locally.

wallrj-cyberark (approved):
Approving. The fallback logic is correctly placed, the behaviour is now covered by a test that asserts the validation gate rather than the absence of a value, and the two questions I could not settle from this repository have been answered with evidence from the systems concerned. Thank you for chasing both down rather than waving them through.

Verified locally at 62a26ad: `go build ./...` is clean and all fourteen subtests of `TestConfig_CyberArk_Validation` pass. The only failures are the pre-existing `KUBEBUILDER_ASSETS` envtests, which fail identically on master and are called out in the description.

One documentation point remains, and I am explicitly not blocking on it. `README.md:100` is still headed "Per-tenant Conjur onboarding" and the deploy snippet at `README.md:125` still passes `--set config.cyberark.serviceId=disco-agent`, which is the same shared literal removed from the generated comment in 62a26ad for not being unique. As it stands the file asks an operator to paste a fixed value in one place and warns that the value must be cluster-unique in another. Worth correcting the heading, the "tenant must be onboarded" sentence and the literal in the snippet, either before merge or as a follow-up — your call.

The backend overwrite behaviour discussed in the second thread is a pre-existing gap affecting every client, not something this change introduces, and is better tracked separately as you suggested.

roeezis:
Fixed in c876edf — retitled to "Per-cluster Conjur onboarding", corrected the onboarding sentence, and replaced the shared `disco-agent` literal in the worked example with a placeholder `$SERVICE_ID` variable, with a note that it's cluster-specific.

pkg/agent/config.go
wallrj-cyberark: Confirmed fixed in c876edfc — all three spots, thank you. The heading is now "Per-cluster", the opening sentence says the cluster rather than the tenant and states outright that the service ID is not shared, and the snippet takes `$SERVICE_ID` with a comment telling the reader not to reuse it. An operator following the deploy block on three clusters can no longer end up with three identically named ones.

I checked the edit is durable rather than about to be regenerated away: `generate-helm-docs` injects only between the `<!-- AUTO-GENERATED -->` markers at lines 153 and 552, and this prose sits above line 153, so the next `make ark-generate` will leave it alone. The generated reference block is still consistent with `values.yaml`.

Nothing further from me. My approval stands and covers this commit — it is documentation only, with no change to any Go file since the version I verified.

PR #828, opened 2026-09-03
test: document why the migration-precedence test already proves Conjur wins over username/password

roeezis:
## Summary

- `TestCyberArkClient_PostDataReadingsWithOptions_MockAPI` already sets both a non-empty Conjur `serviceID` and `ARK_USERNAME`/`ARK_SECRET` together, and passes -- proving `selectAuthenticator`'s documented precedence (Conjur wins when both are set) actually holds. That property wasn't documented as deliberate, so it read as incidental copy-paste. Adds a doc comment making it explicit.
- Mechanism: `FakeCyberArk`'s `Identity.API` is an unreachable `.invalid` host, so if the username/password path were mistakenly chosen instead of Conjur, this test would fail with a connection error rather than silently passing against the wrong path.
- A few adjacent legacy-compatibility properties needed no new code: removing the old credentials after cutover is already covered by the existing Conjur-only test; routing both credential shapes correctly across tenants is server-side (already covered on the authorizer side); and an un-onboarded agent failing closed (no silent fallback) is structurally guaranteed by `selectAuthenticator`'s switch/case having no fallback branch once Conjur is chosen.

## Test plan

- [x] `go test ./pkg/client/... -run TestCyberArkClient_PostDataReadingsWithOptions_MockAPI -v` passes; log line confirms the precedence decision at runtime ("both Conjur service_id and ARK_USERNAME/ARK_SECRET are set; using the Conjur JWT exchange and ignoring the username/password credentials")
- [x] `go test ./pkg/client/... ./internal/cyberark/...` -- no new failures (2 pre-existing unrelated failures confirmed present on unmodified master too: `KUBEBUILDER_ASSETS` env requirement, and a known Go jsonv2 error-string drift)

### DIFF
--- a/pkg/client/client_cyberark_test.go
+++ b/pkg/client/client_cyberark_test.go
@@ -28,18 +28,10 @@ import (
 // The environment variables are chosen to match those expected by the mock
 // server.
 //
-// ARK_USERNAME/ARK_SECRET are set here deliberately, alongside a non-empty
-// serviceID (test plan Track F, case F2 -- migrating install, add serviceId
-// while old creds are still present). This is not incidental: it is the
-// integration-level proof that selectAuthenticator's Conjur-wins precedence
-// actually holds through the full NewCyberArk/PostDataReadingsWithOptions
-// path, not just in the selectAuthenticator unit tests (auth_select_test.go).
-// It works because FakeCyberArk's Identity.API is an unreachable
-// `.invalid` host: if the username/password path were mistakenly chosen
-// instead of Conjur, the login attempt would fail with a connection error
-// and this test would fail. Do not remove these two lines as "unnecessary" --
-// doing so would silently delete this test's only coverage of migration
-// precedence.
+// ARK_USERNAME/ARK_SECRET are set here on purpose alongside a non-empty
+// serviceID: selectAuthenticator ignores them once serviceID is set, and
+// FakeCyberArk's Identity.API is an unreachable `.invalid` host -- a
+// mistaken fallback to username/password would fail this test.
 func TestCyberArkClient_PostDataReadingsWithOptions_MockAPI(t *testing.T) {
 	t.Setenv("ARK_SUBDOMAIN", servicediscovery.MockDiscoverySubdomain)
 	t.Setenv("ARK_USERNAME", "test@example.com")
@@ -60,12 +52,9 @@ func TestCyberArkClient_PostDataReadingsWithOptions_MockAPI(t *testing.T) {
 }
 
 // TestCyberArkClient_PostDataReadingsWithOptions_UsernamePasswordMockAPI is
-// the legacy-auth-path counterpart to the Conjur test above. Without it, the
-// only integration-level test of NewCyberArk's username/password path is one
-// that sets ARK_USERNAME/ARK_SECRET but then also passes a non-empty
-// serviceID — which selectAuthenticator prioritises, so it silently exercises
-// Conjur regardless of those env vars. This test leaves serviceID empty so
-// the username/password path is what's actually under test.
+// the legacy-auth-path counterpart to the Conjur test above (see its comment
+// for why serviceID must stay empty here for this test to actually exercise
+// the username/password path).
 func TestCyberArkClient_PostDataReadingsWithOptions_UsernamePasswordMockAPI(t *testing.T) {
 	t.Setenv("ARK_SUBDOMAIN", servicediscovery.MockDiscoverySubdomain)
 	httpClient, username, password := testutil.FakeCyberArkUsernamePassword(t)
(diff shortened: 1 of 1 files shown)


pkg/client/client_cyberark_test.go
```diff
@@ -27,6 +27,19 @@ import (
 // dataupload code works with the mock CyberArk APIs.
 // The environment variables are chosen to match those expected by the mock
 // server.
+//
+// ARK_USERNAME/ARK_SECRET are set here deliberately, alongside a non-empty
+// serviceID (test plan Track F, case F2 -- migrating install, add serviceId
+// while old creds are still present). This is not incidental: it is the
```
wallrj-cyberark: These two comments now describe the same setup in opposite terms, about thirty lines apart.

Yours says the env vars are "not incidental" and are "the integration-level proof" of precedence. The comment on `TestCyberArkClient_PostDataReadingsWithOptions_UsernamePasswordMockAPI` just below says the same test "sets ARK_USERNAME/ARK_SECRET but then also passes a non-empty serviceID — which selectAuthenticator prioritises, so it silently exercises Conjur regardless of those env vars".

A reader who meets both takes away that the env vars are deliberate proof and that they are silently ignored. Both are true from different angles, but "not incidental" and "silently ... regardless" read as a straight contradiction.

Worth reconciling the two while you are here. Something like "the env vars are ignored by design, and this test is what proves they are ignored" covers both readings in one sentence, and the sibling comment can then point at yours rather than restating it.

pkg/client/client_cyberark_test.go
```diff
@@ -27,6 +27,19 @@ import (
 // dataupload code works with the mock CyberArk APIs.
 // The environment variables are chosen to match those expected by the mock
 // server.
+//
+// ARK_USERNAME/ARK_SECRET are set here deliberately, alongside a non-empty
+// serviceID (test plan Track F, case F2 -- migrating install, add serviceId
```
wallrj-cyberark: "test plan Track F, case F2" will not mean anything to most people reading this file.

This repository is public, and the test plan is not in it. A future reader cannot look up Track F to find out what F2 was meant to establish, so the reference cannot be checked or maintained. It also cannot tell them whether the case still exists.

The good news is the comment already explains the scenario in words — "migrating install, add serviceId while old creds are still present" is the part that carries the meaning. I would keep that and drop the Track F label, or move it to the PR description where the context lives.

While you are trimming: thirteen lines is long for a comment guarding two lines of setup. The mechanism paragraph earns its place because the `.invalid` trick is genuinely surprising. The rest could compress to two or three sentences without losing anything.

wallrj-cyberark (commented):
Approving in substance — I verified the mechanism this comment claims, and it holds. Two wording points below, neither blocking.

I checked the load-bearing claim empirically rather than by reading. I copied the test, set `serviceID` to empty so `selectAuthenticator` picks username/password, and ran it against the same `FakeCyberArk` fixture. It fails exactly as you describe.

<details>
<summary>Counterfactual run: forcing the username/password path against FakeCyberArk does fail</summary>

```
COUNTERFACTUAL RESULT err=while initializing data upload client:
CyberArk Identity username/password login failed:
failed to perform HTTP request to start authentication:
Post "https://identity.example.invalid/Security/StartAuthentication":
dial tcp: lookup identity.example.invalid on 127.0.0.53:53: no such host
```

`MockDiscoveryServer` returns `server.Client()`, a real client with a real transport, so nothing rewrites the host and `.invalid` reaches the resolver. The guarantee is genuine, not an artefact of the mock.

I also confirmed the tests pass and the build is clean at 07acbb6b, and I removed my scratch test afterwards.

</details>

One suggestion on top of the two comments. The strongest argument for "do not remove these two lines" is not in the comment: 3e753e2 records that CI sets `ARK_USERNAME` and `ARK_SECRET` at job level from real secrets, for the `ARK_LIVE_TEST` live test. So without these `t.Setenv` calls the test would read ambient real credentials in CI and nothing locally, and would behave differently in the two places. That is a concrete past incident in this repo, and citing it would carry more weight than the instruction not to remove the lines.

roeezis:
Fixed in 0162d71 — reconciled the two comments (one framing: ignored by design, this test proves it) and trimmed 13 lines to 4, dropping the internal test-plan reference.

wallrj-cyberark (approved):
Both points addressed in 0162d71, thank you.

The two comments now agree: the Conjur test explains that the env vars are ignored once serviceID is set and that the unreachable `.invalid` Identity host is what would catch a mistaken fallback, and the username/password test points at it rather than restating it in opposite terms. The internal test-plan reference is gone, and 13 lines are down to 4.

I confirmed the mechanism empirically rather than by reading: forcing the username/password path against the same fixture fails with `dial tcp: lookup identity.example.invalid ... no such host`, which is exactly what the comment claims. Both tests pass at this head, and CI is green.

PR #836, opened 2026-09-10
CP-25962: Correct the sendSecretValues chart comment

roeezis:
## Summary
The chart comment for `config.sendSecretValues` said "the actual values of Secrets are not sent by default," but the value below it defaults to `true`. The sentence was accurate when the default was `false`; when the default was flipped to `true`, an adjacent line was updated but this one was missed. Fixing the comment, not the default — no behavior change.

`README.md` and `values.schema.json` are generated from `values.yaml` (`make ark-generate`), so they're regenerated to match.

## Test plan
- [x] `make ark-generate` run, diff limited to the corrected text
- [x] No functional change — value unchanged, docs/schema only

### DIFF
--- a/deploy/charts/disco-agent/README.md
+++ b/deploy/charts/disco-agent/README.md
@@ -419,7 +419,9 @@ This description will be associated with the data that the agent uploads to the 
 > true
 > ```
 
-Enable sending of Secret values to CyberArk in addition to metadata. Metadata is always sent, but the actual values of Secrets are not sent by default. When enabled, Secret data is encrypted using envelope encryption using a key managed by CyberArk, fetched from the Discovery and Context service.
+Enable sending of Secret values to CyberArk in addition to metadata. Metadata is always sent, and Secret values are sent by default too.  
+Set this to false to send metadata only.  
+When enabled, Secret data is encrypted using envelope encryption using a key managed by CyberArk, fetched from the Discovery and Context service.
 #### **config.cyberark.serviceId** ~ `string`
 > Default value:
 > ```yaml
--- a/deploy/charts/disco-agent/values.yaml
+++ b/deploy/charts/disco-agent/values.yaml
@@ -200,7 +200,8 @@ config:
   clusterDescription: ""
 
   # Enable sending of Secret values to CyberArk in addition to metadata.
-  # Metadata is always sent, but the actual values of Secrets are not sent by default.
+  # Metadata is always sent, and Secret values are sent by default too.
+  # Set this to false to send metadata only.
   # When enabled, Secret data is encrypted using envelope encryption using
   # a key managed by CyberArk, fetched from the Discovery and Context service.
   sendSecretValues: true
--- a/deploy/charts/disco-agent/templates/deployment.yaml
+++ b/deploy/charts/disco-agent/templates/deployment.yaml
@@ -86,7 +86,7 @@ spec:
                 key: ARK_SECRET
                 optional: true
           - name: ARK_SEND_SECRET_VALUES
-            value: {{ .Values.config.sendSecretValues | default "false" | quote }}
+            value: {{ .Values.config.sendSecretValues | quote }}
           {{- with .Values.http_proxy }}
           - name: HTTP_PROXY
             value: {{ . }}
--- a/deploy/charts/disco-agent/values.schema.json
+++ b/deploy/charts/disco-agent/values.schema.json
@@ -200,7 +200,7 @@
     },
     "helm-values.config.sendSecretValues": {
       "default": true,
-      "description": "Enable sending of Secret values to CyberArk in addition to metadata. Metadata is always sent, but the actual values of Secrets are not sent by default. When enabled, Secret data is encrypted using envelope encryption using a key managed by CyberArk, fetched from the Discovery and Context service.",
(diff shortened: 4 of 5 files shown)


deploy/charts/disco-agent/values.yaml
```diff
@@ -200,7 +200,8 @@ config:
   clusterDescription: ""
 
   # Enable sending of Secret values to CyberArk in addition to metadata.
-  # Metadata is always sent, but the actual values of Secrets are not sent by default.
+  # Metadata is always sent, and Secret values are sent by default too.
```
wallrj-cyberark, line 203: The chart's own template does not agree with the default this comment now documents. `templates/deployment.yaml:89` renders `{{ .Values.config.sendSecretValues | default "false" | quote }}`, so the template-level fallback is `false`, not `true`.

Today the two only disagree when the key is absent: `helm template --set config.sendSecretValues=null` renders `ARK_SEND_SECRET_VALUES="false"` while the README says values are sent by default. More to the point, this is the same rot trap you are fixing here. If anyone later drops the `sendSecretValues: true` line from values.yaml on the assumption that the template fallback covers it, the behaviour silently flips and the docs stay wrong.

Helm's `default` also treats a genuine `false` as empty, so the pipeline buys nothing even in the normal case. Changing line 89 to `value: {{ .Values.config.sendSecretValues | quote }}` would make the template say what the docs now say. Optional for this PR if you would rather keep the diff docs-only, but worth a follow-up.

deploy/charts/disco-agent/README.md
```diff
@@ -419,7 +419,9 @@ This description will be associated with the data that the agent uploads to the
 > true
 > ```
 
-Enable sending of Secret values to CyberArk in addition to metadata. Metadata is always sent, but the actual values of Secrets are not sent by default. When enabled, Secret data is encrypted using envelope encryption using a key managed by CyberArk, fetched from the Discovery and Context service.
+Enable sending of Secret values to CyberArk in addition to metadata. Metadata is always sent, and Secret values are sent by default too.  
+Set this to false to send metadata only.
```
wallrj-cyberark, line 423: There is one more copy of the same stale claim, and it is stronger than the one you fixed. `docs/datagatherers/k8s-dynamic.md:80` still says:

> Before Secrets are sent to the Preflight backend, they are redacted so no secret data is transmitted.

With `sendSecretValues` defaulting to `true`, `pkg/agent/run.go:170` sets `encryptSecrets`, `redactList` keeps the `data` field and adds `_encryptedData`, so secret data *is* transmitted (encrypted). A reader who checks the docs to answer "does the agent send my Secret data?" gets opposite answers depending on which file they open.

That line also points at `pkg/datagatherer/k8s/fieldfilter.go`, which no longer exists — the file moved to `pkg/datagatherer/k8sdynamic/fieldfilter.go`.

Out of scope if you want to keep this PR to the chart, but it is the same comment rot and worth sweeping in the same pass.

deploy/charts/disco-agent/values.yaml
roeezis: Confirmed via `helm template`: unset renders "false" today while the docs now say true — same rot class. Dropped `default "false"` (`f368c94`); unset now renders "true" (values.yaml's default), explicit `false` still renders "false".

deploy/charts/disco-agent/README.md
roeezis: Confirmed both — the file did move, and the claim contradicts run.go's encryptSecrets/_encryptedData path once sendSecretValues defaults true. Fixed in `f368c94`: corrected the redact-by-default / encrypt-when-enabled description, fixed the fieldfilter.go link, and pointed at the chart's sendSecretValues doc instead of restating it.

wallrj-cyberark:
The three red e2e checks here are my doing, and they are not telling you anything about your change.

I added the `test-ark`, `test-ngts` and `test-e2e` labels to get end-to-end coverage before merging. The suites did run, but they cannot pass on a pull request from a fork: GitHub does not expose repository secrets to `pull_request` runs from forks, so the credentials the suites need are empty. That shows up as:

- `Error: invalid reference: invalid registry ""` in `ark-test-e2e` and `ngts-test-e2e`, because the registry variable is empty
- `ARK_SECRET` logged as empty
- `test-e2e` failing at teardown with `gcloud.container.clusters.delete: argument NAME must be specified`, because setup never got as far as creating a cluster

Nothing to fix on your side. I am leaving the labels in place and will run the suites against `master` after this merges.

On the change itself, both points I raised earlier are addressed, and I have checked the new wording against the code:

- `pkg/datagatherer/k8sdynamic/fieldfilter.go` exists, and `SecretSelectedFields` keeps exactly `data.tls.crt`, `data.ca.crt` and `data.conjur-map`, so the new sentence is accurate.
- Dropping `| default "false"` is safe. `pkg/agent/run.go:170` is `strings.ToLower(os.Getenv("ARK_SEND_SECRET_VALUES")) == "true"`, so the Go side fails closed: anything that is not the literal string `true` means redact. There is no path where removing the Helm default silently starts sending secret values.

wallrj-cyberark (approved):
Verified rather than eyeballed: regenerated the chart docs and schema in a clean checkout of the head and got byte-identical output, confirmed the comment rot with `git log -L`, and checked the new wording against `SecretSelectedFields` and `pkg/agent/run.go:170`. Both review points are addressed.

The e2e checks are red only because fork pull requests do not receive secrets. I will run those suites against `master` after merge.

PR #838, opened 2026-09-16
CP-26533: allow config.cyberark.subdomain in agent YAML

roeezis:
## Summary

The CyberArk tenant subdomain is not a credential, but the agent could only read it via the `ARK_SUBDOMAIN` environment variable, bypassing its own YAML config entirely. This forced every install -- including Conjur-JWT-only installs that need no other credentials -- to provision a Secret just to carry one non-sensitive value.

It's a carry-over from before the Conjur JWT auth path existed, when the subdomain lived in the same Secret as the legacy username/password credentials because that Secret was the only config-delivery mechanism available at the time. The newer `service_id`/`account`/`jwt_source` fields were added to the proper YAML config struct; subdomain never was.

Adds `config.cyberark.subdomain` as a real config field. The environment variable remains a fallback for installs that still set it via Secret, so this is backward compatible. The chart's `ARK_SUBDOMAIN` Secret key becomes optional now that the config field can supply it instead.

## Changes

- `pkg/agent/config.go`: new `CyberArkConfig.Subdomain` YAML field
- `internal/cyberark/client.go`: `LoadClientConfigFromEnvironment` takes an explicit subdomain, falling back to `ARK_SUBDOMAIN` when empty
- `pkg/client/client_cyberark.go`: `NewCyberArk` threads the new subdomain param through
- Chart: `config.cyberark.subdomain` value, rendered into `config.yaml`, `ARK_SUBDOMAIN` Secret key now optional
- Regenerated chart docs/schema/snapshot tests

## Test plan
- [x] `make test-unit` passes clean
- [x] `make ark-verify` (helm lint, kubeconform policy checks, chart unittest) passes clean
- [x] New test case covering `config.cyberark.subdomain` set instead of the env var


deploy/charts/disco-agent/README.md
```diff
@@ -119,11 +122,13 @@ value for `config.cyberark.serviceId` below.
 
 ```sh
 # $SERVICE_ID is this cluster's own authn-jwt service ID from onboarding above — do not reuse it across clusters.
+# No Secret needed for this Conjur-JWT install — the subdomain isn't a credential.
 helm upgrade agent "oci://${OCI_BASE}/charts/disco-agent" \
      --install \
      --create-namespace \
      --namespace "$NAMESPACE" \
      --set fullnameOverride=disco-agent \
+     --set config.cyberark.subdomain="$ARK_SUBDOMAIN" \
```
wallrj-cyberark, line 135: This walkthrough is broken. `$ARK_SUBDOMAIN` is only exported in the "Add credentials to a Secret" section, and the comment two lines up tells the reader to skip that section.

Follow the Conjur-JWT path from the top of the page and you render `subdomain: ""`, then the agent exits at startup with `ErrMissingSubdomain`.

Either move the `export ARK_SUBDOMAIN=` line out of the Secret section, or inline the value here.

internal/cyberark/client.go
```diff
@@ -61,26 +62,22 @@ type ClientConfig struct {
...
-// Behavioral keys (ServiceID, Account, JWTSource, JWTFilePath) are set by the
-// caller from the agent YAML config (config.cyberark.*).
-func LoadClientConfigFromEnvironment() (ClientConfig, error) {
-	subdomain := os.Getenv("ARK_SUBDOMAIN")
+// LoadClientConfigFromEnvironment loads the CyberArk client config. subdomain
+// (from config.cyberark.subdomain) takes precedence; falls back to
+// ARK_SUBDOMAIN when empty. Also reads legacy ARK_USERNAME/ARK_SECRET, used
+// only when no Conjur service-id is configured.
+func LoadClientConfigFromEnvironment(subdomain string) (ClientConfig, error) {
+	subdomain = cmp.Or(subdomain, os.Getenv("ARK_SUBDOMAIN"))
```
wallrj-cyberark, line 82: Please log when both sources are set. Today config wins silently.

`selectAuthenticator` logs in the equivalent dual-source case just below, and the README explicitly promises "a warning is logged" there. This is the odd one out.

The case that bites: an operator migrating a tenant sets `config.cyberark.subdomain` to the new value while the old Secret still holds the previous one. The agent uses the config value and fails later inside service discovery with a host mismatch, with nothing in the logs to say two sources were in play.

internal/cyberark/client.go
```diff
@@ -61,26 +62,22 @@ type ClientConfig struct {
...
-//
-// Behavioral keys (ServiceID, Account, JWTSource, JWTFilePath) are set by the
-// caller from the agent YAML config (config.cyberark.*).
-func LoadClientConfigFromEnvironment() (ClientConfig, error) {
-	subdomain := os.Getenv("ARK_SUBDOMAIN")
+// LoadClientConfigFromEnvironment loads the CyberArk client config. subdomain
+// (from config.cyberark.subdomain) takes precedence; falls back to
+// ARK_SUBDOMAIN when empty. Also reads legacy ARK_USERNAME/ARK_SECRET, used
+// only when no Conjur service-id is configured.
+func LoadClientConfigFromEnvironment(subdomain string) (ClientConfig, error) {
```
wallrj-cyberark: The name now says the opposite of the contract: the parameter takes precedence *over* the environment. Your own new test file is called `subdomain_source_test.go`, which hints at the same discomfort.

Rename to `LoadClientConfig` while there is still exactly one production caller.

deploy/charts/disco-agent/tests/__snapshot__/configmap_test.yaml.snap
```diff
@@ -7,6 +7,7 @@ custom-cluster-description:
         cluster_description: "A cloud hosted Kubernetes cluster hosting production workloads.\n\nteam: team-1\nemail: team-1@example.com\npurpose: Production workloads\n"
         period: "12h0m0s"
         cyberark:
+          subdomain: ""
```
wallrj-cyberark, line 10: All four snapshots pin `subdomain: ""`. No chart test sets a non-empty subdomain, so the chart half of this feature is only covered at its default.

A template key typo *is* caught (nil renders as a bare value, not `""`), so this is a narrow gap rather than a hole. One `set: config.cyberark.subdomain: tlskp-test` case would pin the plumbing the PR exists to add.

wallrj-cyberark (commented):
No correctness bugs. The change is narrow and the generated files are in sync. Five low-severity points, four inline.

**Not inline, because the file is not in this diff:** `hack/ark/test-e2e.sh` still hard-requires `ARK_SUBDOMAIN` and writes it into the Secret. So the headline claim of this PR — that a Conjur-JWT-only install needs no Secret — is never exercised against a live cluster. The script also requires `ARK_USERNAME`/`ARK_SECRET`, so only the legacy path runs; that part is pre-existing, not yours. Worth a follow-up issue.

I added the `test-ark` and `test-e2e` labels and re-opened the PR to make the suites actually start. Note the green `ark-test-e2e` does not cover the no-Secret path, for the reason above.

<details>
<summary>What I checked, and how</summary>

- Traced every caller of the two changed signatures. `LoadClientConfigFromEnvironment` and `NewCyberArk` each have exactly one non-test production caller, both updated. Nothing else in the tree reads `ARK_SUBDOMAIN` or builds a `cyberark.ClientConfig` outside tests.
- `go test ./internal/cyberark/` passes, including the four `--machine-hub` subtests in `Test_ValidateAndCombineConfig`. The only failures in `pkg/agent` and `pkg/client` are the pre-existing `KUBEBUILDER_ASSETS` skips.
- `helm unittest ./deploy/charts/disco-agent` passes. I confirmed the new `deployment_test.yaml` is not vacuous by flipping an expected value: it fails as it should.
- Ran `make ark-generate` on the PR head. Zero drift: the regenerated `README.md` and `values.schema.json` match what is committed, including alphabetical `$defs` placement.
- `pkg/agent/config.go` uses plain `yaml.Unmarshal`, not strict. The new always-present `subdomain: ""` key will not break an older pinned agent image.
- Checked the other two charts (`venafi-kubernetes-agent`, `discovery-agent`). Neither renders cyberark config, so nothing was missed there.
- `res.CyberArk` is a value type, not a pointer, so `cfg.CyberArk.Subdomain` at the call site cannot nil-deref.
- `optional: true` on a `secretKeyRef` covers a missing Secret as well as a missing key, so the "no Secret needed" claim is sound.

</details>

wallrj-cyberark:
The `test-e2e` failure is not yours — ignore it.

I dispatched the same suite against `master` (run [35108371622](https://github.com/jetstack/jetstack-secure/actions/runs/35108371622)) and it fails identically:

```
Timed out after 300s waiting for certificate venafi-kubernetes-agent-e2e.<uuid> to appear in the Venafi inventory
make: *** [make/02_mod.mk:53: test-e2e-gke] Error 1
```

In both runs the agent itself is healthy: it acquires its JWT and OAuth token and logs `Data sent successfully` with 36 data readings. The certificate then never appears in the inventory. That is a backend condition, and it is the exact failure mode #835 was written to surface rather than hang on.

`ark-test-e2e`, `ngts-test-e2e`, `test` and `verify` are all green on master, and `ark-test-e2e`, `test` and `verify` are green here.

deploy/charts/disco-agent/README.md
roeezis: Fixed in 12ee476. Moved `export ARK_SUBDOMAIN=` out of the skippable Secret section into its own step before either path forks, so the JWT-only walkthrough no longer depends on a variable it told the reader to skip past.

internal/cyberark/client.go
roeezis: Fixed in 12ee476. LoadClientConfig now logs a warning (matching selectAuthenticator's dual-source message right below) when both config.cyberark.subdomain and ARK_SUBDOMAIN are set. Threaded logr.Logger through from the one call site in validateCredsAndCreateClient that already has it, rather than introducing a ctx param this function never had. Covered by a new test case with a capturing log sink asserting the warning actually fires.

internal/cyberark/client.go
roeezis: Fixed in 12ee476. Renamed to LoadClientConfig while there's still exactly one production caller, as suggested.

deploy/charts/disco-agent/tests/__snapshot__/configmap_test.yaml.snap
roeezis: Fixed in 12ee476. Added a custom-subdomain case setting config.cyberark.subdomain: tlskp-test, snapshot regenerated.

deploy/charts/disco-agent/README.md
wallrj-cyberark: Confirmed. `export ARK_SUBDOMAIN=` is now its own step before either path forks, so the Conjur-JWT walkthrough works read top to bottom.

internal/cyberark/client.go
wallrj-cyberark: Confirmed, and the new test is not vacuous. I removed the `log.Info` call locally and `TestLoadClientConfig_SubdomainPrecedence` fails, so the assertion really does pin the warning.

Threading `logr.Logger` rather than adding a `ctx` parameter is the right call. It does leave two logger idioms in one file, since `selectAuthenticator` uses `klog.FromContext(ctx)` a few lines below, but adding a context purely to carry a logger would be worse.

internal/cyberark/client.go
wallrj-cyberark: Confirmed.

deploy/charts/disco-agent/tests/__snapshot__/configmap_test.yaml.snap
wallrj-cyberark: Confirmed. The `custom-subdomain` case renders `subdomain: "tlskp-test"` at line 502 of the snapshot, so the non-default path is now pinned.

wallrj-cyberark:
All four review points are addressed in 12ee476. I have verified each one against ed4296b28bfe6c271969ea653e636ba2267aba6a rather than taking the replies on trust, and I have no further findings.

The `test-e2e` failure is still the pre-existing master breakage, not this PR. Same signature as before: `Timed out after 300s waiting for certificate ... to appear in the Venafi inventory`. `ark-test-e2e`, `test` and `verify` are all green.

One thing this PR does not close, which I raised earlier and is worth a follow-up issue rather than more work here: `hack/ark/test-e2e.sh` still hard-requires `ARK_SUBDOMAIN` and writes it into the Secret. So the Conjur-JWT-only install, the case this PR exists to enable, is still never exercised against a live cluster.

<details>
<summary>What I ran to verify the four fixes</summary>

- `go build ./...` passes on the new signatures.
- `go test ./internal/cyberark/...` passes, all packages. The only failure anywhere is the pre-existing `KUBEBUILDER_ASSETS` skip in `pkg/client`, which is unrelated.
- `TestLoadClientConfig_SubdomainPrecedence` runs all four subtests, none skipped.
- I removed the `log.Info` call locally and re-ran that test. It fails, so the dual-source warning assertion is real rather than incidental.
- The snapshot for the new `custom-subdomain` case renders `subdomain: "tlskp-test"`, so the non-default value is pinned.
- Read the regenerated README top to bottom. `export ARK_SUBDOMAIN=` now precedes both authentication paths, so neither walkthrough depends on a variable the reader was told to skip.
- `ClientConfigLoader` is still used by `pkg/client/client_cyberark.go` on both master and this branch, so the rename left nothing stranded.

</details>

roeezis:
Confirmed: `hack/ark/test-e2e.sh:23` hard-requires `ARK_SUBDOMAIN` and lines 61/65 always write it into the Secret, never exercising `--set config.cyberark.subdomain`. Agreed this is a follow-up, not scope for this PR — will open an issue to track adapting that script to the Secret-free path.

roeezis:
Opened #839 to track adding a live e2e path for the Conjur-JWT-only, Secret-free install. Not touching `hack/ark/test-e2e.sh`'s existing legacy path — that stays as-is for backward compatibility, this would be a second path alongside it.

internal/cyberark/subdomain_source_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+		_, err := cyberark.LoadClientConfig(logr.Discard(), "")
+		require.ErrorIs(t, err, cyberark.ErrMissingSubdomain)
+	})
+}
+
+// capturingSink is a minimal logr.LogSink that records Info() messages, used
+// to assert the dual-source warning actually fires rather than just trusting
+// the code path was reached.
+type capturingSink struct {
```
wallrj-cyberark, line 61: Non-blocking, and please do not change it in this PR — I am approving as is.

For next time: `k8s.io/klog/v2/ktesting` already does this, so the hand-rolled sink and its six methods are not needed. `auth_select_test.go` in this same package already uses `ktesting.NewLogger`.

One catch — buffering is off by default, so `ktesting.DefaultConfig` captures nothing. You need `BufferLogs`:

```go
logger := ktesting.NewLogger(t, ktesting.NewConfig(ktesting.BufferLogs(true)))

cfg, err := cyberark.LoadClientConfig(logger, "from-config")
require.NoError(t, err)

underlier, ok := logger.GetSink().(ktesting.Underlier)
require.True(t, ok)
assert.Contains(t, underlier.GetBuffer().String(), "both config.cyberark.subdomain and ARK_SUBDOMAIN are set")
```

`GetBuffer().Data()` gives you the structured entries if you would rather assert on fields than on text.

wallrj-cyberark (approved):
Approved. All four points from my earlier review are fixed in 12ee476, and I verified each one rather than taking the replies on trust — details are in the comment above.

The failing `test-e2e` is the pre-existing master breakage, not this PR, and there are no required status checks on `master`, so it does not block the merge.

One non-blocking note on the test logger, for next time rather than for this PR.

